PR #1, opened 2015-08-20
Deleting unneccessary comments (changelog)

kyrsjo:
As discussed, part of the changelog was inserted in the sources as comments, three times. This removes the noise.


kyrsjo:
Doing this as a separate fork / pull request has the extra benefit of testing the system.

We should not get a fast forward, even if this is clearly very fast-forwardable...

kyrsjo:
To note for later:
- The commit title is the title of this thread
- The commit message is empty - it is probably the textbox which opened when I clicked "merge pull request".

PR #2, opened 2015-08-25
Delete old/outdated files carried over from the SVN repository

kyrsjo:
I propose to delete the old stuff from the repository. To be merged when we are have an inclusive list of things to be deleted.


kyrsjo:
As @rdemaria commented off-line, this is a "trash-bin branch"

PR #16, opened 2015-09-23
Update user manual

kyrsjo:
This update to the sixtrack manual includes descriptions of new features, and cleans up/modernizes the LaTeX. Compatibility with Tex4HT is unfortunately broken, please use pdflatex.

Also included is a few fixes to some comments in sixtrack.s.


kyrsjo:
Note that I also added the PDF. I think this is reasonable every time we have a "released" manual which goes into the master branch, but not every time we change a bit of LaTeX....

Issue #12, opened 2015-09-18
CRLIBM issues: Compile with SSE2 support, drop x87


kyrsjo:
The function cos_rn() returns different results for different the same and large argument when compiled with different versions of GCC.

This can be avoided by using SSE2 math instead of x87, as this produces consistent results (consistent across compiler versions, and for gcc version 4.4.7 x87 and SSE agrees).

For gcc (crlibm), this is done by using the compilation flags "-mfpmath=sse -msse2".

kyrsjo:
@ldeniau @ericmcintosh

ldeniau:
Hi Kyrre,

Please use my cern.ch email address. gmail.com is my private emails.

cheers,
Laurent.

kyrsjo:
Hi Laurent,

I just tagged your username on GitHub, which then sent you the email. Do you have another github handle?

By the way: Apparently (from reading the email I forwarded you again) I did try to reduce the amount of code to just what was needed, but then the problem disappeared.

Cheers,
Kyrre

On 18 Sep 2015, at 14:57, Laurent Deniau <notifications@github.com<mailto:notifications@github.com>> wrote:

Hi Kyrre,

Please use my cern.ch<http://cern.ch> email address. gmail.com<http://gmail.com> is my private emails.

cheers,
Laurent.

ldeniau:
Hi Kyrre,

On Sep 18, 2015, at 3:03 PM, Kyrre Ness Sjøbæk <notifications@github.com> wrote:

> Hi Laurent,
> 
> I just tagged your username on GitHub, which then sent you the email. Do you have another github handle?
> 
> By the way: Apparently (from reading the email I forwarded you again) I did try to reduce the amount of code to just what was needed, but then the problem disappeared.

Interesting, it might be an hint that the problem is elsewhere…

Cheers,
Laurent.

> 
> Cheers,
> Kyrre
>

kyrsjo:
Hi Laurent,

As I said: I’ve already spent too much time on this, and given that (1) using SSE2 “fixes” (or at least masks it, if it is a GCC change) it and (2) new CRLIBM is on the way, I decided to stop trying to fix this problem. Especially since reporting floating point reproducibility issues to GCC seems to be very hard and fixing them is not something they seem to have much interest in.

However, it is still something to be aware of, and a reminder that the very long SixTest runs are sometimes useful.

Cheers,
Kyrre

[...]

PR #19, opened 2015-10-01
Use SSE for floating point math, no more x87.

kyrsjo:
This patch changes make_six, the crlibm makefile, and the sixtrack makefile to always use SSE math with gfortran. This makes the newer gfortran versions pass all the SixTrack checks, while having no effect on the results from earlier versions of gfortran.

This should probably be discussed with Eric etc.

I could not get the exact same results using the sixtrack makefile, but I suspect that this is rather due to the OPTIONS which I passed to it.

This patch fixes issue #12 .

### DIFF
--- a/SixTrack/crlibm/Makefile.template
+++ b/SixTrack/crlibm/Makefile.template
@@ -13,7 +13,8 @@ crlibm.a: $(OBJS)
 	ar -rv crlibm.a $(OBJS)
 AM1_CFLAGS = -fPIC -std=c99 -Wall -Wshadow -Wpointer-arith  -Wcast-align -Wconversion -Waggregate-return -Wstrict-prototypes -Wnested-externs -Wlong-long -Winline -pedantic -fno-strict-aliasing
 DEFS = -DLINUX_INLINE -DHAVE_CONFIG_H -I.
-CFLAGS = -g -O2 -m32
+CMATHFLAGS = -mfpmath=sse -msse2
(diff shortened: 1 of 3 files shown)


kyrsjo:
Note that since crlibm's makefile is changed, this affects ALL sixtrack+crlibm builds, since gcc is always used to build crlibm.

kyrsjo:
Discussed with Eric ( @ericmcintosh ) , he was OK with it.

Something else: With this patch, the enable_xp/disable_xp calls will be vestigal, i.e. they won't serve any purpose. Should they be disabled in crlibm, like they are for OS X platform (which always uses SSE)? Or is it OK as it is?

@rdemaria : Could you please check that my mods to the makefile are OK, and if so, do the merge?

ericmcintosh:
Hmm  I need to sleep on this (and run some tests).   Eric.

kyrsjo:
Eric: As stated above, all tests included in SixTest produce the same results with ifort, old GCC with or without SSE, or new GCC with SSE. The only "odd one out" is the new GCC without SSE, as discussed previously.

ericmcintosh:
Right, good. Need to clarify though. Remember gfortran, ifort,
nagfor, etc etc.

Look forward to reviewing with you Monday.    Eric.

kyrsjo:
After discussing with Eric, I will merge this. It passed all the tests, and will make things compatible with newer gfortran.

PR #21, opened 2015-10-07
collimation fix

rbruce1:
Updated string matching logic to recognize new collimator names (TCTPV, TCTPH, TCSP)

### DIFF
--- a/SixTrack/sixtrack.s
+++ b/SixTrack/sixtrack.s
@@ -29467,6 +29467,12 @@ C     Convert r(1), r(2) from U(0,1) -> rvec0 as Gaussian with cutoff mcut (#sig
      &                 ) then
                      nsig = nsig_tcstcdq
                   endif
+               elseif(bez(myix)(1:4).eq.'TCSP' .or.                        &
+     &                 bez(myix)(1:4).eq.'tcsp') then
+                  if(bez(myix)(9:11).eq.'6.B'.or.
+     &                 bez(myix)(9:11).eq.'6.b') then
+                     nsig = nsig_tcstcdq
+                  endif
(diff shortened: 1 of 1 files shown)


rdemaria:
Thanks, I will make a basic check...

rdemaria:
Can you provide a test that proves this change works for you?

kyrsjo:
I fetched Roderik's branch (my "upstream" remote is the main SixTrack repository):

```
10:06:30 kyrsjo@pcbe15843 SixTrack-git $ git fetch upstream pull/21/head:collimation_update
remote: Counting objects: 22, done.
remote: Total 22 (delta 14), reused 14 (delta 14), pack-reused 8
Unpacking objects: 100% (22/22), done.
From github.com:SixTrack/SixTrack
 * [new ref]         refs/pull/21/head -> collimation_update
10:06:39 kyrsjo@pcbe15843 SixTrack-git $ git checkout collimation_update 
Switched to branch 'collimation_update'
```

Only code quality comment I could make is that the first new end-of-line line-wrapping-ampersand is in the wrong column (it is 3 places too far); but these are anyway ignored by the compiler so we could actually just get rid of them. I don't even use them in the code that I write. So whatever.

Regarding compilation least it compiles fine with gfortran 4.7. Since Roderik has tested that this change works, it looks reasonable (it is after all only , is only affecting the collimation block, I say "merge!".

Let's keep the version number where it is; I'll also merge the SSE stuff, and together with my FIR fix + a few other small things (use the right index when printing BLOCK element names in linopt!) it could be v2.5.30.

rdemaria:
I would just wait for a test to merge it..

On Thu, Oct 8, 2015 at 10:20 AM, Kyrre Ness Sjøbæk <notifications@github.com

Issue #15, opened 2015-09-22
Binary output from DUMP


kyrsjo:
Today, DUMP can only produce ASCII files of varying format. Which format to use is controlled by the format flag, set on a per-file basis.

In order to be able to produce smaller files, which can be read and written faster, a format type giving binary output would be very useful.

kyrsjo:
Fixed in #220

PR #42, opened 2016-01-15
FMA

mfittere:


kyrsjo:
Cool that it is getting closer to release!

Some initial comments:
 * There seem to be quite a few debug outputs which are still in place
 * Generated files (.a and the Makefile) should not be committed to the repository
 * Writes should be both for CR and non-CR versions
 * I need to evaluate the changes to stringzerotrim and DUMP...
 * I'm about to (next monday probably) make a somewhat significant commit to LINOPT, but I think we're orthogonal: https://github.com/kyrsjo/SixTrack/commit/a27a9e9a6760b07efdbed94fa250dc2a71570509

mfittere:
- yes, I still have to remove the debug outputs
- yes there are still the cr and non-cr writes
- Stringzerotrim didn't really change except for the name and that it is an indpendent subroutine now
- I added some variables to dump. I think e.g. the closed orbit might be also useful in general, so what about putting it e.g. in the header of the dump files (this is not needed for the FMA, so just a suggestion)
- linopt is orthogonal to the fma stuff

mfittere:
- I removed all the debug statements in the code and added the +cr for the output. I only left the file with the normalized coordinates in as I think it is still useful to keep for the moment for debugging.
- I still have to check the unit of dp or dp/p or dE/E for the closed orbit ...
- there is still the error message when running the code (haven't figured that out yet):
Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG IEEE_DENORMAL
- Then as a general question for handling the errors:
I now always put an error message and then aborted (call prror(-1) ). I don't know if in some cases it would be better to just print an error message, but continue? How is it handled in the rest of the code?
- Kyrre, I inserted a part, where you have to resume the dumpfile. Could you please check, that it is done correctly (I marked the place with "Kyrre" in the code? Thanks :-)!!!

mfittere:
for Kyrre:
here are the lines for the dump file. Could you please check that the file is handled correctly? Thanks!!!
!     Kyrre: resume position in dumpfile if file was open, otherwise close it
              if(lopen) then
                open(dumpunit(j),file=dump_fname(j), status='old',      &
     &form='formatted',action='readwrite')
                do k=1,dumpfilepos(j)
                  read(dumpunit(j),'(A)',iostat=ierro) ch
                  call fma_error(ierro,'while resuming file ' //        &
     &dump_fname(j),'fma_postpr')
                enddo
              endif

kyrsjo:
Hi Miriam;

I'll have a look. Meanwhile, could you please merge the latest master into your branch? Also, if you could describe your stuff in the user manual (six.tex)...

Cheers,
Kyrre

kyrsjo:
Thanks for the merge; that was really quick!

Eventually, test input would also be great; for now maybe just attach the fort.2/3 to this thread (rename the files to "fort.2.txt" etc. so GitHub accepts them)?

--- Kyrre

mfittere:
Hi Kyrre,
thanks for taking a look. I put some example here:
/afs/cern.ch/user/m/mfitter0/public/kyrre/test_fma/

We can create a test example once the sixdesk part is also finished.

Miriam

kyrsjo:
Hi, a comment regarding your FMA branch: Could you please remove the files "SixTrack/Makefile", from the repository? I also see a couple of binaries which have been checked in:
- SixTrack/libmathlib.a
- SixTrack/libkernlib-shift.a
- SixTrack/lib32/libmathlib.a
- SixTrack/plato_seq.o
Is there any reason why these files needs to be there? Also, what is plato?

kyrsjo:
Something else: I see you've added some more stuff to the dbdump block; especially the parameter dump_struc_max and the arrays dump_tas and dump_clo.

Regarding the parameter: If this is really the max number of STRUCTURE ELEMENTS in fort.2, rather use nblz from block parpro, this flag is rescaled for collimat/bignblz/hugenblz binaries.

mfittere:
Hi Kyrre,

I can remove the makefile.
The binaries are intentional. We tried for a long time to get the cernlib working on linux, but didn't manage. Therefore we directly put the libaries there. The libmathlib and libkernlib are needed for the plato stuff = different methods to calculate the tune. plat_seq.o is the binary for this tune analysis.

This is an ugly temporary fix, which should be removed at some point.

Miriam

mfittere:
Hi Kyrre,

I now replaced the dump_struc variable. Please check if this is fine for you now.

Cheers,
Miriam

mfittere:
Hi Kyrre and Riccardo,

I cleaned up everything except for implementing the crlibm in plato and adding it to the makefile. The fma branch is merged with SixTrack/master branch.

Thanks for having a look!
Cheers,
Miriam

kyrsjo:
Comment by Miriam copied from https://github.com/kyrsjo/SixTrack/pull/1 :

> 1) resuming position of dump file:
> see code after:
> ```
>      &eps123_avg(3),eps123_0(1),eps123_0(2),eps123_0(3),                &
>      &phi123_0(1),phi123_0(2),phi123_0(3)
>             enddo
>             close(200101+i*10)! filename NORM_* (normalized particle amplitudes)`
> ```
> I simply used open(...,position='append',...). Please check if you see any problem with that.
(tripple backtick to have the "code formatting preserve newline)

The next lines are:
```
!    resume initial position of dumpfile = end of file
            open(dumpunit(j),file=dump_fname(j), status='old',          &
     &form='formatted',action='readwrite',position='append',            &
     &iostat=ierro)
            call fma_error(ierro,'while resuming file '//dump_fname(j), &
     &'fma_postpr')
```

I would think that's OK (and it is a more elegant solution than my "open, read everything to '/dev/null', backspace once when hitting EOF"). @rdemaria , what do you think? The best may be to simply test it: Run the code once, copy out the dump file, then edit the code to insert a write just after this open, run the modified code and diff the dumpfiles. If everything works as it should, the files should be the same except for the extra line from the extra write.

kyrsjo:
Please delete these files:
 * plato_seq.o
 * plato_seq_org.f
 * libmathlib.a (?)
 * libkernlib-shift.a
 * lib32/libmathlib.a (?)

The autogenerated Makefile should probably also be removed. Further, make_plato should be incorporated in the makefile and make_six, and the ubuntu-specific hack "-L/usr/lib/gcc/i686-linux-gnu/5.2.1" in the makefile should be removed. I also suspect you no longer need the "-lmathlib" in the CERNLIB command?

mfittere:
Hi Kyrre,

ok for all! I'll try to have a look in the afternoon (I now still have to prepare sth. and then have a meeting).

Cheers,
Miriam

kyrsjo:
Things remaining (TODO):
- Fix make_six and makefile so that it always compiles on ubuntu and non-ubuntu alike
- Add compilation of PLATO to makefile/make_six: `gfortran -m32 plato_seq.f -o plato_seq.o -c` (or maybe rather $FC?
- Make sure PLATO is correctly linked
- CRMLIB+PLATO
- Delete .o's etc. as discussed in comment above.

mfittere:
Hi Riccardo,

we just discussed the above to do list with Kyrre. 
- I have trouble to get make_six working -> do you have some time to help?
- add plato to makefile/make_six -> could you please add it?
- crlibm+plato -> any ideas? we would have to replace all the sin and log etc. in the code

Do you have time some time this week to attack the above points? I guess to call over skype would be the easiest?

Thanks a lot!
Miriam

kyrsjo:
For the last bulletpoint (CRLIBM+PLATO), I would think the easiest would be to convert plato.f to plato.s, and use astuce to create plato.f, putting +if/+ei for crlbm/not-crlibm. In addittion, someone would need to go over the code and check that it is properly paranthesizied.

mfittere:
That sounds like a good idea. If one of you could give me some instructions/explanations, I could do the tedious the +if/+ei ...

rdemaria:
Parenthesis may not needed immediately because this part does not end up in
fort.10

mfittere:
We could do a first step and just replace the functions and add a comment that at some point the parenthesis would have to be handled correctly (if needed?)? This version we could then already put in the main branch of sixtrack.

kyrsjo:
If we anyway don't do the replacement, why bother with CRLIBM at all? It will anyway not be 100% reproducible (but likely better than in the tracking, since it is not a chaotic system)...

kyrsjo:
Another thing that's known to be missing: A description of the new block in FORT.3 in the user_manual.

mfittere:
It's on my todo list ;-).

rdemaria:
We need reproducibility to test for errors and resubmit simulation. It is
true that there could be a mistake in the tune evaluation, but you can spot
it more easily. Still it is a temporary measurre to speed up the
deployment...

mfittere:
to put some names behind the different items:

**1. Makefiles (Riccardo/Kyrre):**
-  Fix make_six and makefile so that it always compiles on ubuntu and non-ubuntu alike
- Add compilation of PLATO to makefile/make_six: gfortran -m32 plato_seq.f -o plato_seq.o -c (or maybe rather $FC?)
- Make sure PLATO is correctly linked

**2. CRLIBM/PLATO (Miriam + help from Riccardo/Kyrre):**
- substitute functions by crlibm functions
- check brackets
- convert plato.f to plato.s + astuce to convert plato.s to plato.f. Use +if/+ei for crlbm/not-crlibm

**3. Documentation (Miriam):**
- Theory documentation in SixTracklib
- User documentation in SixDeskmanual
- code implementation documentation in Twiki

kyrsjo:
Miriam;

Do you know anything about this warning when compiling PLATO?
```
gfortran -c -frecord-marker=4 -O4 -m32 -g -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 plato_seq.f -o plato_seq.o
plato_seq.f:330.21:

      CALL FFT_PLATO(ZSING,NPOINT,-1)                                   
                     1
Warning: Type mismatch in argument 'data' at (1); passed COMPLEX(8) to REAL(8)
plato_seq.f:393.21:

      CALL FFT_PLATO(ZSING,NPOINT,-1)                                   
                     1
```

It looks like it might be able to give wrong results...

kyrsjo:
Hi Miriam,

This pull request should fix your build problems:
https://github.com/mfittere/SixTrack/pull/4

Also, I setup a virtualbox with Ubuntu 15.10 to test your compilation issues and found none. Both make_six and the makefile runs fine?

The packages I have installed are: gfortran, gfortran-multilib, build-essential, git, virtualbox-guest-dkms virtualbox-guest-utils, virtualbox-guest-x11 and of course the dependencies for the listed packages. I also checked "install multimedia stuff" during the installation. Also installed are all updates as of today. Otherwise, it is a 100% stock 64-bit ubuntu install.

kyrsjo:
For the docs, please also write how to use it in the user manual.

mfittere:
Normalisation is now working, I checked it for several cases. I have a problem with crlibm. The log_rn function is not calculated correctly. It just gives back 0.
```
DOUBLE PRECISION FUNCTION TUNENEWT1(X,XP,MAXN)
...
DUEPI=ATAN_RN(1D0)*8D0
MFT=INT(LOG_RN(FLOAT(MAXN))/LOG_RN(2D0))"
```
Did you have this issue before?

kyrsjo:
Please see this pull request:
https://github.com/mfittere/SixTrack/pull/7

Also: A lot of the functions in plato_seq have statements of the type "implicit double precision (A-H,O-Y)" etc. Please replace these with an IMPLICIT NONE statement and explicit declaration of the variables used in that function. IMPLICIT is one of terrible-est ideas ever implemented in a computer language, it leads to bugs just like the one that bit you...

mfittere:
Here a short summary of the changes:
1) plato_seq.s is completely explicit and crlibmed
2) I fixed a bug with the normalization matrix
still to do:
1) add flag for optional normalization
2) write a test: I have a test, if you could help me convert it to an official test, this could be done fast
Otherwise I think it should be now hopefully finally done (soon ;-) )

mfittere:
I am now just adding the flag for the normalisation and then except for the test, it should be hopefully ready and bug-free :-)

kyrsjo:
I think I will just have to trust you on that; let me know when you're ready to merge!

Probably it would be a good idea to first merge master into fma, resolve all the conflicts (both git conflicts and others, like for the dynk_stringzerotrim renaming), then once we're satisfied, merge it back into master.

Having "pre-merge" examples would probably be a good idea...

Since this runs with the "standard" SixTrack, maybe the best way to test it would be to add a folder (and it's "in" partner and sixin.zip) to the SixTest folder, then run to produce OUTPUTS.canonical, where OUTPUTS is fort.6 (running ./SixTrack_blabla > fort.6), fort.10, fort.90, and whatever you think is relevant to compare for the FMA outputs?

mfittere:
For a last check I will run an FMA example in SixTrack and Lifetrac and compare the results of the plots, now that I have another code to compare to. Once I have done this finally test, we can merge to the master (of course after resolving all the conflicts first).

kyrsjo:
OK, do your tests, and add the test-examples with canonicals to SixTest. Then let me know when you're ready to merge. As I said, since we now can't merge automatically due to both merge conflicts and also things which have diverged (and thus won't compile even when we resolve the merge conflicts), I think we should merge to a new branch first, fix things here, then merge this into master once everything is working.

I have a pretty good handle of what's going on in master, so I could help here.

mfittere:
so the test is in my directory: test_fma_coll_coup_dp1e-5/15/512turns/ (for me to remember)

kyrsjo:
OK, almost there now!

Master was merged into FMA branch, bringing them up to date:
https://github.com/kyrsjo/SixTrack/commit/268a3fe2ab55d9d12d23825deb8ff779ee780ab0
This will soon be merged into miriam:fma.

Todo:
 * Documentation (Doc/user_manual/six.tex)
 * Run the SixTest tests
 * Integrate your test into the main test suite (test at /afs/cern.ch/user/m/mfitter0/public/kyrre/test_fma )
 * General code review

mfittere:
Tests:
returned all same. I got one warning with ./check_10 :
`     ====================================
      CASE exact
      ====================================
Warning: Both fort.10.canonical and fort.10.run_pro_SixTrack_4534_cernlib_crlibm_gfortran_O4_1000_27514 are zero size
fort.10 SAME`

Missing items:
- Integrate your test into the main test suite (test at /afs/cern.ch/user/m/mfitter0/public/kyrre/test_fma )
- General code review

To run on lxplus:
- change the utilities/sixtracking.lsf file and add to copy the fma_sixtrack file

kyrsjo:
Yes, this is normal: The "exact" test doesn't produce a fort.10 (or rather, it is empty IIRC). See #64 and #65. We should maybe adjust the message a bit, but the old version of the script was just saying "not yet available fort.10.canonical" IIRC, so the new version who at least says "SAME" is an improvement.

kyrsjo:
Another thing to check: That we can still build the DA version, after the changes to make_six / makefile.

kyrsjo:
Hi Miriam, thanks for improving the manual! Still a few more things:
* Could you add a reference to a description of the methods, or describe each of them?
* "in the mask file prior to the SixTrack run" -- This mask file is not described anywhere in the manual, if I remember correctly it's a SixDesk feature? Please say so.
* "It is important to install the additional markers after cycling the machine if the machine
is cycle at the location of the additional (e.g. IP3), as they are installed in front of the element given
in the from statement in the cycle command." -- Please take a look at this sentence. And "cycle" is not a SixTrack command...
* Bump the manual version number to what we have today+1 (to 4.5.35), bump "January 2016" to June, and add your name next to mine.

kyrsjo:
For the code, I'm preparing a pull request here:
https://github.com/mfittere/SixTrack/pull/9

In it, I will fix any "aestetic" problems I find while reviewing the code. If I find any logic problems etc. I'll just describe them here and let you handle it.

kyrsjo:
Some small things you need to check:
* Definition of `dump_clo`, you wrote a comment "check units used in dump_clo (is x' or px used?)"; is this resolved? If so, please modify the comment. A similar comment appears in block `umlalid`: "-> check units used in dump_clo (is x' or px used?) ", please fix this also. This commit may be interesting: https://github.com/SixTrack/SixTrack/pull/42/commits/a4a84fefc5620b9bad51f229dd0ebb3b9f396701
* Parameter fma_npart_max - is there any reason not to just make this equal to npart, or why it is 70 not 64? We might (in the very near future) want to increase npart...
* Comment "calculate optics parameter for element typ" in block `umlalid` - could you expand on this?

kyrsjo:
One more thing for the manual: Could you add your input/output units to table C.1?

kyrsjo:
Did you ever try to compile it with ifort? It fails for me...
```
$ source /afs/cern.ch/sw/IntelSoftware/linux/setup.sh
$ source /afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2011/composerxe-2011.2.137/bin/ifortvars.sh ia32
$ source /afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2011/composerxe-2011.2.137/bin/iccvars.sh ia32
$ ./make_six ifort
```
At the bottom of lots of output:
```
plato_seq.f(476): error #6404: This name does not have a type, and must have an explicit type.   [COMPLEX]
        CDATA(I/2+1)=COMPLEX(DATA(I),DATA(I+1))
---------------------^
compilation aborted for plato_seq.f (code 1)
Makefile_makesix:34: recipe for target 'plato_seq.o' failed
make: *** [plato_seq.o] Error 1
```

kyrsjo:
Hi Miriam;

I've done my initial checks and wrote the testing harness (which will probably be usefull for many other things as well). If you could merge the modifications, that would be great.

Before we can merge to master, you need to finish the manual, address the comments I've noted above, and make it work with ifort. Then I'll do another code review, full exercise of test harness (including multiple compilers and versions), and if everything is OK, we merge!

kyrsjo:
Quick howto for running SixTest:
1. Compile the executables with make_six: The relevant combinations of options are "gfortran", "gfortran cr", "ifort" and "ifort cr".
2. Each in a separate terminal (I assume you have 4 cpu's -> 4 SixTracks; if you are running in VirtualBox check how many CPUs you gave to Linux!):
`export EXECS=/absolute/path/to/one/of/the/SixTrack/execs/that/you/made/above`
then `export TESTS="bb bbe51 bbe52 bbe571ib0 crabamp dipedge distance eric exact fma frs frs60 javier lost lost lostnotilt notilt s316 thick4 tilt prob1 prob3"`
3. Then you run the tests - if `EXECS` is pointing to a normal version, use `./run_pro SOMENUMBER`, if it is a cr version use `./run_kill SOMENUMBER`. When it finishes, run `./check_10 SOMENUMBER`, `./check_90 SOMENUMBER` and `./check_extras SOMENUMBER`. The SOMENUMBER should be different each time you run the scripts `run_pro` or `run_kill`.

A small hint: run the `prob1` and `prob3` tests last and only if everything else is OK, they take about 24 hours...

mfittere:
-  manual is updated
- the units of the dump_clo have to be checked, but there is no easy check for this and the differences are very, very, very small for the LHC. So for the moment please just leave the comment.
- Thanks for inserting the initial sanity checks for the FMA input block.
- why did you change my beautiful if ... else for the tune methods to select case? I think it is fine with the if statements. Anyway, it's also fine like this.
- I don't understand the error message from the ifort compiler, in particular as cdata is explicitly defined with `COMPLEX*16 CDATA(NN)`

kyrsjo:
Hi, this is great :) Will read and nitpick manual tomorrow. One thing: Is there a more stable reference to SixDesk than overleaf?

I've learned (and then had it drilled into me by Laurent) that one should always prefer switch statements over if/else when it is possible. If you consider it, the "if" has to execute the two trim functions for every case it checks, and the same for someone reading the code - the switch is much more straight-forward (and easier for the compiler to optimize, not that this is a big deal in this case).

The problem with the COMPLEX function is probably just that it is one of the many (very useful) GNU extensions of the FORTRAN language:
https://gcc.gnu.org/onlinedocs/gfortran/COMPLEX.html
Note the "Standard: Gnu extension". 
There is also "CMPLX" and "DCMPLX", altough the latter may be another GNU extension, altough it is one that is supported by ifort. Nagfor may be a bit more picky, but they propose a solution - maybe that could work everywhere? If you need double complex of course...
https://gcc.gnu.org/onlinedocs/gfortran/CMPLX.html#CMPLX
https://gcc.gnu.org/onlinedocs/gfortran/DCMPLX.html#DCMPLX
https://software.intel.com/en-us/node/526080
http://www.nag.com/nagware/np/doc/faq.asp#faq25

mfittere:
as a summary:
it compiled with all 4 options (gfortran, gfortran cr, ifort, ifort cr). I am now running the tests except the prob1 and prob3.

kyrsjo:
Great!

Regarding the FMA particle number, why is this at all a free, hard coded parameter? Why not just use npart?

mfittere:
the results of the tests:
**SixTrack_4534_cernlib_crlibm_gfortran_cr_O4**: two tests failed

- exact:
```
      ====================================
      CASE exact
      ====================================
Error : One of the files is zero size while other is not
fort.10 DIFFERENT!!!
```
- fma:
```
Files 'fma/fma_sixtrack.canonical' and 'fma/fma_sixtrack.run_kill_SixTrack_4534_cernlib_crlibm_gfortran_cr_O4_1000_25314.killed' are DIFFERENT!! 

      *****************************************
      Test fma FAILED
      *****************************************

```

kyrsjo:
Huh. Did you run `prob1`/`prob3` also? Interesting that the `fma` test failed for gfortran, which was also used to produce that canonical... Was this after changing the cmplx thing? What happens for ifort or if you change it back? Maybe a `KIND` argument should be used?

I'll look into the `exact` problem, can you look at the `fma`?

kyrsjo:
For the crash (fma/ifort/cr): What? `forrtl: severe (64): input conversion error, unit -5, file Internal Formatted Read` That was an ... interesting ... error.

Could there be some variable you calculate during tracking and not saved, that is used for picking a output unit for a write? Does it fail for normal ifort/no CR?

mfittere:
HI Kyrre,

yes, I'll take a look at the fma.

Miriam

kyrsjo:
Fixed the FMA crash. Now fixing some NORM_ output for FMA. The cr/exact problem may actually be an issue with the running script vs. the check_10...

kyrsjo:
Pull request fixing the FMA crash + fixing output consistency of normalized DUMP files: https://github.com/mfittere/SixTrack/pull/10

Strangely, there are now differences in fma_sixtrack compared to the canonical you gave me some days ago, and which worked when I checked it in, even when using the same compiler. The issue there seems to be some numerical differences in TUNEABT and TUNEABT2.

Also to be fixed in this output file: Use a specified format string (maybe just '(a)'?) for the header lines. Also, the `adjustl` is totally unnecessary - it only removes leading spaces in a hardcoded string...

kyrsjo:
FYI, don't worry about the difference in the CR version + exact test + fort.10: This is the outputted fort.10:
```$ cat fort.10.run_kill_SixTrack_4534_cernlib_crlibm_gfortran_cr_O4_301_25540
  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  40534.000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  0.0000000000000000e+000  6.9829499721527100e-001
```
where last number fluctuates. This is true for the upstream/master version as well as for the FMA version. The non-CR version just produces an empty (0 bytes) fort.10, so this is a CR/not-CR difference, not a FMA problem. Perviously this was hidden, but @vikasnt 's fixes to the check_10 unmasked it (see #64 and #65 ).

mfittere:
thanks for the changes!

So the ifort crashed because of?
`write(200101+i*10,*) adjustl('# id turn pos[m] nx[1.e-3'`

kyrsjo:
This fixed the crash:
https://github.com/kyrsjo/SixTrack/commit/399ae2a7f758cd2238e83bafeb1ca73b8343a202
The problem was that the strings from getfields etc. are zero terminated. That's great when interfacing c codes, gfortran didn't mind (it's probably all c underneath), but ifort wasn't too happy.
On Jun 8, 2016 00:40, mfittere <notifications@github.com> wrote:thanks for the changes!

So the ifort crashed because of?
write(200101+i*10,*) adjustl('# id turn pos[m] nx[1.e-3'

kyrsjo:
Re. new merge conflicts: It's probably just some stuff with stringzerotrim after I merged PIPEFUN into master. I'll give you a pull request with the updated master merged into your branch when you're ready to merge.

mfittere:
test summary:
gfortran cr: the exact case failed (this we know why)
gfortran: passed
ifort cr + ifort: exact failed, the header is screwed up in fma_sixtrack (note the newline in the header!!!!):
```
 # inputfile method id q1 q2 q3 eps1_min eps2_min eps3_min eps1_max eps2_max eps
 3_max eps1_avg eps2_avg eps3_avg eps1_0 eps2_0 eps3_0 phi1_0 phi2_0 phi3_0
```

kyrsjo:
Good news!

For the `fma_sixtrack` output file, please always specify explicit format statements like I did for the normalized dumps in, i.e. write(WHATEVER,'(a)'): https://github.com/SixTrack/SixTrack/pull/42/commits/76dc52ee76ae74d00b12b03ade4409bc127cbfad#diff-521f1c7cafbcb45c9dd06685c77d9c81

Also please get rid of the adjustl() statements, I don't understand what you need them for?

Did this fix the numerical differences in the fma analyses? https://github.com/SixTrack/SixTrack/pull/42/commits/4013bac7520b2dc06f8d9e7e940633ff100ca53e

mfittere:
summary and changes:
- I extended the test to 4096 turns. I think 512 might be a bit little for some long term behaviour ...
- changed fma_npart_max to npart
- run the fma test in SixTest for all gfortran, gfortran cr, ifort, ifort cr

ToDo:
- merge to main branch
- update README in SixTest for extra tests (e.g. check_extras ...)
- redo SixTest tests

kyrsjo:
Did the merge, only conflict was six.pdf. Recompiled it as it now contains both FMA and DYNK:PIPE. Next step is to fix a few things that cannot really be automatically merged, i.e. `dynk_stringzerotrim` -> `stringzerotrim`. Pull request for the merge (in work, doesn't even compile right now so don't accept it just yet!) https://github.com/mfittere/SixTrack/pull/11

Seems the DUMP canonicals after the merging have a quite significant file size (34 MB each), but whatever... At some point, I will provide a binary output from DUMP ( #15 ) and it will be all OK.

kyrsjo:
FYI, you can delete the branch from your fork; this way you don't need to download it every time you clone your fork or add it as a remote :)

PR #48, opened 2016-02-22
bug fix to rbruce-morecollimator.

amereghe:
When writing in sigmasettings.out/twisslike.out, tctps where not correctly identified, since string matching was truncating the bez() to 4 chars instead of 5.

Moreover, few extremely long lines (full of empty spaces) removed.


kyrsjo:
Whoops! Thanks for the fix! I've fetched and checked that it compiles well; again it's fairly minor so I'll just merge, no version number bump needed.

kyrsjo:
@rbruce1

PR #49, opened 2016-02-29
Add DYNK to SixDA

kyrsjo:
Fix from Eric which makes the Differential Algebra version compile without complaining about linker error to COMNUL and various DYNK routines.


kyrsjo:
Note that the DA version still needs both CERNLIB and NAGLIB, which needs the -lg2c flag. This works on LXPLUS, but not on my local workstation (SLC7)...

PR #50, opened 2016-03-09
In LINOPT: Treat crab cavity dipoles the same way as other RF multipoles.

kyrsjo:
Including writing out the twiss functions for +23 (not only -23) when in collimation version.


kyrsjo:
@rdemaria I think this is a pretty simple fix, OK if I just merge it?

Thanks to @KFubuki for pointing out the bug.

PR #51, opened 2016-03-11
Improved error detection and bugfixes in COLLIMAT:

kyrsjo:
- Detect if no particles will be tracked (such as if there is no collimation block)
- Detect if initial distribution file is missing
- Use an unused FORTRAN unit number when loading initial distribution
- Detect if collimation database is missing and output appropriate error message


kyrsjo:
@rbruce1

kyrsjo:
Could we merge this one? @rbruce1 ?

PR #52, opened 2016-03-15
Remove cernlib from plato + clean up plato_seq.f

mfittere:
Hi Kyrre and Riccardo,

1) I cleaned up the plato_seq.f file. It compiles for me using:
gfortran -nostartfiles -g -m32 plato_seq.f -o plato_seq.o

I get now a weird error message if I do **not** use -nostartfiles -g also with the file we compiled previously with kyrre (branch kyrsjo/PLATO_LOBOTOMIZE_FOREALZ)
2) using this plato_seq.o file and removing the linking of the cernlib libraries from the makefile (Kyrre, the makefile is unchanged to the one we did together!), I get the following error message:
/usr/bin/ld: error in plato_seq.o(.eh_frame); no .eh_frame_hdr table will be created.

Could you please take a look and hopefully figure out what goes wrong?

Thanks a lot!
Miriam


kyrsjo:
Resolved after Skype discussion, "plato" branch is actually (manually?) merged into "fma" branch anyway. Please continue discussion in #42 .

PR #53, opened 2016-03-16
Add missing prror() calls after detecting a bug.

kyrsjo:
Thanks to Miriam for pointing that out!

Before merge: Update modification date.


kyrsjo:
Before merging: Test that it doesn't choke on DYNK:FILE.

PR #64, opened 2016-05-13
commiting fort10.canonical dummy file

vikasnt:


kyrsjo:
Looks good, works as intended.

A few cosmetic changes before I can accept the pull request:
- Avoid using tabs, use only spaces for indenting. Tabs look different in different editors etc., see for example github's diff. A good editor can take care of this automatically.
- The message 'Error : One of the file is zero size while other is not, Exiting with error!!': You're not really exiting, so please put a better error message, and print the line 'fort.10 DIFFERENT!!!' afterwards, as is done when running the normal test and it fails.
- After printing the message 'Warning: Both fort.10.canonical and fort.10.run_pro_SixTrack_4534_cernlib_crlibm_gfortran_O4_100_13601 are zero size', can you still print the message 'fort.10 SAME' afterwards, as is done when running the normal test?
- Minor nitpick: If you use '-f' instead of '-e', you check not only that something with this filename exist, but that it is a normal file (as is expected). http://www.tldp.org/LDP/abs/html/fto.html

kyrsjo:
Thanks Vikas, I've accepted your pull request, updating the master branch in the main fork. You may now update your master branch as described in the email, i.e.:

- In your working copy, add the main repository as a remote. To do this, in the file .git/config in your working copy, add these lines at the end:
[remote "upstream"]
        url = git@github.com:SixTrack/SixTrack
        fetch = +refs/heads/*:refs/remotes/upstream/*
- Fetch the contents of upstream:
git fetch upstream
- Merge it onto your master:
git checkout master
git merge upstream/master
Note that if you are using an old version of GIT, the last command may be slightly different. Please let me know exactly what it was, as I'm writing a little guide...
- Push your updated master to your fork on github:
git push

PR #68, opened 2016-05-19
Revert "BDEX: Beam Distribution EXchange"

rdemaria:
Reverts SixTrack/SixTrack#67

PR #67, opened 2016-05-19
BDEX: Beam Distribution EXchange

kyrsjo:
As described in issue #59 , this patch allows exchanging the particle distribution at some element.

TODO:
- Binary format in order to avoid roundoff errors, or at least CRLIBM (if not already there?)
- Documentation
- More than just thin6d tracking (this could also be implemented in a later patch)?
- Testing...

Related development: DYNK-PIPEFUN, pull request #66 .

PR #71, opened 2016-05-20
Fix nagdummy so that "./make_six da -naglib" (ifort) works.

kyrsjo:
It would be nice to cleanup the linker mess in make_six for da + gfortran also.

I'll accept this PR once the mess in master that was caused by #67 and #68 has been cleaned up, and possibly also the make_six linker mess.


kyrsjo:
Also fixed: Hardcoded library paths in make_six
Added: Examples for DA.

PR #75, opened 2016-05-30
Options in make

kyrsjo:
Have the makefile process option dependencies/incompatibilities.


kyrsjo:
(in progress)

Idea to be implemented:
1. First, check if the ADD are conflicting a default, if so, remove the default
2. create the (preliminary) options list so that in case some common ADD is conflicting with a default, one can just ADD this option and the script will take care of it
3. loop until convergence (or loop counter kills it), adding options required for the current options
4. once the option list is finished, check if any of the selected options are conflicting; if so print error and exit

This should provide the actual options list, which is then used to setup compilers and generate ast files.

kyrsjo:
Closing, made redundant by new branch; see #76 .

PR #76, opened 2016-06-02
Fix makefile

kyrsjo:
Making the option parser for the makefile work.

This makes #75 redundant.


rdemaria:
I am a bit concerned about portability. Can you make an early test in windows?

kyrsjo:
With the shell scripts you mean? If we can give Vikas access to that strange machine of Laurent... He's still waiting for his account AFAIK, but it should be any day now (his name is in the phonebook).

It may anyway be reviewed how the astuce `sed` expressions are used; as it is it's a bit of a mess. Maybe it would be better to run sed once per command and per flag/deck to remove, as is done in make_six? At least that is readable / debugable...

rdemaria:
The windows machine can be accessed with CERN credentials.

I think the target should be:
- implement and document the correct logic.
- reduce as much as possible the dependencies

With the previous makefile we had gcc, gfortran, make, sed. Now we added
bash without removing sed. I am tempted to propose Python to replace sed
and bash. Do you see alternatives, possible issues?

kyrsjo:
As long as we are confident that the Python can be used everywhere, this is fine with me. It was also the reason why I insisted that the depsolver should be a clearly delineated piece of code running a clearly described algorithm, to make modifications or switching it out easier.

It may have the benefit of making the code easier to read as well...

@vikasnt , what do you think?

kyrsjo:
On the other hand, as @vikasnt points out: Don't we anyway need a shell to run make etc.? There are also other shell commands in the makefile, so bash is anyway needed.

SixTrack/makefile
```diff
@@ -40,18 +43,17 @@ ifeq ($(temp),incompatible)
...
 
 # list of valid user options
-USER_OPTIONS := $(sort api beamgas bignblz bnlelens boinc bpm cernlib collimat cpss cr crlibm da debug fast fio g77 g95 gfortran hdf5 ifort lf95 nagfor naglib sse2 sse3 sse41 sse42 tilt tracking vvector hugenblz windows O0 O1 O2 O3 O4)
+USER_OPTIONS := $(sort api beamgas bignblz bnlelens boinc bpm cernlib collimat cpss cr crlibm da debug fast fio g77 g95 gfortran hdf5 ifort lf95 nagfor naglib sse2 sse3 sse41 sse42 tilt tracking vvector hugenblz windows m32 m64 O0 O1 O2 O3 O4)
 
-ALL_OPTIONS := $(sort api beamgas big bignblz bnlelens boinc bpm cernlib collimat cpss cr crlibm da debug fast fio g77 g95 gfortran hdf5 ifort lf95 nagfor naglib small sse2 sse3 tilt tracking vvector windows hugenblz)
+ALL_OPTIONS := $(sort api beamgas big bignblz bnlelens boinc bpm cernlib collimat cpss cr crlibm da debug fast fio g77 g95 gfortran hdf5 ifort lf95 nagfor naglib small sse2 sse3 tilt tracking vvector windows hugenblz m32 m64)
```
kyrsjo: What is the `big` option? And why is `ALL_OPTIONS` and `USER_OPTIONS` sorted differently?

SixTrack/makefile
```diff
@@ -24,19 +24,36 @@ Windows_MS_SDK_PATH    := "C:\Program Files\Microsoft SDKs\Windows\v7.0A\Lib\Use
...
+DEFAULTS := $(filter-out $(REMOVE),$(DEFAULTS))
+# OPTIONS := $(filter-out fio,$(OPTIONS))    # commenting out assuming not needed as of now
+
+#solving dependency for rest options
+SORTED_OPTIONS:= $(shell ./option_resolver.sh "$(DEFAULTS)" "$(ADD)")
+temp=$(shell echo "$(SORTED_OPTIONS)" | grep -o "incompatible" )
+ifeq ($(temp),incompatible)
+   $(error $(SORTED_OPTIONS))
+else
+ OPTIONS := $(SORTED_OPTIONS)
```
kyrsjo: Please be consistent with your indentation - 2 indentations per level is used elsewhere in the file, and that's a good compromise.

kyrsjo:
Starting to look good; could you check ADD, REMOVE, and defaults against ALL_OPTIONS?

@rdemaria : What are the options "big" and "small"?

rdemaria:
related to array size, number of elements and friends?

kyrsjo:
The only appearance I can find for `big` is in `dabnew.ast`, and `small` in `dabnews.ast` (note: not the same file!).

The file "dabnew.s" uses both flags:
```
+if small
      parameter(lda=10000,lst=200000,lea=500,                           &
     &lia=10000,lno=120,lnv=40)
      common /fordes/ nda,ndamaxi
      common / da / cc(lst),eps,epsmac
      common / dai / i1(lst),i2(lst),                                   &
     &ie1(lea),ie2(lea),ieo(lea),ia1(0:lia),ia2(0:lia),ifi(lea),        &
     &idano(lda),idanv(lda),idapo(lda),idalm(lda),idall(lda),           &
     &nst,nomax,nvmax,nmmax,nocut,lfi
      common /factor/ facint(0:lno)
+ei
+if big
      parameter(lda=10000,lst=20050000,lea=100000,                      &
     &lia=5000000,lno=120,lnv=40)
      common /fordes/ nda,ndamaxi
      common / da / cc(lst),eps,epsmac
      common / dai / i1(lst),i2(lst),                                   &
     &ie1(lea),ie2(lea),ieo(lea),ia1(0:lia),ia2(0:lia),ifi(lea),        &
     &idano(lda),idanv(lda),idapo(lda),idalm(lda),idall(lda),           &
     &nst,nomax,nvmax,nmmax,nocut,lfi
      common /factor/ facint(0:lno)
+ei
```

I don't remember when which of these files are used, and for what purpose... The flags are not defined in make_six. And it sounds logical that they should be mutually exclusive?

rdemaria:
It is then related to the memory for non linear maps. I thinl they should
be mutually exclusive

kyrsjo:
One would think that, yeah.

Looking at `make_six`, I see the following:
```if test "$da" = ""
then
    export FORS='sixda sixsc sixscfox dabnew lielib beamgas'
else
    export FORS='track sixve sixvefox dabnews lielib beamgas'
fi```

i.e. for da version make the `dabnew.f`->small, while normal version uses `dabnews.f`->big.

kyrsjo:
So I'm working on the windows bit -- see here:
https://github.com/vikasnt/SixTrack/compare/fix_makefile...kyrsjo:fixm64_windows_XPonoff

The main problem is that MSYS2/CYGWIN makes a mess with the `float.h` header, used in enable_/disable_xp. I've reached out to MSYS for help:
https://sourceforge.net/p/msys2/discussion/general/thread/0b04fcbd/

On the bright side, with the `_controlfp` calls disabled on win64 / 64-bit executable, it compiles passes all tests. Eric reports that he can compile with `_controlfp` on using his setup. Compiling with crlibm OFF also works.

One thing we may want to change in the makefile & option parser, is noticing that `cernlib` is not really possible on Windows (and m64, and mac), since we base this on a pre-built binary. We could implement this by automatically adding the platform as one of the options, which is then passed to the options parser. The options parser could then contain a rule stating that windows and mac conflicts these libraries.

kyrsjo:
Except for the Windows header problem - which is not directly related to the makefile - this thing works. I'm tempted to merge it as-is - maybe include these two changes
https://github.com/vikasnt/SixTrack/commit/2c2be4f05aa8327f2793b728189a7a588a5e09ae
https://github.com/vikasnt/SixTrack/commit/fd5563b8a37e759dc3d80dd059adea7abb062ad1
and merge. I could also easily include these in a separate PR.

The MSYS2 discussion forum seems to be dead. Some information which might be useful for whoever wants to handle this in the future:
https://msdn.microsoft.com/en-us/library/e9b52ceh.aspx

Something else: We could actually consider calling the option_resolver.sh from the old make_six, effectively making them mostly just two interfaces to the same option solver. It would be another branch, another time - but it's still an idea to keep in mind...

PR #82, opened 2016-06-30
Single track file

kyrsjo:
Adding a new option "stf" which writes all the fort.90...(90-31) to a single FORTRAN binary file. This opens the possibility to track more than 64 particles.

Not yet ready to merge, but the PR is useful in order to track progress.


SixTrack/sixtrack.s
```diff
@@ -56185,17 +56316,29 @@ c$$$            endif
...
++ei
++if stf
++if .not.cr
+      subroutine postpr(posi)
++ei
++if cr
+      subroutine postpr(posi,nnuml)
++ei
++ei
++ei
```
kyrsjo: Seems you have one extra +ei here? Check that you're not missing a +if somewhere...

SixTrack/sixtrack.s
```diff
@@ -25492,6 +25495,20 @@ C     Convert r(1), r(2) from U(0,1) -> rvec0 as Gaussian with cutoff mcut (#sig
...
++if cr
+          write(93,*) 'Calling POSTPR nnuml=',nnuml
+          endfile (93,iostat=ierro)
+          backspace (93,iostat=ierro)
+          call postpr(i,nnuml)
++ei
+  70      continue
++ei
         call sumpos
```
kyrsjo: Check if deactivated by astuce / for which combinations of flags it is active

vikasnt: It is actually compiled into sixve.f but the final code is 

    if(ipos.eq.1.and.napx.eq.0) then
      open(10,file='fort.10',form='formatted',status='unknown')
        do 70 i=1,ndafi
          call postpr(91-i)
    70   continue
        call sumpos
        goto 520
      endif

but napx shouldn't be zero since it represents number of particles tracked,which means this code will never run and this is a bug?

kyrsjo:
To be remembered: We are skipping "imad=1", join etc. in postpr. So no join.

This should be picked up and guarded against in DATEN.

kyrsjo:
To recap from discussion - tasks:
 1. Check that the file is written correctly. To do this, write a converter from new format (single file) to the old format. Then compare these converted files with what is produced directly from non-STF SixTrack.
Here SixTest/read90.f90 is useful both as an example code and to check the files you produce. The read90 is called from SixTest/check_90, so be sure to read that script.
Also, having such a converter will anyway be useful, as there are some legacy tools that use the fort.90 etc. files, and we can't rewrite these tools.
 2. Once you are 100% certain that you're writing the file correctly, work on the post-processor which reads the files. This should produce a file fort.10, which should be identical to the one produced by the non-STF version of SixTrack.
 3. When the code modifications are complete, the only thing remaining is to modify the tests. For this, run_pro and run_kill (or more likely, their "slave" scripts try_pro and try_kill) must be modified to copy the new file back, and an equivalent to check_90 must be written.

Once this is done, we can merge!

Uncertainty to be resolved: Exactly how does `ntwin` flag work? In the user manual:
> For the analysis of the Lyapunov exponent it is usually sufficient to store the calculated distance
> of phase space together with the coordinate of the first particle (ntwin set to one). You may
> want to improve the 6D calculation of the distance in phase space with sigcor, dpscor (see 3.6.2)
> when the 6D closed orbit is not calculated with iclo6 6 = 2. If storage space is no problem, one can
> store the coordinates of both particles (ntwin set to two). The distance in phase space is then
> calculated in the post–processing procedure (see 3.5.12). This also allows a subsequent refined
> Lyapunov analysis using differential–algebra and Lie–algebra techniques ([26]).

kyrsjo:
Hi, I see that you're working on the converter, looks good! Just a question, what is unit number 1? I see you're reading some stuff from it, but no open statement.

Another thing: While FORTRAN allows the use of GOTO, and a lot of the old code in SixTrack uses that, today it is generally considered an obsolete style which should be avoided, since it is harder to read and debug. Rather use do/enddo and other normal flow control statements (if, select case, etc.). Only exception is if this code was copy-pasted verbatim from inside SixTrack, then you don't need to rewrite it.

vikasnt:
hey , since fort.90 is also one of the output later , for simplicity i renamed stf output to fort.1 from fort.90.
Also i didn't use open , since it seems to be able to read without it as the file already exists.
yeah i used goto since i am reusing code from sixtrack itself and anyway here it wont confuse (small code).

vikasnt:
One issue is i am getting different size after conversion which should mean different output files !!.
I am trying to understand the reason from sixtrack code ( there were some write statement we neglected which most probably aren't actually used as well ).

I used the fort.2 and fort.3 you provided me earlier with for running binary and comparing.

vikasnt:
Hey @kyrsjo ,
As I mentioned earlier I was getting different sized binary files(when compared with old code) after running the fortran code I wrote to create fort.# files for each pair from stf output( commits from 4 days ago), So I actually changed the way I am comparing.
Now I use both cases output to create a single text file in place of binary(for ease of reading), and then compare it using diff command using a small script. Simply putting the output files ( from both cases) into the respective folder and running the script should work. I am getting same file size for created files(.dat) in each case. ( If you wish to see the text output file , just delete the last rm commands at end in the script).
I have used end_of_file to completely read the binaries from .not.stf, since I am not 100% sure about data in them while binary from stf is written by me So I know how it is written. 
Let me know  if you think this is not enough or the correct method of testing them.

kyrsjo:
You've checked that the sizes are now the same; how about the data contained in the files, is this also now identical?

vikasnt:
Yes, same size and diff command returns nothing which should mean same file(it contains only header and tracking data as we expected)
One thing is there were a few zeros added at end of header in sixtrack for "future use", but when i use that to print it always contains different number. ( so I discarded that part while writing output file)

kyrsjo:
Very good! Diff returning no output means no difference (try it on two files that are different!).

Just to confirm, your fort.90 etc. and new track output checkers,  they both write out everything except for those parts of the header?

If you look at the code in sixtrack writing the strange header, do you find any indications on where those numbers come from?

vikasnt:
In Sixtrack manual , Appendix D, Table D.1 —– 50 additional parameters —– explains the last part where 43 dummy parameters are listed. But I am not sure why  `zero` is used in place of `0` or `dummy.`

vikasnt:
Update: 
As written earlier , we have already written converter to ensure new track file has same data as earlier files.
After completing postpr code, I have compared fort.10 by running test="bb" wth executable from old(not.stf) vs stf. fort.10 for both the case is almost same and the only difference is in the last element of postprocessing data i.e sumda(60) which denotes total cpu time 
`sumda(60)=dble(trtime)`.
Thus i believe postpr code for stf works(at least with the standard options).

As discussed earlier , now we should modify test to work for stf case.
I'll go through the code and try to understand how I can modify it.

kyrsjo:
Hi Vikas, this is great news! Does the other tests in SixTest also succeed (i.e. fort.10) when using STF?

Regarding the use of `zero` instead of just `0`: This constant is equal to the literal 0.0d0, which makes sure that an actual double precision floating point value, not relying on the compiler to "promote" a "0" or a "0.0" to the correct type. When writing a file such as is done here, this may actually be necessary, since the binary format of each value written is implicit in the data types of the written values, and "0" and "0.0d0" corresponds to different data types which will be represented differently in the file (one as an integer, the other as a double precision floating point).

SixTest/stf_output/compare.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+cd stf/
+gfortran new_converter.f
```
kyrsjo: FYI: When running the compilers like gfortran or gcc, use the `-o filename` flag to set the name of the executable, so it gets called something more informative than `a.out`. Also use the `-Wall` flag.

SixTest/stf_output/not_stf/old_converter.f
```diff
@@ -0,0 +1,41 @@
+      program old_converter
```
kyrsjo: A comment describing what the program does (converting non-STF format fort.90 files to ASCII), who wrote it and when would be very useful.

SixTest/stf_output/compare.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
kyrsjo: If you could get this program to accept two command line arguments - the STF format and the non-STF format file to compare - that would be useful! It could then be used to compare the outputs of any tests by comparing the new and the old canonical. Will be very useful whenever we need to modify canonicals (after fixing bugs etc.).

SixTest/stf_output/stf/new_converter.f
```diff
@@ -0,0 +1,80 @@
+      program new_converter
```
kyrsjo: In addition to adding the descriptive comment mentioned in the `old_converter.f`:

For the new converter, having an optional command line argument specifying which particle(s) to extract - right now it's only ever pair 1 if I understand your code correctly - would be very useful. For the old converter, maybe just use the filename as an argument?

I would also propose to name your output files as something like "singletrackfile.dat.PAIRNUMBER" (same goes for those extracted from the old format).

kyrsjo:
Regarding the tests:

I think the best way to modify those would be to add an extra canonical "singletrackfile.dat.canoncial" (or whatever the new file is called) alongside the fort.90.canonical. Then add a script "check_singletrackfile" which works the same way as "check_90", comparing the newly produced file with the canonical. You then have to add singletrackfile.dat to the list of files to copy back when running run_pro and run_kill.

Having a way of comparing the outputs of fort.90 from non-stf and fort.90 extracted from the STF file would also be very useful - I think you have written 90% of that already.

vikasnt:
Testing stf version of sixtrack: 

1.check_stf90 works exactly the same way as check_90 but it will compare fort.90(from stf sixtrack run) with fort.90.stf_canonical in place of fort.90 (from not.stf sixtrack run) with fort.90.canonical.
(Although user will have to ensure that they use stf sixtrack binary when running run_pro)

2.As for try_pro and try_kill, since we will still use fort.90 as the output file , and it is already being copied in these scripts,I don't think we actually need to edit these scripts for stf version.

3.I was having some issues while comparing output for stf vs non-stf version .Firstly i had assumed that turn number will always change by 1( ie. 1,2,3) but while running test "bb" ,I can see that turn number varies by 10 ( ie. 10,20,30) in case of bb. This means i will have to slightly change the implementation and this time i will try to keep the code free of any other assumption by my side.
Secondly, header contains time and date , so if i run stf and non-stf version at different time, the "diff" command will always say files are different. So should i just skip header data for comparing output ( and only compare tracking data) ?

Also , I understand that we will need to generate fort.90.stf_canonical files for all tests.(I think we can do this after we have ensured that the output comparing script is properly working and then if output for stf vs non-stf is same, we can use the stf fort.90 as fort.90.stf_canonical for future testing).

kyrsjo:
1&2) Great! Only thing, it may be a good idea to name the new file something else than fort.90, in order to avoid confusion, even it is probably best to keep using unit 90.

Something else, when compiled with stf, is this included in the name of the binary? Having stf in the binary name would be useful.

3) Hmm, interesting - is output every 10 turns only true in the case of BB, or is it generally true? Which variable controls how often to output?

Please compare header data, just skip the time and date.

vikasnt:
1) I have used name fort_stf ( or tell me whatever suits better).
2) Having stf in the binary name would be useful.--> I believe that should be doable by small changes in makefile ,(I'll do it after other things realted to stf are cleaned up).
3) Yeah, i was also intrigued by the issue when i found it. I'll look into it and put it in a separate comment

kyrsjo:
1) The fort.something names are fortran's autogenerated names. I would call it "singletrackfile.dat" - that will make it obvious what it contains (unlike the fort.NN names).
2) Great, I also think so. We also still need to maintain make_six...
3) Riccardo, any idea here? @rdemaria

kyrsjo:
`boincrf` is only active if boinc flag is active. Testing of this must be discussed a bit further in the group here, I think we can leave this and merge a non-cr and non-boinc version for now (and mark `stf` as incompatible with `cr` and `boinc` flags). 

I find that such a subroutine is defined in the `myboinc.f` file; here it is simply returning the first argument ("myname") to the second ("filename"). So when you call `boincrf("fort.10",filename)`, then `filename` will become "fort.10". `filename` is then used for opening the file. My guess is that when BOIC is actually active (i.e. not just the dummy functions in myboinc.f), it will return the "local version" of the file normally called "fort.10" etc. However I couldn't find the docs...

SixTrack/sixtrack.s
```diff
@@ -9684,13 +9685,23 @@ cc2008
       open(89,file='fort.89',form='unformatted',status='unknown')
 +ei
 +if boinc
-      call boincrf('fort.90',filename)
+      call boincrf('fort,90',filename)
```
kyrsjo: Uhm, that should probably still be a `.`, not a `,`.

kyrsjo:
Hi Vikas, could you please merge the latest upstream/master into this branch?

kyrsjo:
Perfect, thanks!

kyrsjo:
Hi, I created a PR to track some changes I'm doing while reviewing your code.
https://github.com/vikasnt/SixTrack/pull/4

kyrsjo:
Regarding the sparse output (not every turn) for test "bb", I found this in the manual:

`nwr(3) Every nwr(3)’th turn at the flat top a write out of the coordinates on unit 90 (and lower)
will occur. For constant energy this number controls the amount of data on unit 90 (and lower),
as the particles are considered on the flat top.`

kyrsjo:
For the TODO list: make_six must know about the STF flag, otherwise it will be enabled by default and impossible to switch off without manually editing the ASTUCE.

Also, AFAIK the number of headers written for ntwin=1 and 2 should be the same (it's just that for ntwin=1 the records written in `writebin` are half as long.

vikasnt:
Although i haven't had the chance for running ntwin =1 case, but consider this. In the old case (not-stf) each file has *one* header and some tracking data. 
When ntwin=2, every file has tracking data for two particles, so 32 particles would require only 16 files But ntwin=1 means every file has tracking data for one particle only. SO for 32 particles we need 32 files
Now this also means header for same number of particle should be double in ntwin=1 case, since each file has one header.

kyrsjo:
As far as I understood the writing code, in the case of `ntwin=1` it writes 1 header with first particle=last particle, then writes one record per turn containing one particle. This means that only particle 1,3,5,etc. is actually written to the file, along with the "distance in angular phase-space" `dam`.

In the case of `ntwin=2` it still writes one header per pair (with `first particle = ia`, `last particle = ia+1`), but each record written during tracking (in `writebin`) now contains two particles.

The number of records are therefore the same for `ntwin=1` and 2, it's just that one field in the header changes, and that the length of the tracking output records contains the position of 1 or 2 particles.

vikasnt:
Okay, I think I saw this sort of behavior in one of the tests "exact" earlier(2-3 days ago while working on comparing output for tests), It had one header and even though max number of particles was 2, there was only tracking data of particle 1 but not for 2nd particle.(The header was fitting the format of ntwin=1i.e first particle=last particle).
I was gonna look into this after the GsoC Submission deadline,but I don't understand why we would want to willingly discard tracking data of half of the particles. Anyway, I think you are right about this and I'll check again if the code is still working as it should in stf, ntwin=1 case.

kyrsjo:
Yeah, it's unfortunate that I didn't have the time to look into this closely before now. The reason for discarding half of the data is to save disk space - the distance in phase space between each pair of particles is what is used in the analysis anyways, so you don't really *need* the full thing for most analyses.

I'm trying to fix this little detail now + adding some more comments and markers as I'm going along - otherwise the new code looks good. I need to rush it a bit, since I will need to have multi-particle tracking working before the week after the next.

Thanks for pointing out that the "exact" test uses `ntwin=1`!

Please let me know when your GSOC report is ready!

SixTest/check_stf
```diff
@@ -0,0 +1,91 @@
...
+      echo "      CASE $i"
+      echo "      ===================================="
+      if test -s $i/singletrackfile.dat.$prov
+      then 
+          if test -s $i/singletrackfile.dat.$v
+          then
+              mkdir -p tmp$$
+              cd tmp$$
+              cp ../$i/singletrackfile.dat.$prov fort.190
+              $BIN/read90 > fort.20
```
kyrsjo: Hmm, there should probably be a special version of read90 which reads all of the new files...

kyrsjo:
Thanks for pulling. I made a few further modifications to the testing framework and ASCII converter, it confirms that everything seems to be working smoothly. Once the tests are finished, I'll add the `singletrackfile.dat.canonical` files for all tests.

What remains here is then really only CR (might be a painfull mess -> I know that part of the code and will do it) and mentioning the new data format in the user_manual appendix D "Data structure of the Data-Files".

kyrsjo:
To be merged as soon as Vikas pulls the bugfixes from https://github.com/vikasnt/SixTrack/pull/5 into this branch.

Next step: Increasing the number of particles :)

Note: I had some problems testing. Apparently, the CR version + btrfs were not friends.

kyrsjo:
Thanks! I noted the BTRFS bug here: #97

PR #88, opened 2016-07-13
Fixing DYNK for thick tracking

KFubuki:
Checking simultaneously that the element's attribute is zero and DYNK is not used in the trauthck subroutine.


kyrsjo:
The modification looks fine, but please only commit only what you need. Thanks for closing it, I'm waiting for the next PR :)

kyrsjo:
(the next PR was #89  )

Issue #77, opened 2016-06-07
Number of particles per tracking limited to 64


kyrsjo:
The number of particles per tracking round is currently limited to 64 particles only.

Problems:
- Output files fort.91-i, where i=1..32
- Big matrices `al`, `as`, `hv`, and `bl1v`

The first issue to tackle, then merge, is the binary output files. These are written in `program maincr` and `subroutine writebin`. They are read in `subroutine join`, `subroutine crcheck`, and `subroutine postpr`. Additionally, they are used in the external program SUSSIX.

kyrsjo:
Also worth watching out for is the arrays `binrecs` and `crbinrecs`.

kyrsjo:
Also becoming important for https://github.com/apsimonr/matlab-beam-loading-llrf

kyrsjo:
For the output files: In order to not break compatibility for already-existing postprocessing routines, let's make the new output an optional feature, and leave the old code alone. For this, the cleanest way is probably to have an astuce flag "singlefiles".

Plan:
* New `make` / `make_six` / `astuce` flag "`singletrackfile`"
  -> Wrap the old code in `+if .not.singletrackfile / +ei` and the new in `+if singletrackfile / +ei`
* For the new output, use unit 90, since we know this is now free...
* For the new output, call the file "`trackout.bin`" to avoid collision with old output in checks etc.

Then, once we have input/output and postprocessing working with both the new and old way, we need to build a converter (a standalone program) which takes `trackout.bin`, reads the first 64 particles, and writes `fort.(90..90-31)`. This is vital for testing the new code using the `check_90`.

Once this is done, we can add another check script `check_trackout` that works in much the same way as `check_90`: Reads `trackout.bin.runName` and comparing it to `trackout.bin.canonical` while ignoring not-important metadata such as time/date etc.

Once this is done, tested, and merged, we can move on with actually increasing the particle limit.

For the header of the new file, we must check if the headers of the old files can ever be different - they contain the arrays with the tunes (`qwcs`), 6d closed orbit (`clo6v` / `clop6v`), dispersion vector (`di0xs` etc.), transfer map (`tas`), and seed - and these all depend on the particle number `ia`. But they may all be identical? If they are, we can essentially keep the same format as before. If not, we should write the header once per pair in essentially the same format as before, then write the particle outputs, again in essentially the same format as before. One thing to consider is to open the file as a "stream", since this will produce a file that is compiler-independent and thus easier to handle with external tools.

[...]

PR #98, opened 2016-08-30
New "bignpart" option

kyrsjo:
This adds a new astuce flag "bignpart", which allows tracking up to 2048 particles in the same simulation. Initial testing indicates that it works well.

Increasing the particle number above 2048 appears to lead to problems with memory size and linking the executable (at least in 32-bit version). With `npart=2048` we use about 700 MB of memory which is acceptable for many things, however it could be reduced by not allocating all the big matrices unless actually needed.

TODO:
 * More testing -- run tests where all particles should be identical etc.
 * Distribution generation

See also issue #77 .


kyrsjo:
Distribution generation postponed to another PR -- see issue #100 . Standard tests and new test for identical tracking for all particles are also OK.

Ideally, we would like to go above 2048 particles/run (especially for collimation), but this would require some architectural changes (making the huge static arrays non-static). It would also be nice to profile & optimize the postprocessing routine better as it seems quite slow, but this can also be done in a new branch.

PR #20, opened 2015-10-02
wire element was modified

anpatape:

PR #99, opened 2016-08-30
Wire implementation

rdemaria:
Do not merge yet. I am trying to recall the status of the wire...


kyrsjo:
Almost-duplicate of #20, which has one more commit and a cleaner history.

PR #101, opened 2016-08-31
Fix form fortran fix

kyrsjo:
Undo mistake in code formatting from c31a07445b6a7dd78916dbbab6a23a13879e7434 and update canonicals.


kyrsjo:
Hmm. I should probably have updated `SixTest/elensidealthin6d/fort.10` as well...

PR #102, opened 2016-09-01
DYNK for E-Lens

jfwagner:


kyrsjo:
The TODO:

In sixtrack.s:
 * Add the ELENS as a SET command -- Modify subroutines `dynk_setvalue` and `dynk_getvalue`, possibly `initialize_element` in case one setted variable should affect multiple "behind the scenes" variables (not currently the case in ELENS)
* Add the REPEAT as a FUN command -- Modify subroutines `dynk_parseFUN` and `dynk_computeFUN`.

Then, update Doc/user_manual/six.tex, and add test(s) in SixTest.

kyrsjo:
The GET/SET support for ELENS appears to be working; need to check tracking. The python script from Miriam is probably a good starting point.

Once this is done, we'll merge and create a new branch for the new FUNs (possibly one branch per 
FUN to make testing and reviewing easier).

kyrsjo:
Also TODO: Update the documentation Doc/user_manual/.

kyrsjo:
Argument from command line in Python:
```
import sys

#assert len(sys.argv) == 2
print sys.argv

oldname = "hi"
newname = oldname + "." + sys.argv[1]
print newname
```

Also, please (1) import everything, even numpy; and (2) no `from xyz import *`, rather use e.g. `import numpy as np`.

kyrsjo:
Joschka, an idea for testing the SET/GET methods: You actually don't really need to check that the kick is matching the physics (this Miriam already did), you really only need to check that the effect you get when using the SET method is the same as when setting it in fort.3.

The easy way to do this is to first run it without DYNK to get canonical tracking data. Then in the elens block "mess up" the values and use DYNK to set it back, and run it again. If everything worked, you should get the exact same tracking results.

kyrsjo:
OK, "mess-it-up & reset then diff to un-messed-up" check is passed -> effect on beam is OK.
Manual is OK, thanks for fixes to physics manual. Tests lacking canonicals, but OK for now since we plan to add more DYNK FUNs for use together with ELENS.

Code review PASS :)

kyrsjo:
Joscka: If you decide to not do the new FUNs (uniform RAND and every-nth-turn-filter), you need to fix the canonicals for the test - the plan was that it would test both the ELENS SET and the new FUNs, but if this plan changes, the work should still be finished.

PR #103, opened 2016-09-02
Move big matrices out of common blocs and into module

kyrsjo:
This should enable taking more memory and thus tracking even more particles, without any significant architectural changes or using pointers.

Implemented as optional feaature "datamods".


kyrsjo:
(some debugging + makefile support still todo, but this version passes tests crabamp, lost, thick4 -- I checked these because they are quick and thick4 has thick tracking)

kyrsjo:
Other huge arrays, this time in FMA:
```
      double precision, dimension(npart,fma_nturn_max,6) ::
     &xyzv,nxyzv ! phase space (x,x',y,y',z,dE/E) [mm,mrad,mm,mrad,mm,1.e-3], normalized phase space variables [sqrt(m) 1.e-3]
      double precision, dimension(npart,fma_nturn_max,3) ::
     &epsnxyzv ! normalized emittances
      integer, dimension(npart,fma_nturn_max) :: turn ! npart = max. number of particles
```
 Miriam @mfittere , any chance you could make these a bit smaller? They seem to be the next thing blowing up the compilation. For example by reading one particle, then processing, then reading the next one etc...?

Since it's a local variable, I may be able to do some things tough...

kyrsjo:
OK, so now fma isn't blocking anymore... However reading the code for loading the dumpfiles makes me a bit worried:
```
            do k=1,fma_nturn(i) !loop over turns
              do l=1,napx !loop over particles
+if .not.crlibm
                read(dumpunit(j),*,iostat=ierro) id,turn(l,k),pos,      &
     &xyzvdummy(1),xyzvdummy(2),xyzvdummy(3),xyzvdummy(4),xyzvdummy(5),
     &xyzvdummy(6),kt
```
Here the file is assumed to always have an unchanging structure... What happens if a particle is lost? Could you at least add a check that the turn and particle ID read from the file is the expected one?

kyrsjo:
Still some problems with the makefile and Makefile_makesix(.template). It works via a hack: make will usually compile dependencies in the order they are listed; at least if `-j` is just 1, as is default... So some better targeting would be very very useful.

Otherwise - it appears to work great. I'll add the "hugenpart" option (which can probably go higher than what even collimation needs, without any memory problems, as long as we're not doing thick tracking), then bump the version number and release! Running tests now.

rdemaria:
Make -jn should work. What has changed in dependencies? astuce was always a
prerequisite for .f

kyrsjo:
Hi Riccardo,

the issue is that some of the files need the file bigmats.mod file, which is built when building datamods.f -> datamods.o. However, to build datamods.o you need astuce.

Oh, and bigmats.mod will only exist (and be used) if the flag "datamods" is true.

So some more explicit dependencies would be nice, without having to rewrite to much stuff...

kyrsjo:
OK, I've confirmed that it works for ifort also. The make_six should be good, standalone makefile is still a bit hack-ish but I think releasable. I wouldn't mind someone making a PR to fix it properly tough...

@rdemaria : Could you take a quick look at the code? If it looks OK to you, I will bump the version number and date, and merge it tomorrow morning.

PR #105, opened 2016-09-26
Using the Merlin scattering routines inside SixTrack/collimation

kyrsjo:
This links the MERLIN scattering routines, written in C++, with SixTrack. The Merlin files are kept in the `merlinscatter` subfolder, and is (or will soon be) built to an `.a` similar to `crlibm`.

The inclusion of these routines are controlled with the `merlinscatter` flag to make_six etc.

The link to SixTrack is done through the `merlinscatter/merlin_interface.cpp` file, which defines a set of interface functions compiled with `extern "C"` to make linking from FORTRAN possible.


JamesMolson:
Hmm, well I have it working now. Our very expensive configuration/pre-calculation routine currently lives in scatin() which seems to get called every time a collimator is encountered. It makes the code a bit too slow. I see there is a firstrun flag earlier on in the code where file headers get written, so I might move the configuration to around there, unless you can think of a better place to put it all? It will need access to the beam energy at this stage.

JamesMolson:
Actually, I suppose I could make some internal flags to keep track of things...

kyrsjo:
There is a counter for the current sample! I believe it is simply the global variable `samplenumber`?

kyrsjo:
Internal flags (a `static bool isinitialized`) on your side of the code should work too... Or just check if your object is a `NULL` pointer, if it is then initalize it.

The `firstrun` flag should work too. There is not a lack of variables which contain rougly the same information and are hopefully synchronized.

By the way: Does the collimation code currently always do this "very expensive initialization" every time we get to a collimator? Any chance we could not do that?

kyrsjo:
I looked (quickly) through the code, and notice that you chose to include the MERLIN RNG; and then there is also some calls to . That's OK with me, however how are you initializing the seed?

Regarding the code around this comment:
https://github.com/JamesMolson/SixTrack/commit/5bba141f3377b45a00d4af5f75bdeaacb919c002#commitcomment-19222261
What are you actually using the `getran` for here?

JamesMolson:
The ionisation routine isn't in there yet, it will be quite complicated because I need lots more material properties in order to calculate it correctly. I could either add to the sixtrack material properties, or move the whole merlin material implementation over. I just let the if/else definition in there for now because that is going to need a new function call, yes.


The RNG gets seeded as:

RandomNG::init(0);

The 0 there replaced by whatever it should be seeded with.


The gettran gives you the momentum transfer, which then gets changed into the scattering angle in another routine. Since I can't label the sixtrack.s on here since its too big:


      t = gettran(inter,mat,p)
!
!++  Tetat calculates from the rms transverse momentum transfer in
!++  monte-carlo fashion the angle changes for x and z planes. The
!++  angle change is proportional to SQRT(t) and 1/p, as expected.
!
      call tetat(t,p,dxp,dzp)
!
!++  Apply angle changes
!
      xp=xp+dxp
      zp=zp+dzp

there is where it gets used^^

JamesMolson:
Also, how many spaces do you want for each level of indentation?

kyrsjo:
Normally we use 3 spaces, but there are various standards in use in the code... Please just use the same as whatever code surrounds the code you are modifying, either 2 or 3 and no tabs.

I now see that in this case, you are using a scheme where the 'else if's and 'else's are indented more than the 'if' and 'end if', I haven't seen that one anywhere else in the code really... As long as you keep inside a single function (maybe it was already there?), it's fine.

I can confirm that the code compiles successfully with `./make_six gfortran collimat merlinscatter`!

Regarding RNG initialization, it would be very useful if it could be tied to the current collimation RNG seed. I think the variable is simply named `rnd_seed`, where the value 0 uses the current time to initialize.

I now see the difference; you've changed gettran *function* in SixTrack. I was confused with what looked like passing the variable gettran as an argument to Merlin, not realizing that you were doing return-through-function-argument!

Regarding the materials properties: I'm not sure what would be better - maybe just keep it in Merlin and use the material name to setup whatever needs to be done on the Merlin side? Alessio / Roderik / Stefano can maybe be helpful here...

JamesMolson:
hm, the indentation around that code is a bit of a mess in general then. I'll see if I can push the RNG seed over too.

kyrsjo:
Yeah, just keep whatever is there, as long as it is consistent it doesn't really matter. In the diff view I only saw the first `if` then the `elseif` which was differently indented...

kyrsjo:
And yes, please push the RNG seed over, we want simulations to be repeatable (with the same seed), and that changing the seed should give you a new piece of statistics.

SixTrack/sixtrack.s
```diff
@@ -67069,6 +67091,51 @@ c$$$     &           myalphay * cos(phiy))
...
+      PartB = AtomicMass * 1e-6
+      CalcElectronDensity = PartA/PartB
+      return
+      end
+
+! Function to calculate the plasma energy in a material
+! CalculatePlasmaEnergy = (PlanckConstantBar * sqrt((ElectronDensity *(ElectronCharge**2)) / (ElectronMass * FreeSpacePermittivity)))/ElectronCharge*eV;
+      function CalcPlasmaEnergy(ElectronDensity)
+      implicit none
+      double precision ElectronDensity
```
kyrsjo: Thank you for the implicit none. Note that it may be possible to make the constants actual constants (parameters)?

JamesMolson: Yes, it shouldn't be a problem.

JamesMolson:
We had stupid output from the SD class, and it turns out I had been calling the get angle function instead of the get mass one for the energy offset.

Ooops. Thanks to Ye for spotting this.

Should be fixed now.

JamesMolson:
@rbruce1 @amereghe Are there any objections to this now?

kyrsjo:
Some fixes after discussing with Sam Tygier and comparing with upstream MERLIN: https://github.com/JamesMolson/SixTrack/pull/6

TODO:
 * Check the interface
 * Check the modifications to sixtrack.s
 * Rerun quick test (single collimator and dump)
 * Bump version number

amereghe:
Hi all,
thanks a lot for the intense work. Still, before pressing merge, I think it would be better to wait for Ye to have meaningful results. Moreover, I would structure the code so that the choice of the MSM is done not at compilation level, but at input parsing (eg we add an optional line in the COLL block of fort.3 for triggering the request of the MSM). While the latter is closer to a problem of style in factoring the code, the former is more substantial.
Thanks,
Cheers

kyrsjo:
Thanks Alessio,
regarding the compilation switch, I think it is useful to retain the possibility of compiling SixTrack+collimation without MERLIN, in case of cranky C++ compilers etc.; and while such a feature would be nice to have, *I don't think it's worth holding up the merge even further*. If this is truly a requirement, an interested party could easily add this feature later.

Regarding Ye's checks, James and I spoke with him this morning, and he said that he had rerun them, successfully this time.

Best regards,
Kyrre

kyrsjo:
Especially, the LXPLUS default compilers do NOT fully support C++11, and are therefore in the "crancky" category. So if we make it impossible to compile SixTrack's collimation version without MERLIN, everyone would need to know how to switch compilers.

kyrsjo:
Code checks finished, as soon as the quick tests have passed once again after the recent small modifications, you can merge.

Please bump the last digit of the version number :)

PR #106, opened 2016-10-25
Fix manual for Beam-Beam

kyrsjo:


kyrsjo:
Riccardo, agree and merge?

kyrsjo:
Merge? Or wait until we have the code modification also?

Issue #36, opened 2016-01-13
Maintain a new executable with large number of element for fcc


rdemaria:

kyrsjo:
Duplicated #31 ?

Issue #35, opened 2016-01-11
Combined function curvature effects


rdemaria:
Update multipole block.

Issue #34, opened 2016-01-06
Auto-assignment of output unit numbers


kyrsjo:
In many parts of SixTrack, output to (or input from) named files, their number and filename specified directly or implicitly through input files, are required. This means that it is often up to the user to specify the fortran file unit number.

This could be avoided if
1. There was a "known safe" range of unit numbers to use
2. There was a mechanism in sixtrack to automatically allocate one of these safe unit numbers, keeping track of which units have already been allocated, and which units are currently free.

amereghe:
I think that a middle step in this direction could be the creation of one common (or more) which parametrises official units (e.g. "lunout" could store the main output unit of 6T), thus blocking unit numbers to the users (i.e. the users should not use them for their own I/O).

kyrsjo:
The real problem is when you have user-specified input/output files, such as for DUMP and DYNK. Output files are especially problematic, since they need to be kept open.

A simple way to implement this would be to have an array of ~1000 booleans, all initialized to false. Each of these bools would correspond to one unit number, "false" meaning the corresponding unit number is free, and "true" meaning that it is free. This array could then be accessed using two subroutines: ```requestUnit(int returnUnit)``` and ```releaseUnit(int unitToRelease```. These would allocate a unit from the pool (and mark it as "taken"), and release a unit from the pool (and mark it as free). Additionally, it could check that the unit is not already in use before allocating it (exiting with an error if it is), and check that it has actually been freed (the file is closed) before accepting a release (exiting with an error if not).

This should be very easy to implement, if we can find a "fore-sure safe" unit range.

kyrsjo:
But one releated thing to what you are saying: An official list of unit numbers (which is more up2date than the manual) would certainly be a great thing. Especially if we could use "lout" instead of "*" in the non-cr version (it would need to have a different value); I'm fed up with having to type in every error message twice!

kyrsjo:
LOUT everywhere is done in Pull request #233 !

JamesMolson:
Auto assignment via #427 and #430 


Needs to be applied to most of the code

rdemaria:
We close the issue since, the module is used for everything but the standard fort.<90 files. The module does automatic closing of unit till 999.

Issue #33, opened 2015-12-16
make_six: collimat and crlibm


JamesMolson:
make_six allows both the crlibm and collimat options to be combined.

This causes a segfault when running sixtrack.

```
#0  0x838D3E3 in _gfortran_backtrace
#1  0x838C76C in .L3 at compile_options.o:?
#2  0xF7FFDB6F
#3  0x8380A1D in do_sub at addition_scs.c:575 (discriminator 3)
#4  0x8383C49 in sine at sine.c:52
#5  0x8383CC2 in scs_sin_rn at sine.c:80
#6  0xFFFFFFFF

Program received signal SIGSEGV, Segmentation fault.
do_sub (result=0xffff9240, x=0xffff9240, y=0xffff9160) at addition_scs.c:575
575       for(j=0; i<SCS_NB_WORDS; i++,j++)    R_HW[j] = res[i];
```

kyrsjo:
The problem is that collimation does not (currently) work with crlibm. The combination of these options should be picked up by make_six, issuing the user an error, as is today done with collimation and cr (checkpoint/restart).

Why it causes a segfault is interesting in itself. It might be useful to run valgrind to see if the initial problem occurs in collimat, or if you just uncovered a symptom of something else. Could you provide an example input causing a segfault?

JamesMolson:
make_six with gfortran collimat crlibm bignblz (using your branch)

The input files attached to #31 will kill it for me.

kyrsjo:
Which version of gfortran and gcc?

JamesMolson:
$ gfortran --version
GNU Fortran (GCC) 5.3.1 20151207 (Red Hat 5.3.1-2)

JamesMolson:
-----------------------------------------------------------------------------------------------------------------------------------

         OOOOOOOOOOOOOOOO
         OO            OO
         OO  TRACKING  OO
         OO            OO
         OOOOOOOOOOOOOOOO

-----------------------------------------------------------------------------------------------------------------------------------



          -------------------------------

           Program      C O L L T R A C K 

             R. Assmann           -    AB/ABP
             C. Bracco            -    AB/ABP
             V. Previtali         -    AB/ABP
             S. Redaelli          -    AB/OP
             G. Robert-Demolaize  -    BNL
             A. Rossi             -    AB/ABP
             T. Weiler            -    IEKP
                  CERN 2001 - 2009

          -------------------------------



 Collimation version of Sixtrack running... 08/2009

[...]

PR #107, opened 2016-10-27
Building Sixtrack using cmake

JamesMolson:
Here is my first attempt at a CMakeLists.txt which contains a lot of extra fluff.

It will build sixtrack with the default options currently without issue.

Make a new folder somewhere then run
cmake /path/to/sixtrack/folder/

If you have the ccmake installed you can pick options this way.

then just run make -j (yes it can build in parallel now).

There is still a lot of work to do. I've not put checks in at all for which options exclude others, neither have I tested anything that isn't the default configuration. It is somewhat ugly as well internally due to all the astuce/dafor steps required to generate the source.


rdemaria:
Great!

For me I can merge since it does not collide with anything, unless you prefer to discuss here.

Some questions.

Do you generate a target name? Do you keep intermediate file like the real fortran  files? Does it include aslo crlibm?

JamesMolson:
It still doesn't support everything that make_six does, but should do eventually.
I generate the target as the Sixtrack binary name, so the target is called ${SIXTRACK_BINARY_NAME}.

I've not bothered to clear out any of the intermediate files yet, so everything should be kept.
crlibm should work fine. I think I've passed the correct set of flags over to it, but that is of course quite trivial to build.

Now it should complain if incompatible sets of build options are selected.

I've not yet fixed linking in other external libraries, e.g. the cernlib/naglib stuff.
It should be possible to do all the option filtering on the .ast files using the cmake internal regex engine (or whatever it uses), so the external sed requirement could also be dropped. I've yet to be able to get sed to work nicely with the cmake calls, so what is in now isn't quite the same as in make_six.

Also it should be possible to use cmake to generate build files on osx, windows, or anything else it supports.

Some other points:

There is a now an option to select either 32 or a 64 bit build.

Since I based this file off other one I had made, there is support for building with MPI, CUDA and OpenMP, but right now those options just do nothing (and I don't see much harm in leaving them there for any future cases).
To select the fotran compiler you need to have it exported in the FC environment variable, for example:

FC=ifort cmake /bla/bla/bla

I tested that on lxplus and it seems to work. So all the compiler options you can pick are now a little redundant.



There is still more to be done - with some options it fails to build.

rdemaria:
'sed' has always been a mess also going to windows. If cmake has something
more robust and portable i would go for it

JamesMolson:
If you look into the CMakeLists.txt, there is a section where the options of what to build are chosen, for example:

OPTION(TILT "??" ON)
OPTION(FAST "??" ON)
OPTION(CRLIBM "Constant rounding library support" ON)
OPTION(CERNLIB "??" ON)
OPTION(NAGLIB "??" ON)

Most of these I've either had a guess at what they do, or I don't know.

Would it be possible for somebody that knows what each one does to provide a short text description of each setting?

JamesMolson:
Okay, and that should remove any requirement for sed, so with the correct generator it might work nicely on windows? I don't have any windows machines to test.

kyrsjo:
Flags and their descriptions:
 * TILT: Allow elements to be tilted (by error)
* FAST: Which implementation of drifts to use in thin6d
* CRLIBM: Use correctly rounded libmath instead of system libmath
* CERNLIB: Link to external CERNLIB library for PAW plots. Otherwise use internally defined dummy functions from deck plotdumy.
* NAGLIB: Link to external NAGLIB library, used in subroutine `coruglo` and `coruord` for DA version.
 
* API: Link to external BOINC API
* DA: Build differential algebra version (NOT dynamic aperture!)
* COLLIMAT: Collimation code
* CPSS: Something related to BOINC??
* BOINC: BOINC distributed computing support 
* CR: Checkpoint / restarting
* NAGFOR: Use NAGFOR compiler
* BPM:Activates the collection of tracking data for sussix (somewhat deprecated, should be replaced by DUMP)
* BEAMGAS: Beam-gas scattering
* BNLELENS: Support for the BNL electron lens (deprecated, should be replaced by the ELENS block in fort.3)
* BIGNBLZ: Allow up to 200'000 structure elements (normally 20'000 (15k for collimat)) - useful for FCC etc.
* HUGENBLZ: Allow up to 400'000 structure elements (normally 20'000 (15k for collimat)) - useful for FCC etc.
* STF: Single Track File - write all tracks for postprocessing to fort.90, instead of 32 separate files fort.90 to fort.(90-32). Required for increased particle numbers
* BIGNPART: Allow tracking up to 2048 particles (normally 64).
* DATAMODS: Allocate big matrices for thick tracking on the heap and only if required.
* HUGENPART: Allow up to 65536 particles (normally 64)
* DEBUG: "Debugging output (??)"
* HDF5: Link to external HDF5 library, write "tracks2.dat" data (for postprocessing with BeamLossPattern) there instead of writing it to ASCII.
* FIO: Use FortranIO from Fortran2003 to correctly round ASCII input/output -- not actually implemented by any compilers...
* MERLINSCATTER: Interaction physics for collimation from Merlin

JamesMolson:
Thanks, updated.

SixTrack/CMakeLists.txt
```diff
@@ -510,12 +510,12 @@ get_filename_component (Fortran_COMPILER_NAME ${CMAKE_Fortran_COMPILER} NAME)
...
 if (Fortran_COMPILER_NAME MATCHES "gfortran.*")
 	# gfortran
-#	set (CMAKE_Fortran_FLAGS_RELEASE "-frecord-marker=4 -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 -O3")
+	set (CMAKE_Fortran_FLAGS_RELEASE "-frecord-marker=4 -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 -O3")
 	set (CMAKE_Fortran_FLAGS_DEBUG   "-frecord-marker=4 -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 -g")
 elseif (Fortran_COMPILER_NAME MATCHES "ifort.*")
 	# ifort (untested)
-	set (CMAKE_Fortran_FLAGS_RELEASE "-f77rtl -O3")
-	set (CMAKE_Fortran_FLAGS_DEBUG   "-f77rtl -O0 -g")
+	set (CMAKE_Fortran_FLAGS_RELEASE "-mia32 -fp-model source -fp-model strict -fp-model no-except -O3")
```
kyrsjo: Regarding -m32, wasn't m32/m64 supposed to be a separate set of flags?

JamesMolson:
From our discussion:

What is CPSS?
Check some valid option combinations.
Change default options.
Check compiler build flags.
Check FindHDF5.cmake - is it sufficient?
Look into testing.
Add wrapper script?
Add individual file astuce targets.
Add native build option.

kyrsjo:
Reply from Eric:

Hi Kyrre, CPSS was the precursor of BOINC at CERN. I am keeping the
code in SixDesk as we shall probably use it for HTCondor.
See paper by Wagner/McIntosh 
https://indico.cern.ch/event/0/contributions/1294557/attachments/928/1559/CHEP04-488-mcintosh-wagner.pdf

Eothred:
Hi,

Probabaly too late now, but did you take note of the branch I had created back in the day called [precompiler_no_astuce](http://svnweb.cern.ch/world/wsvn/SixTrack/branches/precompiler_no_astuce/SixTrack_noastuce/)? It already had basic CMakeLists functionality plus scripts that got rid of the astuce stuff. I assume the latter has already been solved by you in another fashion since I left.

Second, you may want to have a look at how I structured the cmake scripts for Mad-X for inspiration. The CTest setup there allows Laurent to structure the tests for his setup and cmake/ctest automatically picks up the same tests. This would allow people that are used to build tests for Mad-X to do similar stuff for SixTrack which is perhaps a benefit.

kyrsjo:
Hi Yngve, thanks for looking! No, indeed we did not look at this branch... It lives in this repository as well: https://github.com/SixTrack/SixTrack/tree/precompiler_no_astuce

There seems however to be some pieces missing - for example in [track.F](http://svnweb.cern.ch/world/wsvn/SixTrack/branches/precompiler_no_astuce/SixTrack_noastuce/sixtrack/track.F) there is a reference to sixtrack.h and folder `decks`, none of which I can actually find... Maybe some parts were not commited?

Also, I vaguely remember @rdemaria said that there were some problems with assembling the files this way, that it did not completely reproduce the results from astuce. Finally, breaking up `sixtrack.s` would mean that we fork the project. If we are really 100% sure that this is what we want, then fine, but we should take this into consideration...

Eothred:
It was most certainly a work in progress, so it would be more of a starting point or for inspirational purposes. I managed to build the default setup but did not really explore it further since it seemed the community was not interested at the time.

I don't think sixtrack.h or the decks folder existed back in that version.

kyrsjo:
Thanks!

One more reason why I think we should not split `sixtrack.s` at this stage: There are multiple branches of SixTrack around, and splitting it will make (semi-)automatic merging impossible.

kyrsjo:
Could we make a list of what is missing before the build part of this works 100% ?

I was trying to build HDF5 with it yesterday, and had some problems...

Regarding the defaults, do we change them now, or do we keep the current defaults and then change later after discussion in the comunity? I propose to keep the make_six defaults (for now), maybe with the exception of ifort -> gfortran.

I can also make the "make_six"-like wrapper - "cmake_six"?

JamesMolson:
Things that won't work currently:
Any compiler that isn't gfortran or ifort.
Anything boinc related (I've not looked into it).
Same with naglib.
It doesn't run any of the awk test scripts either.

For the HDF5 build, it works on my local machine, but not on lxplus. The default HDF5 install it finds seems to be a little old.

Perhaps a "CERN" flag could be useful in general for selecting all the appropriate work arounds for running on lxplus/lxbatch?

SixTrack/CMakeLists.txt
```diff
@@ -0,0 +1,1079 @@
...
+		message(STATUS "Found CUDA version ${CUDA_VERSION_STRING} at ${CUDA_TOOLKIT_ROOT_DIR}")
+		message(STATUS "Includes at ${CUDA_INCLUDE_DIRS}")
+		ADD_DEFINITIONS("-I${CUDA_INCLUDE_DIRS}")
+	endif()
+endif(CUDA)
+
+#Possible check for Root
+if(ROOT)
+	#/usr/share/root/cmake/FindROOT.cmake
+	SET(CMAKE_MODULE_PATH /usr/share/root/cmake/)
```
Eothred: This does not look very flexible to hard-code specific paths? If so, shouldn't it rather be at least
`set(CMAKE_MODULE_PATH /usr/share/root/cmake/ ${CMAKE_MODULE_PATH})`?

Consider rather adding a folder with the modules that needs specific setup for your case, and modify CMAKE_MODULE_PATH once at the top of file. As an example, you may look [here](https://bitbucket.org/jairhul/bdsim/src/3e88830019953f558c21e86daa04a4f8130aa39d/CMakeLists.txt?at=master&fileviewer=file-view-default), where they have made a folder cmake/Modules containing e.g. FindMadX.cmake and several others.

SixTrack/CMakeLists.txt
```diff
@@ -0,0 +1,1079 @@
...
+		set_target_properties(xcb PROPERTIES IMPORTED_LOCATION ${CMAKE_SOURCE_DIR}/lib32/libxcb.a)
+		set_target_properties(Xdmcp PROPERTIES IMPORTED_LOCATION ${CMAKE_SOURCE_DIR}/lib32/libXdmcp.a)
+
+		find_library(pthread STATIC pthread)
+		find_library(dl STATIC dl)
+		find_library(gcc_eh STATIC gcc_eh)
+
+		target_link_libraries(SixTrack graflib grafX11 packlib_noshift kernlib_noshift g2c X11 xcb Xau Xdmcp dl pthread gcc_eh)
+#	endif()
+endif(CERNLIB)
```
Eothred: Similar as my previous comment, consider rather making a FindCERNLIB.cmake which contains the necessary searches for the libraries needed, and which should set up variables such as CERNLIB_LIBRARIES etc

kyrsjo: Exactly for CERNlib, and NAGlib, I think this is (and still should!) be using pre-build binaries from who-knows-where. Personally, I consider CERNlib obsolete...

Because of this, CERNlib should probably also conflict with 64-bit.

SixTrack/CMakeLists.txt
kyrsjo: Personally I would just remove root...

kyrsjo:
Regarding the dafor mess, if you could review and merge #125 , then merge it back to your branch...

Eothred:
"I am keeping the code in SixDesk as we shall probably use it for HTCondor."

Wait, you are planning to use HTCondor as well? It's like you are planning to use all the tools I've come across in the last years! What's next, a python module?? 😄

Eothred:
> There seems however to be some pieces missing - for example in track.F there is a reference to sixtrack.h and folder decks, none of which I can actually find... Maybe some parts were not commited?"

@kyrsjo To answer your previos questions, I almost thought I had forgotten to commit a few things. Then I just realised I had a somewhat non-standard setup in order to build this. Instead of adding all the actual source I added just the python script (sixconv.py) which creates the files for sixtrack.s, so that one can do this with future versions of sixtrack.s as well (hopefully).

The sixtrack.h is created from sixtrack.h.in by cmake, depending on your options.

I made a pull request #127 which adds instructions on how to try out the proposal. However as you are planning on **not** splitting the sixtrack.s anytime soon, I would think there is not all that much interesting going on in this proposal for your case.

rdemaria:
Hello Yngve, we are interested in python tools too, see for instance what is in pytools and in the SixTrack/*.py. If you have ideas or code just let us know...

kyrsjo:
Regarding the BOINC libraries:
https://twiki.cern.ch/twiki/bin/view/LHCAtHome/SixTrackBuildBoinc

--- Riccardo

kyrsjo:
More BOINC, for Linux:
https://twiki.cern.ch/twiki/bin/view/LHCAtHome/SixTrackBuildSLC5

JamesMolson:
I now have the testing (mostly) working, which I still need to push.
This involves a tool that acts as a wrapped and calls fork() and exec() on all the required tools.

So far I check fort.10, fort.90 and everything in extra_checks.txt
I still need to do STF checks.

Some tests have issues (I've not checked what all these do, so just take them as general observations):

the output of elensidealthck6d does not match the HEL_DUMP_ canonicals - there are a few bits different. All other tests are okay.

elensidealthin6d_DYNK is missing the files in extra_checks.txt and the fort.10, fort.90 etc.

Some fort.10 canonical files are empty (thick6dblocks elensidealthck6d dynk_globalvars thick6ddynk).

There exists a lostnumxvin folder but not a lostnumxv (and it has no fort.90).

thick6dsingles doesn't seem to have much to check.

The fort.10.canonical in elensidealthin6d and elensidealthin6din are different. Which should be used?

Testing with CR enabled and killing the run after a number of seconds also seems to work, though I did get a failure on dynk_globalvars depending on the kill time.

I made some tests to see if the checkfxxx tools will run, but these seem to not initialise any memory so if they are called without any input files then you get what happens to be in the memory at the time (see prob(52) in checkf10.f for example), which makes my check for the default output fail sometimes.


Other notes: running with 64 bit enabled on everything makes no differences to the tests.

Running with -march=native which on my machine will call in avx/avx2/fma will cause some bit differences and cause all the tests to fail.

I'm going to try and have this finished by the meeting next week.

rdemaria:
'fma' is known to produce different results, we need numerical
reproducibility do identify hardware errors in the framework of large
dynamic aperture simulation campaign.

JamesMolson:
yes, using -mno-fma etc with -march=native on gcc lets all the tests pass, so we could probably combine the two safely, then avx2 and friends can still be pulled in safely.

JamesMolson:
In addition, from the ifort manual:

"However, if you specify -fp-model strict (Linux* and OS X*), but do not explicitly specify -fma, the default is -no-fma."

We do specify -fp-model strict, but I'll add in -no-fma for ifort too just in case something odd will happen on windows.

JamesMolson:
So with the read90 patch in #131 the test suite should also run with the help of a wrapper program.

It will check fort.10, fort.90, or STF instead if that is enabled.
It will also do a comparison on the first files listed in extra_checks.txt

Of course all the tests can now be ran in parallel and submitted to the dashboard.

Some of the tests may need fixing a little.

JamesMolson:
Also if we really want lf95 to work we will need to do something like this https://cmake.org/pipermail/cmake/2010-December/041293.html

Eothred:
@JamesMolson I started that work but never finished it. My commits can still be found in the lahey branch on https://gitlab.com/Eothred/cmake/commits/lahey

kyrsjo:
cmake_six: https://github.com/JamesMolson/SixTrack/pull/2

JamesMolson:
okay, so with the FMA fix almost all tests now pass. I need to fix some ifort flags I broke, and also set some linker flags correctly. There are a few things I have not looked at yet, like the DA version, naglib, and boinc, but for most people it should be good to use soon, so we could finally merge it all into the master branch. Soon.

kyrsjo:
Awesome! Then we're getting really close. Before merging to master, it would be good to have DA compiling (this is a different executable, which requires naglib iirc). We might want to write some documentation also - maybe a something like "Doc/buildAndTest/buildAndTest.tex"? Some way dumping the switches would be good too; or we may just document the most common ones in the LaTeX and direct people to run ccmake to find the rest...

Once it's merged to master and DA is working, we can remove the old build system, and start fixing libraries. The first would be to fix crlibm/liberic, the second would be to fix BOINC...

kyrsjo:
Regarding Lahey, there is a newer version on AFS in `/afs/cern.ch/sw/fortran/lahey/lf6481/`, however there is some license problem. I've contacted IT about it...

JamesMolson:
@kyrsjo could we have a "cmake_six clean" also?

kyrsjo:
Excellent idea! Will do it after lunch.On Dec 13, 2016 12:13, James Molson <notifications@github.com> wrote:@kyrsjo could we have a "cmake_six clean" also?

JamesMolson:
For that lf95 I get:

ERROR --  License error: 6.

From the readme:

  ERROR -- License Error : 6     The MAC address encrypted in lflic.dat does not
                                 match any of the network card addresses

pretty pointless to try and support it if it will only run on one machine.

kyrsjo:
Whops. I got the same error, but I did not see that. Why did they put it on AFS then?!?

According to ` swdkeyreq.ini`, the license belongs to a certain Harry Renshall... I'll try to add both you and Harry to the CERN helpdesk request.

kyrsjo:
Hmm. I just tried to build with option DATAMODS. That failed miserably...

```
[ 86%] Built target generate_input
[ 91%] Generating datamods.f
[ 91%] Generating beamgas.f90
Scanning dependencies of target SixTrack
[ 97%] Building Fortran object CMakeFiles/SixTrack.dir/beamgas.f90.o
[ 97%] Building Fortran object CMakeFiles/SixTrack.dir/datamods.f.o
[100%] Linking Fortran executable SixTrack_4543_datamods_crlibm_fast_tilt_cmake_gfortran_static_64
/usr/lib/gcc/x86_64-redhat-linux/5.3.1/../../../../lib64/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
CMakeFiles/SixTrack.dir/datamods.f.o: In function `__bigmats_MOD_allocate_thickarrays':
datamods.f:(.text+0xe71): undefined reference to `prror_'
collect2: error: ld returned 1 exit status
CMakeFiles/SixTrack.dir/build.make:129: recipe for target 'SixTrack_4543_datamods_crlibm_fast_tilt_cmake_gfortran_static_64' failed
make[2]: *** [SixTrack_4543_datamods_crlibm_fast_tilt_cmake_gfortran_static_64] Error 1
CMakeFiles/Makefile2:71: recipe for target 'CMakeFiles/SixTrack.dir/all' failed
make[1]: *** [CMakeFiles/SixTrack.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

Done! If everything went well, you'll find the executable in SixTrack_cmakesix_DATAMODS_defaultcompiler
```

Compiled using `cmake_six DATAMODS`

JamesMolson:
1/36 Test #12: crabamp ..........................   Passed    0.25 sec
 2/36 Test  #1: SelfTest_verify10 ................   Passed    0.47 sec
 3/36 Test  #2: SelfTest_checkf10 ................   Passed    0.46 sec
 4/36 Test  #3: SelfTest_compf10 .................   Passed    0.56 sec
 5/36 Test  #5: SelfTest_checkf1014 ..............   Passed    0.56 sec
 6/36 Test  #4: SelfTest_checkf110 ...............   Passed    0.66 sec
 7/36 Test  #6: SelfTest_read90 ..................   Passed    0.66 sec
 8/36 Test #31: thick6dblocks ....................   Passed    2.36 sec
 9/36 Test #36: thick6ddynk ......................   Passed    2.81 sec
10/36 Test #34: elensidealthck6d .................***Failed    3.50 sec
11/36 Test #19: exact ............................   Passed    6.76 sec
12/36 Test #22: javier ...........................   Passed    6.96 sec
13/36 Test #16: elensidealthin4d .................   Passed    8.93 sec
14/36 Test #33: elensidealthck4d .................   Passed    9.90 sec
15/36 Test #25: lostnotilt .......................   Passed   14.23 sec
16/36 Test #11: bb_ntwin1 ........................   Passed   15.28 sec
17/36 Test #10: bb ...............................   Passed   15.41 sec
18/36 Test #26: notilt ...........................   Passed   18.27 sec
19/36 Test #24: lost .............................   Passed   20.01 sec
20/36 Test #29: s316 .............................   Passed   25.31 sec
21/36 Test #15: dynk_globalvars ..................   Passed   28.89 sec
22/36 Test #13: dipedge ..........................   Passed   47.01 sec
23/36 Test #30: thick4 ...........................   Passed   73.43 sec
24/36 Test  #9: bbe571ib0 ........................   Passed   86.74 sec
25/36 Test #17: elensidealthin6d .................   Passed  119.84 sec
26/36 Test #32: tilt .............................   Passed  183.44 sec
27/36 Test #23: lostevery ........................   Passed  353.28 sec
28/36 Test #35: fma ..............................   Passed  394.78 sec
29/36 Test #21: frs ..............................   Passed  515.55 sec
30/36 Test #14: distance .........................   Passed  630.07 sec
31/36 Test #18: eric .............................   Passed  638.22 sec
32/36 Test  #7: bbe51 ............................   Passed  969.59 sec
33/36 Test #20: frs60 ............................   Passed  973.33 sec
34/36 Test  #8: bbe52 ............................   Passed  977.90 sec


All good so far.

JamesMolson:
Okay, this is now in a reasonable state. Please test it with various situations and try and break things. If it works this should be ok to merge now, as long as you don't use DA or BOINC or NAGLIB.

Also please try to run the test suite on interesting machines.

kyrsjo:
Wow, this is a lot of good work! I'll test more and maybe comment on a few things in the files.

SixTest/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+endif()
+
+if(NOT COLLIMAT)
+	SET(SIXTRACK_TESTS bbe51 bbe52 bbe571ib0 bb bb_ntwin1 crabamp dipedge distance dynk_globalvars elensidealthin4d elensidealthin6d eric exact frs60 frs javier lostevery lost lostnumxv lostnotilt notilt prob1 prob3 s316 thick4 thick6dblocks tilt elensidealthck4d elensidealthck6d fma thick6ddynk ${SIXTRACK_TESTS})
+endif(NOT COLLIMAT)
+
+#For CR builds, test the CR functions, so kill/resume the test
+if(CR)
+	SET(CR_ENABLE 1)
+	SET(CR_KILL_TIME 10)
```
kyrsjo: Could we do multiple kills, like before? Say at 2,4,8,16,32,64,... seconds?

SixTest/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+	#if STF, we need to use the STF flag on read90
+	if(STF)
+		LIST(FIND STF_TESTS ${TST} IS_STF)
+		if(NOT ${IS_STF} MATCHES -1)
+			SET(STF_ENABLE 1)
+		else()
+			SET(STF_ENABLE 0)
+		endif()
+		##Disable checking fort.90
+		SET(FORT90 0)
```
JamesMolson: This is slightly harder to do so I'll have to have a think.

kyrsjo: Actually, in principle we could cross check not only fort.90 but 89, 88 etc. as well. But then we would need more canonicals...

For this, maybe there should be not only STF_ENABLE (which triggers CheckSTF in the `SixTestWrapper.cpp`) but STF_90_crosscheck, which should be run if (a) we're running the STF version and (b) there is a fort.90 present.

SixTrack/cmake_six
```diff
@@ -0,0 +1,255 @@
...
+    rm -rf $NAME
+fi
+mkdir $NAME
+cd $NAME
+
+set +e #From here on we catch errors manually...
+
+#Use EVAL to split the options from $CMAKE_OPTIONS_LIST
+# http://stackoverflow.com/questions/6087494/bash-inserting-quotes-into-string-before-execution
+eval cmake  $CMAKE_OPTIONS_LIST ../../.
```
kyrsjo: It would be great of cmake would fail (stop and return non-zero status) in case of unknown options. I've tried to make this shell script as lightweight as possible, so any real option checking is done at the cmake level.

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+	message(STATUS "Please remove these files from ${CMAKE_BINARY_DIR} first:")
+	message(STATUS "CMakeCache.txt")
+	message(STATUS "CMakeFiles")
+	message(STATUS "Once these files are removed, create a separate directory")
+	message(STATUS "and run CMake from there")
+	message(FATAL_ERROR "in-source build detected")
+endif()
+
+# Update the documentation string of CMAKE_BUILD_TYPE for GUIs to include the MPI options
+SET(CMAKE_BUILD_TYPE Release CACHE STRING
```
kyrsjo: Hmm. This should probably be supported through cmake_six.

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+OPTION(BIGNPART "Allow tracking up to 2048 particles (normally 64)." OFF)
+OPTION(HUGENPART "Allow tracking up to 65536 particles (normally 64)" OFF)
+OPTION(DATAMODS "Allocate big matrices for thick tracking on the heap and only if required" OFF)
+OPTION(DEBUG "Debugging output" OFF)
+OPTION(HDF5 "Link to external HDF5 library, write tracks2.dat data (for postprocessing with BeamLossPattern) there instead of writing it to ASCII." OFF)
+OPTION(FIO "Use FortranIO from Fortran2003 to correctly round ASCII input/output" OFF)
+OPTION(MERLINSCATTER "Interaction physics for collimation from Merlin" OFF)
+
+#Compiler stuff
+OPTION(IFORT "Use the intel fortran compiler" OFF)
```
kyrsjo: Do we still need these?

CMakeLists.txt
JamesMolson: Yes, this is how you turn on debug flags and so on.

CMakeLists.txt
JamesMolson: No, the only one that changes the code is the lf95 astuce flag.

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+OPTION(LF95 "Use the Lahey-Fujitsu lf95 fortran compiler" OFF)
+OPTION(PGF90 "Use the pgf90 fortran compiler" OFF)
+OPTION(F95 "Use the g95 fortran compiler" OFF)
+OPTION(GFORTRAN "Use the gfortran fortran compiler" OFF)
+OPTION(NAGFOR "Use NAGFOR compiler" OFF)
+
+#Build options
```
kyrsjo: Maybe comment out the irrelevant build options like OPENMP, MPI, CUDA, ROOT until it becomes relevant?

CMakeLists.txt
kyrsjo: OK, let's see if we really need this then. Could it anyway be triggered by choosing CMAKE_Fortran_COMPILER=lf95 ? The others can go I guess...

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+if(32BIT)
+	SET(SIXTRACK_FEATURES "${SIXTRACK_FEATURES}_32")
+endif()
+if(64BIT)
+	SET(SIXTRACK_FEATURES "${SIXTRACK_FEATURES}_64")
+endif()
+
+#Ugly, but does the job...
+SET(SIXTRACK_BINARY_NAME "${PROJECT_NAME}_${${PROJECT_NAME}_VERSION_MAJOR}${${PROJECT_NAME}_VERSION_MINOR}${${PROJECT_NAME}_VERSION_PATCH}_${SIXTRACK_FEATURES}" )
+
```
kyrsjo: Another comment here, indicating that you're done with making the binary name?

CMakeLists.txt
kyrsjo: (including the code below to actually setup these packages)

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+	# nagfor
+	#Here I am going to assume that since we are using nagfor we are on lxplus for now
+	set (CMAKE_Fortran_FLAGS_RELEASE "-O4 -dusty -ieee=full -dcfuns")
+	set (CMAKE_Fortran_FLAGS_DEBUG   "-O0 -g -dusty -ieee=full -dcfuns")
+
+	if(32BIT)
+		set(CMAKE_Fortran_FLAGS_RELEASE "-abi=32 ${CMAKE_Fortran_FLAGS_RELEASE}")
+		set(CMAKE_Fortran_FLAGS_DEBUG "-abi=32 ${CMAKE_Fortran_FLAGS_DEBUG}")
+		if(STATIC)
+			find_library(g2cLink NAMES libg2c.a PATHS /usr/lib /usr/lib/gcc/x86_64-redhat-linux/3.4.6/32 )
```
kyrsjo: Ooh, fun. I guess this will break spectacularly if the last folder isn't there? Is it really necessary to hardcode a path like this?

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+else (Fortran_COMPILER_NAME MATCHES "gfortran.*")
+	message ("CMAKE_Fortran_COMPILER full path: " ${CMAKE_Fortran_COMPILER})
+	message ("Fortran compiler: " ${Fortran_COMPILER_NAME})
+	message ("No optimized Fortran compiler flags are known, we just try -O2...")
+	set (CMAKE_Fortran_FLAGS_RELEASE "-O2 ${CMAKE_Fortran_FLAGS_RELEASE}")
+	set (CMAKE_Fortran_FLAGS_DEBUG   "-O0 -g ${CMAKE_Fortran_FLAGS_DEBUG}")
+endif (Fortran_COMPILER_NAME MATCHES "gfortran.*")
+
+#Now for C flags (crlibm for now)
+#${CMAKE_C_COMPILER_ID} : From the docs: "one of "Clang", "GNU", "Intel", or "MSVC". This works even if a compiler wrapper like ccache is used."
```
kyrsjo: How is this variable actually set? Automagically when we set the C compiler?

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+	if(${CMAKE_C_COMPILER_ID} MATCHES "GNU")
+		target_compile_options(crlibm PRIVATE -fPIC -std=c99 -Wall -Wshadow -Wpointer-arith -Wcast-align -Wconversion -Waggregate-return -Wstrict-prototypes -Wnested-externs -Wlong-long -Winline -pedantic -fno-strict-aliasing)
+	elseif(${CMAKE_C_COMPILER_ID} MATCHES "Intel")
+		target_compile_options(crlibm PRIVATE -fPIC -std=c99 -Wall -Wshadow -Wpointer-arith -Wconversion -Wstrict-prototypes -Winline -pedantic -fno-strict-aliasing)
+	else()
+		target_compile_options(crlibm PRIVATE -fPIC -std=c99 -Wall -Wshadow -Wpointer-arith -Wcast-align -Wconversion -Waggregate-return -Wstrict-prototypes -Wnested-externs -Wlong-long -Winline -pedantic -fno-strict-aliasing)
+	endif()
+endif()
+
+if(HDF5_FOUND)
```
kyrsjo: If we're building with HDF5 and HDF5 is not found, shouldn't we throw our arms in the air and go home already then (when looking for HDF5)?

CMakeLists.txt
JamesMolson: Yes, cmake will set it. And it needs to know about the C compiler for crlibm.

CMakeLists.txt
JamesMolson: Yes, nagfor seems to need to link to it. If you can find a better way please say. I don't think it is that bad because the only place I can see anybody running nagfor is on lxplus.

CMakeLists.txt
JamesMolson: Sure.

SixTest/CMakeLists.txt
JamesMolson: What happens when you try and run a CR binary on a run that has already finished? Will it just do nothing or re-run everything from the start?

SixTest/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+##
+
+SET(STF_TESTS bbe51 bbe52 bbe571ib0 bb_ntwin1 bb crabamp dipedge distance elensidealthck4d elensidealthin4d elensidealthin6d eric exact fma frs60 frs javier_bignpart javier lostevery lostnotilt lost notilt prob1 prob3 s316 thick4 tilt)
+
+#only turn on the bnl test if we have BNLELENS enabled due to slight differences in the test
+if(BNLELENS)
+	SET(SIXTRACK_TESTS bnl ${SIXTRACK_TESTS})
+endif(BNLELENS)
+
+if(STF AND (BIGNPART OR HUGENPART))
```
JamesMolson: > Do we need to check for STF here?

Yes
> Should not the "if STF" bit be taken care of by the fact that javier_bignpart is marked as STF compatible?

No, note the "AND"

SixTrack/cmake_six
JamesMolson: Currently it prints a warning and just ignores the variable.

CMakeLists.txt
JamesMolson: Sure.

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+	endif(NATIVE)
+
+	if(STATIC)
+		SET(CMAKE_Fortran_LINKER_FLAGS "-static ${CMAKE_Fortran_LINKER_FLAGS}")
+	endif(STATIC)
+
+elseif (Fortran_COMPILER_NAME MATCHES "nagfor.*")
+	# nagfor
+	#Here I am going to assume that since we are using nagfor we are on lxplus for now
+	set (CMAKE_Fortran_FLAGS_RELEASE "-O4 -dusty -ieee=full -dcfuns")
```
JamesMolson: Fixed.

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+
+	if(STATIC)
+		SET(CMAKE_Fortran_LINKER_FLAGS "-Bstatic ${CMAKE_Fortran_LINKER_FLAGS}")
+	endif(STATIC)
+
+	if(NOT STATIC)
+		SET(CMAKE_Fortran_LINKER_FLAGS "-Bdynamic ${CMAKE_Fortran_LINKER_FLAGS}")
+	endif(NOT STATIC)
+
+elseif (Fortran_COMPILER_NAME MATCHES "g77")
```
JamesMolson: Fixed. I deleted it.

CMakeLists.txt
JamesMolson: Fixed.

JamesMolson: Anyway, the "REQUIRED" here:
```
find_package(HDF5 COMPONENTS Fortran REQUIRED)
```

Should make it error out if it can't find the package.

CMakeLists.txt
JamesMolson: I've commented out the option selection, not the other stuff below for now.

SixTest/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+#SET(SIXTRACK_TESTS bbe51 bbe52 bbe571ib0 bb bb_ntwin1 crabamp dipedge distance dynk_globalvars elensidealthck4d elensidealthck6d elensidealthin4d elensidealthin6d_DYNK elensidealthin6d eric exact fma frs60 frs javier lostevery lost lostnotilt notilt prob1 prob3 s316 thick4 thick6dblocks thick6ddynk tilt ${SIXTRACK_TESTS})
+
+##
+## Needs checking: thick6dsingles (missing files)
+## thick6dblocks elensidealthck6d dynk_globalvars thick6ddynk - empty fort.10
+## Disabled: elensidealthin6d_DYNK
+## "Totally untested": bnl
+##
+
+SET(STF_TESTS bbe51 bbe52 bbe571ib0 bb_ntwin1 bb crabamp dipedge distance elensidealthck4d elensidealthin4d elensidealthin6d eric exact fma frs60 frs javier_bignpart javier lostevery lostnotilt lost notilt prob1 prob3 s316 thick4 tilt)
```
JamesMolson: Fixed.

CMakeLists.txt
JamesMolson: Fixed.

kyrsjo:
Thanks!

CMakeLists.txt
```diff
@@ -76,17 +76,17 @@ OPTION(HDF5 "Link to external HDF5 library, write tracks2.dat data (for postproc
...
+#OPTION(LF95 "Use the Lahey-Fujitsu lf95 fortran compiler" OFF)
+#OPTION(PGF90 "Use the pgf90 fortran compiler" OFF)
+#OPTION(F95 "Use the g95 fortran compiler" OFF)
+#OPTION(GFORTRAN "Use the gfortran fortran compiler" OFF)
+#OPTION(NAGFOR "Use NAGFOR compiler" OFF)
+
+#Build options - some are commented out for future support
+#OPTION(OPENMP "Enable OpenMP support" OFF)
+#OPTION(MPI "Enable MPI support" OFF)
 OPTION(CUDA "Enable CUDA support" OFF)
```
JamesMolson: oops

SixTest/CMakeLists.txt
kyrsjo: Hmm. I don't fully understand, but if you say so...

SixTrack/cmake_six
kyrsjo: Yes. Would it be possible to make that warning fatal?

CMakeLists.txt
```diff
@@ -0,0 +1,1042 @@
...
+	if(NOT DEBUG)
+		ast_masks(debug ${loop})
+	endif()
+	if(NOT HDF5)
+		ast_masks(hdf5 ${loop})
+	endif()
+	if(NOT FIO)
+		ast_masks(fio ${loop})
+	endif()
+	if(NOT LF95)
```
kyrsjo: How is the LF95 flag now set? It should be tied to the compiler...

See also the NAGFOR flag, does this even exist in the AST files?

JamesMolson: No, the nagfor flag was just to set the compiler. The lf95 flag set the compiler and an ast flag, but since we are now not supporting lf95....

CMakeLists.txt
```diff
@@ -0,0 +1,1042 @@
...
+	target_link_libraries(SixTrack crlibm)
+endif(CRLIBM)
+
+#if we are using merlinscatter, link to the crlibm library
+if(MERLINSCATTER)
+	target_link_libraries(SixTrack merlinscatter)
+endif(MERLINSCATTER)
+
+#if we are using root, link to all the root libraries
+if(ROOT)
```
kyrsjo: Die, root, die?

CMakeLists.txt
```diff
@@ -0,0 +1,1042 @@
...
+	set (CMAKE_CXX_FLAGS_DEBUG   "-O0 -g ${CMAKE_CXX_FLAGS}")
+
+	if(STATIC)
+		SET(CMAKE_CXX_LINKER_FLAGS "-static ${CMAKE_CXX_LINKER_FLAGS}")
+	endif(STATIC)
+
+endif()
+
+#MPI default build options
+SET( CMAKE_CXX_FLAGS_MPI "-DNDEBUG -Wall -O3 -pedantic -DENABLE_MPI -march=native" CACHE STRING
```
kyrsjo: Comment out MPI etc. for now?

CMakeLists.txt
```diff
@@ -0,0 +1,1042 @@
...
+	if(STATIC)
+		SET(CMAKE_Fortran_LINKER_FLAGS "-Bstatic ${CMAKE_Fortran_LINKER_FLAGS}")
+	endif(STATIC)
+
+	if(NOT STATIC)
+		SET(CMAKE_Fortran_LINKER_FLAGS "-Bdynamic ${CMAKE_Fortran_LINKER_FLAGS}")
+	endif(NOT STATIC)
+
+#generic catch option
+else (Fortran_COMPILER_NAME MATCHES "gfortran.*")
```
kyrsjo: What's the difference between this and `if (Fortran_COMPILER_NAME MATCHES "gfortran.*")` above?

CMakeLists.txt
JamesMolson: I thought we discussed to kill the CERNLIB (PAW?) stuff and replace it with root bits?

CMakeLists.txt
JamesMolson:
```
if()

else()
```
Its to match nested if/else statements. The current versions of cmake don't really need this so it could just be else() with nothing in the brackets.

SixTest/CMakeLists.txt
kyrsjo: It does need bignpart:
```
TRACKING PARAMETERS-----------------------------------------------------
10000 0 64 0.84498 0.42249 0 1
1 1 2 1 2
0 0 1 1 1 50000 2
NEXT
```
64 pairs -> 128 particles.

However in principle it doesn't really need STF (say, if we in the future have an HDF5-based STF which is a different flag and a different canonical), however big/huge-npart can't be compiled without STF, and the test is STF compatible.

CMakeLists.txt
kyrsjo: Ah, I tought of it as an "else if", not an "else" catch-all. This quirk of CMake should at the very least be pointed out in a comment.

CMakeLists.txt
kyrsjo: There is a nagfor flag in this gigantic IF, a few lines above.

Leave the LF95 flag for now, it may come back (you've seen the emails) - apparently there is a newer version. The pre-2003 version is dead.

CMakeLists.txt
kyrsjo: It's just some plotting in the post processing, probably only used by Frank. And I REALLY don't want to make a ROOT<->FORTRAN interface, that does not sound like fun. I would just switch it off by default, but leave it there for him...

kyrsjo: But anyway, let's remove (or comment out, if it may be used in the future) anything we're not using.

CMakeLists.txt
```diff
@@ -0,0 +1,1045 @@
...
+
+	if(32BIT)
+		target_compile_options(crlibm PRIVATE -m32)
+	endif(32BIT)
+	if(64BIT)
+		target_compile_options(crlibm PRIVATE -m64)
+	endif(64BIT)
+
+	if(${CMAKE_C_COMPILER_ID} MATCHES "GNU")
+		target_compile_options(crlibm PRIVATE -fPIC -std=c99 -Wall -Wshadow -Wpointer-arith -Wcast-align -Wconversion -Waggregate-return -Wstrict-prototypes -Wnested-externs -Wlong-long -Winline -pedantic -fno-strict-aliasing)
```
kyrsjo: Great! Can't see that splitup yet tough, did you push it?

> Do we ever still want to deal with ARM based computers?

I know Eric is keen to empty my phone battery. But let's get this thing working on x86 first...

kyrsjo:
Getting really close now. One thing that still needs to be done is to write a short description of this in `Doc/build/build.tex`... Hopefully we can merge tomorrow!

CMakeLists.txt
JamesMolson: Oh, looking now there is a nagfor ast flag.

And it gets used a few times in sixtrack.s
Let me turn it on if we are using nagfor. Seems to be mostly wrapped in boinc stuff though.

JamesMolson: Seems to change the arguments to open() calls.

JamesMolson: Anyway, fixed this now. The nagfor gets set if we are using nagfor.

CMakeLists.txt
JamesMolson: Not yet, it is the last thing to do.

JamesMolson:
So, one of the remaining options left to fix is NAGLIB/DA
Currently in make_six we link to:

export NAGLIB="/afs/cern.ch/sw/nag/mark18/lnx/fllux18dg/libnag.a

looking in  /afs/cern.ch/sw/nag/ we have a more up to date version, mark24.

The fllux folders seem to contain 32bit libs and the fll6i folders 64 bit versions.

These all seem to have been built with ifort and are linked to a number of intel libs, which seem to exist in the rtl folder if you have a look around.

So far I'm linking to libnag_nag, and in rtl/ libsvml and libirng.

Building dynamic linking, it is good, and I can run the DA version of sixtrack
For static linking, I'm getting missing symbols and I can't seem to find any archives that contain what is missing.

May need to restrict it to building with ifort, I'll have a bit more of a look around.

http://www.nag.co.uk/doc/inun/fl25/l6idcl/un.html

JamesMolson:
So, there are 2 tests in SixTest_da.
We have a fort.12 and a fort.18 to check.

For the thick4da test both match, for the frs test we seem to have the same numbers, but I get different output formatting so the test fails.

SixTrack/cmake_six
JamesMolson: Not that I have found, the only way seems to be to catch stderr and see if anything ends up there or not.

JamesMolson:
So after the above, naglib/da works with ifort at least, and the crlibm changes work with gcc and clang/LLVM as C compilers. Building crlibm with icc causes the tests to fail, and I'm not really going to try and work out why right now.

JamesMolson:
For @kyrsjo and others maybe, the relevant parts of my current lxplus ~/.bashrc

```
if [ -f /afs/cern.ch/sw/fortran/nag/usenag.bash ]; then
        source /afs/cern.ch/sw/fortran/nag/usenag.bash 6.0 > /dev/null
fi

if [ -f /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh ]; then
        source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh > /dev/null
fi

if [ -f /afs/cern.ch/sw/lcg/external/gcc/4.9/x86_64-slc6-gcc49-opt/setup.sh ]; then
        source /afs/cern.ch/sw/lcg/external/gcc/4.9/x86_64-slc6-gcc49-opt/setup.sh
fi

if [ -d /afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/ ]; then
        export PATH=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64/bin:$PATH
        export CMAKE_ROOT=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64
fi

if [ -f /afs/cern.ch/sw/lcg/external/geant4/10.3/x86_64-slc6-gcc49-opt/CMake-setup.sh ]; then
        source /afs/cern.ch/sw/lcg/external/geant4/10.3/x86_64-slc6-gcc49-opt/CMake-setup.sh
fi
```
This should get you the latest ifort, nagfor, cmake, geant4. The version of geant4 on there is linked to the libs of a specific gcc version which is why 4.9 is used. For ifort, if you want 32bit just run "17-all-setup.sh ia32". I think even the computing people say the default lxplus cmake install is pretty ancient and to update it to something newer.

kyrsjo:
Regarding DA/naglib: Great. I think now it should be up to Frank etc. to test it.
All the other stuff is really great.

How to setup compilers etc. should really go into the manual. Some of it may also be done from cmake_six? Just check if /afs/cern.ch is there, if it is -> source the relevant compiler setup.

For ctest, if you could save the SixTrack STDOUT output to "fort.6" in the same folder as the rest of the output files, that would be great...

I then need to sit down, try a bunch of flags, and compare the generated sources and flags, to check that what we are building is really equivalent. Then testing testing testing...

SixTest/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+ADD_TEST(NAME SelfTest_checkf110 COMMAND checkf110)
+set_tests_properties(SelfTest_checkf110 PROPERTIES PASS_REGULAR_EXPRESSION "Comparing VERSION    0.0000000000000000 *to *0.0000000000000000.*\n.*DIFF I\\/O error, wrong no of lines\\!\\! line no *0")
+
+ADD_TEST(NAME SelfTest_checkf1014 COMMAND checkf1014)
+set_tests_properties(SelfTest_checkf1014 PROPERTIES PASS_REGULAR_EXPRESSION "Comparing VERSION    0.0000000000000000 *to *0.0000000000000000.*\n.*checkf1014_DIFF I\\/O error, wrong no of lines\\!\\! line no *0")
+
+ADD_TEST(NAME SelfTest_read90 COMMAND read90)
+set_tests_properties(SelfTest_read90 PROPERTIES PASS_REGULAR_EXPRESSION "At line .* of file .*read90.f90 \\(unit = 190\\)\nFortran runtime error: Cannot open file 'fort.190': No such file or directory")
+
+SET(SIXTRACK_TESTS ${SIXTRACK_TESTS})
```
Eothred: Not 100% if I am right about this (haven't tested how sixtrack behaves directly), but I think this command might do "not a lot". If you want from shell/environment I think it should be

```SET(SIXTRACK_TESTS $ENV{SIXTRACK_TESTS})```

Similarly, if you want to set an environment variable you can do

```SET(ENV{MYVAR} "This is some string")```

It is simple to check with a message before and after the command that prints the variable.

**Comment by JamesMolson** [comment]
I've only run a small subselection of the tests, and there are still dashboard submission issues, but I think this is what you wanted @rdemaria @kyrsjo 

http://abp-cdash.web.cern.ch/abp-cdash/viewCoverage.php?buildid=30885

**Comment by kyrsjo** [comment]
Coverity! Cool! Very very cool.

I see for example plato_seq, which is 0% due to the fma test not being ran?

**Comment by JamesMolson** [comment]
Yes! I'll re-run it with everything except the prob tests now.

**Comment by kyrsjo** [comment]
Cool. Very cool. Is it a lot slower to run with this instrumentation than normal runs?

I'm writing some beginning of Doc/building_sixtrack now...

**Comment by kyrsjo** [comment]
Docs!
https://github.com/JamesMolson/SixTrack/compare/cmake...kyrsjo:cmake_six

Anything I'm clearly missing? Anything that isn't really necessary?

File: CMakeLists.txt (hunk 7 above)
Comment (JamesMolson) [inline CMakeLists.txt]: Yes.

File: SixTest/CMakeLists.txt (hunk 24 above)
Comment (JamesMolson) [inline SixTest/CMakeLists.txt]: Yes, I'll just remove it.

**Comment by JamesMolson** [comment]
Here it is with everything except prob1 and 3 http://abp-cdash.web.cern.ch/abp-cdash/viewCoverage.php?buildid=30886

I should run without any -O flags because that can change the code a little bit.

For the docs, that looks fine. Perhaps I should add a target that builds the latex files too?

**Comment by rdemaria** [comment]
Yes!

**Comment by kyrsjo** [comment]
If cmake shall build the latex, I have two comments:
 * We need to distribute a built version somewhere. However I agree that having this out of the repository would be very nice (it's a neverending superstupid merge conflict).
 * Let's not use date \today in sixphys.tex...

**Comment by JamesMolson** [comment]
Other issues:
see: http://abp-cdash.web.cern.ch/abp-cdash/viewErrorLog.php?buildid=30886&projectid=15&date=

```
 Error reported on 2016-12-16 18:52:35 for build #30886
(Test XML parser): Cannot insert test

Error reported on 2016-12-16 18:52:35 for build #30886
(Test::Insert): Output is bigger than max_allowed_packet

Error reported on 2016-12-16 18:52:35 for build #30886
(Cannot insert test: fma into the database): SQL error: Got a packet bigger than 'max_allowed_packet' bytes
```

And so on for other tests. Looks like some settings need to be changed on the cdash database

https://stackoverflow.com/questions/93128/mysql-error-1153-got-a-packet-bigger-than-max-allowed-packet-bytes/104176#104176


For the run time question, no it runs just as fast on my machine.

Running with valgrind takes forever though. I'm yet to turn that on..

**Comment by kyrsjo** [comment]
Fortran runtimes alone makes valgrind go a bit crazy...On Dec 16, 2016 19:12, James Molson <notifications@github.com> wrote:Other issues:
see: http://abp-cdash.web.cern.ch/abp-cdash/viewErrorLog.php?buildid=30886&projectid=15&date=
 Error reported on 2016-12-16 18:52:35 for build #30886
(Test XML parser): Cannot insert test

Error reported on 2016-12-16 18:52:35 for build #30886
(Test::Insert): Output is bigger than max_allowed_packet

Error reported on 2016-12-16 18:52:35 for build #30886
(Cannot insert test: fma into the database): SQL error: Got a packet bigger than 'max_allowed_packet' bytes

And so on for other tests. Looks like some settings need to be changed on the cdash database
https://stackoverflow.com/questions/93128/mysql-error-1153-got-a-packet-bigger-than-max-allowed-packet-bytes/104176#104176
For the run time question, no it runs just as fast on my machine.
Running with valgrind takes forever though. I'm yet to turn that on..

**Comment by kyrsjo** [comment]
Stupid question: Tabs or spaces? Emacs seems to be hell-bent on doing *both at the same time* in a consistent way (cats and dogs living together, etc....) -- se for example the big block of if's calling the linters in the top of `cmake_six`.

For the CMakelists, I changed the one in crlbm into spaces in https://github.com/SixTrack/SixTrack/pull/107/commits/438264c33c2e7ccd44f02f0e45f566c66fec3b5f - it was actually a mixture?

Anyway, strange stuff. Maybe just go with what the editor wants, and don't mix the convention in a single file (except when that appears to be the convention, as in cmake_six)?

File: CMakeLists.txt (hunk 7 above)
Comment (kyrsjo) [inline CMakeLists.txt]: Hmm. This doesn't really work, does it? Well: If I comment out these lines and don't set -DCMAKE_BUILD_TYPE on the command line, it doesn't build. However, I don't see anywhere the special flags for a debug release?

PS: Before modifying the CMakeLists.txt again, please merge from my cmake_six branch...

In make_six, I use for gfortran when debugging:
`-frecord-marker=4 -$myopt -m32 -g -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 -ffpe-trap=invalid,zero,overflow,underflow -fcheck=all -O0 -g`; normal flags are ` -frecord-marker=4 -$myopt -m32 -g -fno-second-underscore -funroll-loops -mfpmath=sse -msse2`
This is really quite efficient for finding bugs!

File: SixTest/CMakeLists.txt (hunk 14 above)
Comment (kyrsjo) [inline SixTest/CMakeLists.txt]: Ping!

File: SixTest/CMakeLists.txt (hunk 4 above)
Comment (kyrsjo) [inline SixTest/CMakeLists.txt]: Ping!

File: SixTest/CMakeLists.txt (hunk 5 above)
Comment (kyrsjo) [inline SixTest/CMakeLists.txt]: Ping?

**Comment by kyrsjo** [comment]
I'm making tests now (and found & fixed a bunch of bugs in make_six)... Just for future note: The "standard" `make_six gfortran` build can be reproduced by `./cmake_six gfortran -64BIT 32BIT CERNLIB`. Now, let's check that it (and others are identical...

**Comment by kyrsjo** [comment]
I don't see any difference in the .f files, however *something* is clearly different in the linking: The make_six binary is 7.4 MB, but the cmake binary is 5.6 MB??

**Comment by JamesMolson** [comment]
Is this on lxplus? make_six will link to a version of cernlib it can find on there I think.

Same build flags too?

**Comment by kyrsjo** [comment]
no, my local box, but it has AFS...

**Comment by JamesMolson** [comment]
In make_six

```
if test "$cernlib" != ""
then
    if [[ "$OSTYPE" == Linux ]]
    then
        if test -f /cern/pro/bin/cernlib -a -f /usr/X11R6/lib/libX11.a
        then
```

etc.

is it linking to the files it finds there?

**Comment by kyrsjo** [comment]
It doesn't take those, I don't have the  `/cern` mount, just `/afs`. But I found some differences
* Linking from make_six (running `make -f Makefile_makesix` in the build folder after running `./make_six gfortran`, changing O4 -> O3 in the makefile ):
```
gfortran -static -m32 -o SixTrack_4545_cernlib_crlibm_gfortran_O4 track.o sixve.o sixvefox.o dabnews.o lielib.o plato_seq.o datamods.o  \
 -L../lib32 -lgraflib -lgrafX11 -lpacklib_noshift -lkernlib_noshift -lg2c -lX11 -lxcb -lXau -lXdmcp -ldl -lpthread -lgcc_eh crlibm/crlibm.a  > sixtrack.map 2>&1
```
* Linking from cmake (running `make VERBOSE=1` in the build folder, created by `./cmake_six gfortran -64BIT 32BIT CERNLIB CRLIBM`):
```
/usr/bin/gfortran   -static -m32  -m32 -frecord-marker=4 -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 -O3 CMakeFiles/SixTrack.dir/track.f.o CMakeFiles/SixTrack.dir/sixve.f.o CMakeFiles/SixTrack.dir/sixvefox.f.o CMakeFiles/SixTrack.dir/dabnews.f.o CMakeFiles/SixTrack.dir/lielib.f.o CMakeFiles/SixTrack.dir/plato_seq.f.o CMakeFiles/SixTrack.dir/beamgas.f90.o  -o SixTrack_4545_cernlib_crlibm_fast_tilt_cmake_gfortran_static_32  ../lib32/libgraflib.a ../lib32/libgrafX11.a ../lib32/libpacklib_noshift.a ../lib32/libkernlib_noshift.a ../lib32/libg2c.a ../lib32/libX11.a ../lib32/libxcb.a ../lib32/libXau.a ../lib32/libXdmcp.a -ldl -lpthread -lgcc_eh SixTrack/crlibm/libcrlibm.a 
```

I both cases I get the warning:
```
../lib32/libX11.a(x11trans.o): In function `_X11TransSocketOpen':
x11trans.c:(.text+0x7b): warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
../lib32/libX11.a(x11trans.o): In function `_X11TransSocketINETConnect':
x11trans.c:(.text+0xb2b): warning: Using 'gethostbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
x11trans.c:(.text+0xb6b): warning: Using 'getservbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

Trying to reduce make the linking flags from cmake more similar to those I get from make_six makes no difference:
```
18:51:40 kyrsjo@pcbe15843 SixTrack_cmakesix_32BIT_CERNLIB_CRLIBM_NO64BIT_gfortran $ ll SixTrack_4545_cernlib_crlibm_fast_tilt_cmake_gfortran_static_32 
-rwxrwxr-x. 1 kyrsjo kyrsjo 5598804 Dec 20 18:49 SixTrack_4545_cernlib_crlibm_fast_tilt_cmake_gfortran_static_32
18:51:45 kyrsjo@pcbe15843 SixTrack_cmakesix_32BIT_CERNLIB_CRLIBM_NO64BIT_gfortran $ /usr/bin/gfortran   -static -m32 CMakeFiles/SixTrack.dir/track.f.o CMakeFiles/SixTrack.dir/sixve.f.o CMakeFiles/SixTrack.dir/sixvefox.f.o CMakeFiles/SixTrack.dir/dabnews.f.o CMakeFiles/SixTrack.dir/lielib.f.o CMakeFiles/SixTrack.dir/plato_seq.f.o CMakeFiles/SixTrack.dir/beamgas.f90.o  -o SixTrack_4545_cernlib_crlibm_fast_tilt_cmake_gfortran_static_32  -L../lib32 -lgraflib -lgrafX11 -lpacklib_noshift -lkernlib_noshift -lg2c -lX11 -lxcb -lXau -lXdmcp -ldl -lpthread -lgcc_eh SixTrack/crlibm/libcrlibm.a
../lib32/libX11.a(x11trans.o): In function `_X11TransSocketOpen':
x11trans.c:(.text+0x7b): warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
../lib32/libX11.a(x11trans.o): In function `_X11TransSocketINETConnect':
x11trans.c:(.text+0xb2b): warning: Using 'gethostbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
x11trans.c:(.text+0xb6b): warning: Using 'getservbyname' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
18:51:52 kyrsjo@pcbe15843 SixTrack_cmakesix_32BIT_CERNLIB_CRLIBM_NO64BIT_gfortran $ ll SixTrack_4545_cernlib_crlibm_fast_tilt_cmake_gfortran_static_32 -rwxrwxr-x. 1 kyrsjo kyrsjo 5598804 Dec 20 18:51 SixTrack_4545_cernlib_crlibm_fast_tilt_cmake_gfortran_static_32
```

crlibm.a is different in size:
```
SixTrack_cmakesix_32BIT_CERNLIB_CRLIBM_NO64BIT_gfortran $ ll -h SixTrack/crlibm/libcrlibm.a
-rw-rw-r--. 1 kyrsjo kyrsjo 179K Dec 20 18:49 SixTrack/crlibm/libcrlibm.a
```
vs. make_six:
```
SixTrack_4545_cernlib_crlibm_gfortran_O4 $ ll crlibm/crlibm.a  -h
-rw-rw-r--. 1 kyrsjo kyrsjo 374K Dec 20 18:10 crlibm/crlibm.a
```

Hmm...

**Comment by JamesMolson** [comment]
I just had a look, make_six adds -g into the build flags. I only add it in for debug builds. Try using -O3 and take out -g and see what happens :)

run ls -l *.o in the make_six build folder and ls -l CMakeFiles/SixTrack.dir/*.o in your cmake build folder.

**Comment by JamesMolson** [comment]
This is with crlibm and naglib and cernlib all off, 32bit:

```
$ VERBOSE=1 make -B track.f.o
make -f CMakeFiles/SixTrack.dir/build.make CMakeFiles/SixTrack.dir/track.f.o
make[1]: Entering directory '/tmp/SixTrack'
Generating track.f
/usr/bin/cmake -E copy trackn.f track.f
Building Fortran object CMakeFiles/SixTrack.dir/track.f.o
/usr/bin/gfortran    -m32 -frecord-marker=4 -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 -O3   -c /tmp/SixTrack/track.f -o CMakeFiles/SixTrack.dir/track.f.o
```
Makes:
```
$ du -b CMakeFiles/SixTrack.dir/track.f.o 
617412	CMakeFiles/SixTrack.dir/track.f.o
```

make_six gives:
```
gfortran -c -frecord-marker=4 -O3 -m32 -fno-second-underscore -funroll-loops -mfpmath=sse -msse2 track.f
```

and

```
$ du -b track.o
617396  track.o
```

a diff on each track.f shows they are both the same.

I then realised I needed to run strip on both binaries, then we have the same size.

**Comment by kyrsjo** [comment]
Yeah, I changed to `-O3` inm make_six. Another diff was that make_six uses `-O2` to build crlibm. Doing that in CMakeLists.txt as well, I got quite similar results:

cmake:
```
-rw-rw-r--. 1 kyrsjo kyrsjo 157188 Dec 21 11:27 SixTrack/crlibm/libcrlibm.a
# After strip -s:
-rw-rw-r--. 1 kyrsjo kyrsjo 121152 Dec 21 11:42 SixTrack/crlibm/libcrlibm.a
```
make_six:
```
#Crlibm:
-rw-rw-r--. 1 kyrsjo kyrsjo 159540 Dec 21 11:23 crlibm/crlibm.a
# After strip -s
-rw-rw-r--. 1 kyrsjo kyrsjo 122738 Dec 21 11:43 crlibm/crlibm.a

#SixTrac:
-rwxrwxr-x. 1 kyrsjo kyrsjo 5734252 Dec 21 11:21 SixTrack_4545_cernlib_crlibm_gfortran_O4
# After strip -s
-rwxrwxr-x. 1 kyrsjo kyrsjo 5162548 Dec 21 11:44 SixTrack_4545_cernlib_crlibm_gfortran_O4

```
Huh, strange...

From a bit of `hexdump -C` and `meld`, it seems that the difference may be how things are ordered - maybe the linking order is different and that made some tiny difference later?

Anyway; I don't think having 1:1 identical binaries is really that important. I'm mainly interested in that the sources match, and that we have some docs.

**Comment by kyrsjo** [comment]
I think I now understand part of the problem: cmake doesn't build and link datamods if it isn't used, right? make_six does build and link an empty file... cmake still copies over the datamods.ast tough -- but it does not modify it.

**Comment by kyrsjo** [comment]
OK, let's do "equivalent flags" - make_six on the left, cmake_six on the right:
- [ ] `gfortran` -- `gfortran -64BIT 32BIT CERNLIB CRLIBM`
- [ ] `gfortran collimat` -- `gfortran -64BIT 32BIT CERNLIB -CRLIBM COLLIMAT`
- [ ] `gfortran da` -- `gfortran -64BIT 32BIT CERNLIB CRLIBM DA NAGLIB`
  * explodes on linking...
- [ ] `gfortran cr` -- gfortran -64BIT 32BIT CERNLIB CRLIBM CR`

I'll edit this post as the data comes in...

**Comment by JamesMolson** [comment]
The naglib install seems to be missing some static libs that come as part of the ifort install so the linking there will fail unless you use ifort. I could have a try with other versions, but ehhh.

I think at some point intel changed their compiler flags.

I'll change the testing wrapper file so that you can pass multiple kill times soon.

**Comment by JamesMolson** [comment]
Which version of cmake are you using? -mia32 doesn't get used at all here. (CMake 3.7).


Have you tried running make_six with the boinc option? It fails totally for me, but there is some weirdness.

For example:
./make_six boinc :
```
The following selections are ON  : tilt fast api cernlib boinc nagfor fio
The following selections are OFF : crlibm naglib da collimat cpss cr pgf90 g95 gfortran bpm beamgas bnlelens bignblz debug hdf5 ifort SSE lf95 hugenblz stf bignpart datamods hugenpart

And compiler options:
FC   = 'ifort'
FCF  = ' -O -mia32 -fp-model source -fp-model strict -fp-model no-except'
FCL  = '-static -m32'  (For linking SixTrack)
FCLD = '-m32' (For linking astuce and dafor)
```

The nagfor option seems to be selected somehow yet it has FC=ifort?


For your other issue:

> With ifort17, make_six explodes on linking as it does not understand -mia32 (setup as source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh)

This gets you a 64bit ifort which can't build 32bit things, and make_six will build as 32bit.
Try running the shell script with the argument "ia32" ?

**Comment by JamesMolson** [comment]
Well that fixed the build option always being release, but seems to have broken linking to crlibm when we have debug set.

**Comment by JamesMolson** [comment]
So it won't link with CMAKE_BUILD_TYPE=Debug on my machine, but it does work on lxplus.

I run the tests and bb/bb_ntwin1 crash

```
          DA INITIALIZATION: ORDER =  1, # of VARIABLES =  6, DIMENSION =  3


         ENTERING 6-D DA CLOSED ORBIT CALCULATION


Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0xB9DDCD in _gfortran_backtrace at backtrace.c:258
#1  0xB88600 in _gfortrani_backtrace_handler at compile_options.c:129
#2  0xBD183F in raise
#3  0xB053FC in damult_
#4  0xAED095 in damul_
#5  0x99E4D3 in umlauda_
#6  0xA45C8D in clorda_
#7  0xAE3DBD in mydaini_
#8  0x71A39B in qmodda_
#9  0x8F8E2A in MAIN__ at sixve.f:0
```

Something odd is happening.

**Comment by JamesMolson** [comment]
so it builds on my computer with -O1 but but not -O0. It still crashes on bb though.
This time I get some debugging info:
```
Backtrace for this error:
#0  0x95BD4E in _gfortrani_backtrace
#1  0x95B350 in _gfortrani_backtrace_handler
#2  0x9C995F in gsignal
#3  0x909012 in damult_ at dabnews.f:1961 (discriminator 12)
#4  0x8F8232 in damul_ at dabnews.f:1910
#5  0x8101ED in umlauda_ at sixvefox.f:4711 (discriminator 8)
#6  0x88489C in clorda_ at sixvefox.f:11948
#7  0x8F1E76 in mydaini_ at sixvefox.f:12208
#8  0x640838 in qmodda_ at sixve.f:21842
#9  0x79B11C in maincr at sixve.f:8124
```

The offending line is marked by 20
```
!
!      CALL DACHK(INA,INOA,INVA, INB,INOB,INVB, INC,INOC,INVC)
!
!     CASE OF FIRST ORDER ONLY
!     ************************

      if(nomax.eq.1) then
         minv = min(inva,invb,invc)
         ccipoa = cc(ipoa)
         ccipob = cc(ipob)
         cc(ipoc) = ccipoa*ccipob
         do 20 i=1,minv
  20     cc(ipoc+i) = ccipoa*cc(ipob+i) + ccipob*cc(ipoa+i)
         do 30 i=ipoc+minv+1,ipoc+invc
  30     cc(i) = 0d0
         return
      endif
```

gdb
```
(gdb) bt
#0  0x0000000000909012 in damult (ina=5, inb=280, inc=6) at /tmp/rel/dabnews.f:1961
#1  0x00000000008f8233 in damul (ina=5, inb=280, inc=6) at /tmp/rel/dabnews.f:1910
#2  0x00000000008101ee in umlauda () at /tmp/rel/sixvefox.f:4711
#3  0x000000000088489d in clorda (nn=6, idummy=..., am=...) at /tmp/rel/sixvefox.f:11948
#4  0x00000000008f1e77 in mydaini (ncase=1, nnord=1, nnvar=6, nndim=3, nnvar2=6, nnord1=1) at /tmp/rel/sixvefox.f:12208
#5  0x0000000000640839 in qmodda (mm=3, qwc=...) at /tmp/rel/sixve.f:21842
#6  0x000000000079b11d in maincr () at /tmp/rel/sixve.f:8124
#7  0x00000000007c4594 in main (argc=1, argv=0x7fffffffd8d4) at /tmp/rel/sixve.f:9154
#8  0x00000000009bb896 in __libc_start_main ()
#9  0x0000000000400c59 in _start ()
```
and
```
(gdb) print i
$2 = 6
(gdb) print ipoc
$3 = 36
(gdb) print ipoa
$4 = 29
(gdb) print ipob
$5 = 1954
(gdb) print ccipoa
$6 = 4.5258815505520758e-304
(gdb) 
```

That last number seems suspect.

JamesMolson:
For the linking issues:
https://stackoverflow.com/questions/30516423/inline-function-at-o0-causes-link-failure-in-clang#30516490

https://stackoverflow.com/questions/6312597/is-inline-without-static-or-extern-ever-useful-in-c99

This can be fixed by disabling these functions from being inlined when we build the debug version.

kyrsjo:
Good work! And great to have the multiple-kill support for CR runs + debugging the inlining weirdness.

The failure of the "bb" test when using Debug I would take as being *uncovered* by Debug, not created. I'm not sure if it's a bad test or a bad algorithm...

Regarding https://github.com/SixTrack/SixTrack/pull/107#issuecomment-268576518 , I think this could have been a bug in make_six that I fixed with https://github.com/SixTrack/SixTrack/pull/107/commits/efdeac7365881178093a9d6e8d35ba45a405f02c ? 

So if I understand correctly, it is now essentially complete, except for the manual? I'll write some more then...

kyrsjo:
Hmm, ifort seems to give a warning:
ifort: command line remark #10148: option '-i_dynamic' not supported

Have we ctest-ed the 2017 version?

JamesMolson:
Use the latest cmake!

kyrsjo:
Hi, I sent you a pull request:
https://github.com/JamesMolson/SixTrack/pull/5

Is there anything now that *must* be done before we can merge this branch with master and start using cmake for production? The docs really could use some more work, however it's probably "good enough" for now...

JamesMolson:
BOINC support
DA tests give the correct output, but the format depends on the compiler so the tests can seem to fail.
The CR kill method is a bit hacky.


Apart from that all seems to be good, I would just merge it now and clean up those and the docs later.

JamesMolson:
Ok, from a discussion with @kyrsjo we are merging now.

Issue #108, opened 2016-10-31
Collimation emittance is read from beam-beam block


KFubuki:
The collimation block defines the opening of the collimators with respect to a certain emittance. This emittance is read from the beam-beam block, and the variables in the collimation block to specify the emittance are completely ignored. 

Since the collimator openings are usually given with respect to an emittance of 3.5 um, and the beam-beam block should be run with the correct emittance (2.5 um for HL-LHC), it is currently impossible to run beam-beam and collimation at the same time without completely changing the input of the collimation block. Furthermore, this fact is not documented, which can lead to many errors.

I suggest that we bring back to life the mute variables in the collimation block for emittances.

kyrsjo:
Hi,

Can you try the kyrsjo/collimatEmittance branch?

KFubuki:
No problem!

kyrsjo:
This is being handled in PR #109

kyrsjo:
Fixed!

PR #109, opened 2016-10-31
Collimation block emittance fix

kyrsjo:
For collimation, use the emittance from the collimation block, not the emittance from the beam block.

This should fix issue #108 

Before merging, we must have the agreement from the collimation group, and make sure everyone notices, since this changes the meaning of their input files!


kyrsjo:
TODO: Separate "distribution emittance" and "collimator emittance", and use the non-presence of these 2 new variables to detect old-format input files.

kyrsjo:
Tested & works perfectly! Modification to fort.3:
 * Change collimation emittance to `3.5 3.5` (units are um), same as was in the `BEAM` block previously.

kyrsjo:
It seems that some of the collimation emittance variables are defined in several `+cd` blocks:
In `dbmaincr`, `dbcommon`, `dbmkdist` (no COMMON block `ralph`):
```
      double precision myemitx0,myemity0,myalphax,myalphay,mybetax,     &
     &mybetay,rselect
      common /ralph/ myemitx0,myemity0,myalphax,myalphay,mybetax,       &
     &mybetay,rselect
```
These blocks are included in
 * dbmaincr:
   * program maincr (if collimat)
   * subroutine crcheck (if collimat)
   * subroutine crpoint (if collimat)
   * subroutine crstart (if collimat)
   * subroutine dumpbl1 (if collimat) -- DEAD SUBROUTINE, NEVER CALLED!
   * subroutine dumpzfz (if collimat) -- DEAD SUBROUTINE, NEVER CALLED (commented out call just before `call closeunits`)!
   * subroutine dumpxy (if collimat) -- DEAD SUBROUTINE, NEVER CALLED!
   * subroutine dumpbnl (if collimat) -- DEAD SUBROUTINE, NEVER CALLED!
   * subroutine dumpsynu (if collimat) -- DEAD SUBROUTINE, NEVER CALLED!
   * subroutine dump (if collimat) -- DEAD SUBROUTINE, NEVER CALLED!
   * subroutine dumpbin (if collimat) -- DEAD SUBROUTINE, NEVER CALLED (multiple commented-out calls, inside `+if debug` / `+ei`)!
   * subroutine dumphex (if collimat) -- DEAD SUBROUTINE, NEVER CALLED!
 * dbcommon:
   * subroutine trauthin (if collimat)
   * subroutine thin6d (if collimat)
   * subroutine dump_beam_population (if collimat)
   * subroutine dynk_apply (if collimat)
   * subroutine readcollimator
   * subroutine beamGas
 * dbmkdist:
   * subroutine makedis
   * subroutine makedis_st
   * subroutine makedis_de
   * subroutine readdis
   * subroutine makedis_radial
   * subroutine makedis_ha

I wonder how we can untangle this mess... Also, can we delete all those unused "dump*" routines? They are never actually used, but looks like they could be useful for debugging...

kyrsjo:
OK, I ended up cleaning up a lot of stuff - the `dbmaincr` block was included all over the place, but never actually used... We're then left with `dbcommon`, which defines global variables holding the various beam parameters (including the emittances) for the collimation code; and `dbmkdist` which defines (identically named) local variables in the distribution generation blocks.

I'll simply rename the `myemitx0` in `dbcommon`to two new variables `myemitx0_collgap` and `myemitx0_dist`, then go over the code to correct all occurences so that they point to the correct version. @amereghe - does this sound OK to you?

kyrsjo:
Another interesting thing: the amplitudes in "betafunctions.dat" are calculated using the emittance. I would think that using the distribution emittance makes most sense here?

A small todo:
In subroutine TRAUTHIN, there are lots of files opened. However these are not protected by any INQURE, to check for double openings with same unit number...

kyrsjo:
Something else: For pencil beam, always use "_collgap" emittance. Also, if pencil beam is active, it is probably a good idea to demand (in daten) that the beam emittance and the collgap emittance are equal...

kyrsjo:
More fun stuff:

When writing data to fort.2, this is done if (1) particle has hit a collimator, or (2) particle positition > 99 mm, or (3) particle position (never mind the angle or the orbit) in sigma > sigsecut2 or sigsecut3.

I'll use the collgaps emittance for this...

More info:
https://twiki.cern.ch/twiki/bin/viewauth/LHCAtHome/SixTrackCollimatVer

kyrsjo:
After some more thinking: Let's use the "beam" emittance for the beam distribution ONLY (except for pencil beam, for which the two emittances should always be the same), and leave everything else (analysis etc) to the "collgaps" emittance, which would normally be set to the "collimation emittance" aka. 3.5um.

kyrsjo:
TODO: Document file 7878, debugging output for pencil beam

kyrsjo:
OK DONE!

Example of new collimation block:
```
COLLIMATION
.TRUE. 
2 7000000
1 5.7 0.04 0. 0. "init_dist.txt" 1.129E-4 75.5
.TRUE.	15.0 	18.0 	999. 	20.0 	5.7 	7.7 	999. 	10. 	12. 	999. 	9. 	8.5 	 999.0
10.9	30.	10.9	15.	10.9	30.	10.9	15.	999.	999.
 0 19789.0 20150.0 1 1
 -1.3899e-6 -9.345e-5 5.05324e-3 -1.6595e-2 2.15955e-2 -9.96261e-3 1.0
 -1.3899e-6 -9.345e-5 5.05324e-3 -1.6595e-2 2.15955e-2 -9.96261e-3 1.0
2.5 2.5 3.5 3.5
.FALSE. .FALSE. 200 .TRUE. TCP.C6L7.B1 .FALSE. .TRUE. .TRUE. .TRUE. 
 0 0 0 0
 0 0 0 0 0 0 0 0 0 0 .FALSE.
.FALSE. 6.003 .0015
0 0 .FALSE. .FALSE. 
0 0.1 0.05 0.05 3
"CollDB_HL.data" 1
.TRUE. .FALSE. WAbsVertLowbcoll 101 1 1.
NEXT
```

Note: Here we place the collimators relative to a 3.5 um emittance, and generate a beam distribution relative to a 2.5 um emittance. To reproduce the old behaviour, put 3.5 um in all 4 columns.

In case something is missed, an error message is provided:
```
 ERR> EMITTANCES NOT DEFINED! CHECK COLLIMAT BLOCK!
 ERR> EXPECTED FORMAT OF LINE 9 IN COLLIMAT BLOCK:
 emitnx0_dist, emitny0_dist, emitnx0_collgap, emitny0_collgap
 ERR> ALL EMITTANCES SHOULD BE NORMALIZED.FIRST PUT EMITTANCE FOR DISTRIBTION GENERATION, THEN FOR COLLIMATOR POSITION ETC. UNITS IN [MM*MRAD].
 ERR> EXAMPLE:
 2.5 2.5 3.5 3.5

         ++++++++++++++++++++++++
         +++++ERROR DETECTED+++++
         ++++++++++++++++++++++++
         RUN TERMINATED ABNORMALLY !!!
```

This is reached during initialization, so it only takes a few seconds to get the message.

Issue #13, opened 2015-09-21
Rank mismatch complilation warnings


kyrsjo:
Problem:
gfortran gives a lot of warnings when compiling "dabnews.f" and "lielib.f". This may not be a real problem, but the excessive output may mask real problems by drowning them.

> gfortran -c -frecord-marker=4 -O4 -m32 -g -fno-second-underscore -funroll-loops dabnews.f
> dabnews.f:376.17:
> 
> ```
>   call daall(iall,1,'$$UNPACK$$',nomax,nvmax)                       
>              1
> ```
> 
> Warning: Rank mismatch in argument 'ic' at (1) (rank-1 and scalar)
> dabnews.f:383.17:
> 
> ```
>   call daall(iall,1,aa,nomax,nvmax)                                 
>              1
> ```

(etc. -- the output is here cropped)

The code for these files are found in dabnew.s and lielib.s, and are preprocesses through the astuce masks lielib.ast and dabnews.ast.

rdemaria:
Is it a new issue, or has always been there?

kyrsjo:
It has been there as long as I've worked with SixTrack.

kyrsjo:
Fixed in #118 by James Molson!

Issue #114, opened 2016-11-17
Remove "Comments as version control"


kyrsjo:
* Delete all lines with "!hrXX" - these are just new parenthesis, and the old expressions are often wrong; this should remove about 3000 unnecessary lines

More?

To be done in a separate branch...

JamesMolson:
sed -i -e '/^\!hr[0-1][0-9].*/d' sixtrack.s

kyrsjo:
Also kill `!hhr` `!hrr` lines

rdemaria:
I think we can remove them in the main branch as well, I just want to be sure that there is no other use about those lines, like automatic tools relying on that...

kyrsjo:
No tools that I know about relies on these... So unless you have any?

I'll ping Eric to check that he is also OK with it.

kyrsjo:
Eric is OK with it, as long as it comes with TWO version bumps. I guess we're due for one anyway... I'll comment more on the general strategy in #119 

> Hi Kyrre, these comments indicate everywhere Harry Renshall
> inserted parentheses for unique order of evaluation.
> They are definitely of historic interest. I guess it is OK to
> delete now as long as we keep the new version and the previous version
> with parentheses for comparison. It would certainly reduce the number
> of lines in sixtrack.s (and lielib.s etc). In other words this should be the
> ONLY change between two consecutive versions apart from the Version Number.
> Eric.
>

JamesMolson:
A lot of the collimation code is guilty of this as well. I'll have a scrub of that in the future when I have time.

kyrsjo:
Partially fixed in #117 . Keeping open until also fixed in collimation.

kyrsjo:
I think we have the go ahead from the collimation team? @rbruce1 @amereghe ?

rdemaria:
@amereghe, can you comment?

amereghe:
I would say 'go ahead'

vkbo:
Still partially an issue in collimation, including in some of the recently added code!

The `!hrXX` etc comments are kill-on-sight. I've removed a lot, but we may just do a global replace of them.

vkbo:
All `!hrXX` type comments removed in PR #944

JamesMolson:
done I guess

vkbo:
Yeah, think I got the rest when I went through the collimation module.

PR #117, opened 2016-11-18
Code comments are not version control.

JamesMolson:
See issue #114


kyrsjo:
Testing (compiling them) and reviewing (reading through the diff) the changes. Unless someone is strongly opposed, I will merge it later today.

kyrsjo:
Let's also see if there are other similar stuff that can be cleansed...

JamesMolson:
Other .s files have the same issue going on but I've not purged them yet.

kyrsjo:
You mean dabnew.s? plato.s (the FMA library) should be fine.

kyrsjo:
Hi can you bump the date in the +cd version block to TODAY, and add your name to the +cd license block? Once this I've done, I'll merge this + #118 #116 + close #119

JamesMolson:
You can close #13 too.

Are we sure the test suite covers the functions in dabnew.s ?

kyrsjo:
Hi,
I'm pretty sure they do not. Also, these files we pretty much never need to touch, so there may not be any very good reason to remove this stuff.

And urk. I did not realize, but you started from an older version; this makes the conflict when you change the license block (and is the reason why we only change the licence/date block at the last moment before merging.

So: Maybe you could do a git reset of this branch back to before 042472e , then force push to get rid of that commit on GitHub, then merge the current upstream/master into the branch, then change license block? For the `lielib.s`, I reviewed the changes manually, for `dabnew` I see that you killed two text comments as well as the old commented-out code.

I would actually really like to look through the `sixtrack.s` changes manually before merging, to look for other deleted comment blocks.

kyrsjo:
Thanks for the rollback!

PR #118, opened 2016-11-18
Warning: Rank mismatch in argument ‘x’ at (1) (rank-1 and scalar)

JamesMolson:
In dabnew and lielib compilers produce lots of rank mismatch warnings.

This is due to some routines expecting to be passed an array but instead getting an integer.

This changes all the cases where a single integer is passed.
The integer is changed to an integer array of size 1, and all references to the variables in question are updated.

It builds okay, but as always, I could have made a mistake somewhere.

See issue #13


kyrsjo:
I looked over the changes, they look good!

PR #131, opened 2016-12-02
Add the option to output to a specified file in read90

JamesMolson:
This should allow writing the read90 output to a file given by the --ofile option.

Could somebody please check I've not done something insane here :)

### DIFF
--- a/SixTest/read90/read90.f90
+++ b/SixTest/read90/read90.f90
@@ -1,4 +1,5 @@
       program read90
+      use, intrinsic :: iso_fortran_env, only : stdout=>output_unit
 !-----------------------------------------------------------------------
 !  Read a fort.90 and print with correct binary/decimal conversion
 !-----------------------------------------------------------------------
@@ -100,7 +101,7 @@
       integer ounit
 !----------------------------------------------------------------------
       ofoutput = .FALSE.
-      ounit = 6
+      ounit = stdout
       fname = "fort.190"
       
       STF = .false. !Use the new STF format?
@@ -428,6 +429,9 @@
       goto 550
   540 continue
   550 continue
+      if(ofoutput .eqv. .TRUE.) then
+         close(ounit)
+      endif
 !----------------------------------------------------------------------
       end
       integer function dtostr(x,results)
--- a/SixTest/read90/read90.f90
+++ b/SixTest/read90/read90.f90
@@ -95,8 +95,12 @@
       integer singleParticle, firstParticle, lastParticle
 
       character(len=100) :: fname
+      character(len=100) :: ofname
+      logical ofoutput
+      integer ounit
 !----------------------------------------------------------------------
-
+      ofoutput = .FALSE.
+      ounit = 6
       fname = "fort.190"
       
       STF = .false. !Use the new STF format?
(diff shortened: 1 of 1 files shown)


kyrsjo:
Looks mostly OK! Could you please close the output file (if not stdout)?

Also, stdout=6 is not really a standard (thus all the `write(*, ...` and `write(lout, ...` duplication in `sixtrack.s`). Apparently this has been solved in fortran2003 - could you try if it works with at least gfotran/ifort/nagfor?

Some materials:
http://fortranwiki.org/fortran/show/iso_fortran_env
http://stackoverflow.com/questions/8508590/standard-input-and-output-units-in-fortran-90

JamesMolson:
gfortran and ifort seem to be happy so far.

kyrsjo:
This is very cool! If we can confirm that this works with nagfor as well, we may get rid of the terrible if cr write lout / if not cr write * in sixtrack!!!!

JamesMolson:
Yeah, it works with nagfor too. In fact:
-f2008    Specify that the base language is Fortran 2008.  This is the default.

It defaults to fortran 2008


BUT
lf95 does not like it from what I have just tried:

  1333-S: "read90.f90", line 2, column 10: Name expected.
  2018-S: "read90.f90", line 104: When IMPLICIT NONE is specified, 'stdout' must be declared in a type declaration statement.

kyrsjo:
Which version of lf95? Are you sure you sourced something recent?

At least on their webpage (which mainly discusses Windows compilers, and if not for some of the post dates I would think it was un-maintaned since ca. anno 2000) http://www.lahey.com/ , Lahey claims to have "GF Silverthrone has full Fortran 95/90/77 compliance with extensive support for the Fortran 2003 and 2008 standards."

Could we do a workaround, something like `#ifndef __FORTRAN2003__ lout=6`?

JamesMolson:
I'm using /afs/cern.ch/sw/fortran/lahey/lf9562e

is there something newer out there?

kyrsjo:
Yeah, that's not THAT old:
$ ll /afs/cern.ch/sw/fortran/lahey/lf9562e/bin/lf95
-rwxr-xr-x. 1 1011 root 694089 Oct 29  2010 /afs/cern.ch/sw/fortran/lahey/lf9562e/bin/lf95

I think there should be a 2016 version tough?

kyrsjo:
Hmm... I wonder if there may be some flag we need to switch on - according to this post, Lahey has supported this feature for at least 10 years:
http://fortran.gcc.gnu.narkive.com/va4SIBxY/iso-fortran-env

kyrsjo:
I went digging a bit further, seems lf95 is version 6.2e, which came out in 2008, as the last release of the 6.2 series which seems to date back to 2003. There seems to be a v8.1 which is from 2009. Finally, there is the is the 7.x series, which is newer, but seems to be Windows-only -> nevermind... And they also seem to be pushing some kind of half-proprietary gfortran.

Trying to compile sixtrack with lf95 (after switching on datamodules, as lf95 also doesn't like it when that file is empty) fails miserably to compile sixvefox.f; this seems to have been introduced by 86d0ba85cf2281afae4e478b39359b2db3a5e301 . The failure seems to be related to the fact that DAFOR isn't actually compiled - I found this error message scrolling by when running make_six:
```
lf95 -m32 -o dafor dafor.o
dafor.o: In function `MAIN__':
(.text+0x1a0): undefined reference to `get_command_argument_'
dafor.o:(.data+0x8): undefined reference to `get_command_argument_'
Makefile_makesix:25: recipe for target 'dafor' failed
make: *** [dafor] Error 1
```

JamesMolson:
https://gcc.gnu.org/onlinedocs/gfortran/GET_005fCOMMAND_005fARGUMENT.html

GET_COMMAND_ARGUMENT — Get command line arguments

Description:
    Retrieve the NUMBER-th argument that was passed on the command line when the containing program was invoked.
Standard:
    **Fortran 2003 and later**

PR #134, opened 2016-12-05
Replace calls to DFLOAT with DBLE.

JamesMolson:
It seems DFLOAT is a language extension that converts an integer to a double.

This does not seem to be supported by nagfor any more.

The standard way seems to be to call DBLE instead.

This just replaces calls to DFLOAT with calls to DBLE.

### DIFF
--- a/SixTrack/make_six
+++ b/SixTrack/make_six
@@ -628,8 +628,8 @@ then
   ####export FCF="-$myopt -float-store -$myopt -C=undefined -abi=32 -dusty -maxcontin=60 -dcfuns -ieee=full -g90"
 # export FCF="-$myopt -C=all -abi=32 -dusty -maxcontin=60 -dcfuns -ieee=full -g90"
 # export FCF="-float-store -$myopt -abi=32 -dusty -maxcontin=60 -dcfuns -ieee=full -g90"
-  export FCF="-$myopt -abi=32 -dusty -maxcontin=60 -dcfuns -ieee=full -g90"
-  export FCF="-$myopt -abi=32 -dusty -maxcontin=60 -ieee=full -g90"
+  export FCF="-$myopt -abi=32 -dusty -dcfuns -ieee=full -g90"
+#  export FCF="-$myopt -abi=32 -dusty -maxcontin=60 -ieee=full -g90" ## Former default, could not build plato
 # The ieee=nonstd is INCOMPATIBLE with abi=32!!!
 # export FCF="-$myopt -abi=32 -dusty -maxcontin=60 -dcfuns -ieee=nonstd -g90"
   export FCL="-Bstatic -abi=32"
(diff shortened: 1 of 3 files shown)


kyrsjo:
Does the sixtest FMA test (especially extra_checks) still work? 

Also paging @mfittere ...

kyrsjo:
FMA test passed with gfortan and ifort. If you bump the date, we can merge!

JamesMolson:
Would you like make_six updating so that nagfor will build too?

kyrsjo:
I'll double check (manually) if it works with nagfor, if it does I'll merge. Will make another PR to make it build with nagfor and no datamods. Nagfor + datamods should work...

kyrsjo:
So I sent you a PR:
https://github.com/JamesMolson/SixTrack/pull/1

It fixes the problems with nagfor and make_six.

kyrsjo:
Ugh. I ran the fma test with a NAGFOR executable, and it failed miserably. It seems that it crashed just as it was calling fma_postpr:
```

----------------------------Runtime Error: sixve.f, line 31789: DIM argument (4096) out of range 1:1 in intrinsic MINVAL
Program terminated by fatal error
-------------------------------------------------------------------------------------------------------
         RANDOM SETS USED

  CASE  |  # OF RANDOM SET  |  MAX. POSSIBLE SETS   |     SEED
-----------------------------------------------------------------
    1              1                    1               100000
    2              1                    1               100000
    3              1                    1               100000
    4              1                    1               100000
    5              1                    1               100000
    6              1                    1               100000
    7              1                    1               100000
    8              1                    1               100000
    9              1                    1               100000
   10              1                    1               100000
   11              1                    1               100000
   12              1                    1               100000
   13              1                    1               100000
   14              1                    1               100000
   15              1                    1               100000
   16              1                    1               100000
   17              1                    1               100000
   18              1                    1               100000
   19              1                    1               100000
   20              1                    1               100000
   21              1                    1               100000
   22              1                    1               100000
   23              1                    1               100000
   24              1                    1               100000
   25              1                    1               100000
   26              1                    1               100000
   27              1                    1               100000
   28              1                    1               100000
   29              1                    1               100000
   30              1                    1               100000
-----------------------------------------------------------------

 Calling FMA_POSTPR
 start FMA analysis using file IP3_DUMP_1: number of particles= 60 , first turn= 1 , last turn= 4096
```

JamesMolson:
I have the same crash.

There is only 1 call to minval in the entire source:

                eps123_min(m)=minval(epsnxyzv(l,1:fma_nturn(i),m),      &
     &fma_nturn(i))      ! minimum emittance

mfittere:
in this line I calculate the minimum value of the normalized amplitudes (= `epsnxyzv`), the maximum and then the average in the following lines.

The differences in the tune calculation are most probably from the plato library which is not very well coded. One would have to find out in detail where the calculation goes wrong.

PR #137, opened 2016-12-06
Missing lostnumxv test folder

JamesMolson:
This just copies lostnumxvin to lostnumxv.


kyrsjo:
James confirms the test is also passing as is. Date of execution according to fort.6: June 3rd, 2012

Issue #138, opened 2016-12-06
FMA test fails on some compilers


kyrsjo:
There seems to be some problems with the FMA module, as it fails when compiled with certain compilers.

With nagfor it crashes - see #134 for the error message - the offending line being:
```
                eps123_min(m)=minval(epsnxyzv(l,1:fma_nturn(i),m),      &
     &fma_nturn(i))      ! minimum emittance
```

With newer versions of GCC (which are also 64 bit), TUNEFFTI gives slightly different tunes. @JamesMolson  is investigating this. I checked, and there are no parentheses missing or anything like that.

Miriam: Could you please give a quick description of what this line is supposed to be doing?

JamesMolson:
This test as 64bit works on my desktop. It fails as 64bit on lxplus. It passes on lxplus as 32bit.

fort.10, fort.90 always are good, all the DUMP files are good, it is just the fma_sixtrack in extra_checks.txt that has a few bits off.

kyrsjo:
It is very interesting that it works in 64-bit on your desktop, but not on LxPlus. Which compilers? Any system-provided libraries that we link with?

kyrsjo:
Note: I checked `plato_seq.s` for missing paranthesis. I did not recheck the relevant parts of `sixtrack.s`.

mfittere:
for the minval, maybe we have to add the dimension of the array:
[minval](https://gcc.gnu.org/onlinedocs/gfortran/MINVAL.html)

Can you try?

mfittere:
i = fma file = number of lines in FMA block
l = particle number
m = mode
1:fma_nturn(i) = number of turns used for the FMA analysis

e.g. if you do the FMA for fma_nturn(i)=10000 turns, the array epsnxyzv(l,1:fma_nturn(i),m) should have dimension 10 000.

Also looking at the line:
```
epsnxyzv(l,1:fma_nturn(i),m),      &
     &fma_nturn(i))
```
the right dimension of the array is given, so this should be correct!

Any other ideas?

JamesMolson:
> e.g. if you do the FMA for fma_nturn(i)=10000 turns, the array epsnxyzv(l,1:fma_nturn(i),m) should have dimension 10 000.


uh. No? It has a length of 10000, not 10000 dimensions...

It just runs from 1 to fma_nturn(i), so a 1 dimensional line.

kyrsjo:
[...]

PR #143, opened 2016-12-08
Fix FMA problems

kyrsjo:
For fixing #138


JamesMolson:
Do the tests still pass with complex*16 ?

kyrsjo:
OK, now it passes the tests with both nagfor and gfortran. @JamesMolson , could you test 		a05e3e1 with 64-bit on LXPLUS?

However, nagfor found me something REALLY worrying: In subroutine ZFUN, which is used by TUNENEWT, we have the following statement:
```
+if crlibm
      ZTUNE1=EXP_RN((-ZU*DUEPI)*TUNEA1) !!! EXP_RN expects a DOUBLE PRECISION, not COMPLEX !!!
+ei
+if .not.crlibm
      ZTUNE1=EXP((-ZU*DUEPI)*TUNEA1)
+ei
```

Note the comment (my recent addition) - `DUEPI` is 2*pi, TUNEA1 is some real number (DOUBLE PRECISSION); but ZU is the imaginary unit: `ZU=DCMPLX(0D0,1D0)`...

From nagfor I get the following oh-shit-shit-shit warning:
` Warning: plato_seq.f: Argument 1 to EXP_RN has data type DOUBLE PRECISION in reference from TUNELASK and COMPLEX(KIND(0d0)) in reference from ZFUN `
I guess we could rewrite it as `ZTUNE1=cos_rn(DUEPI*TUNEA1) + ZU*sin_rn(DUEPI*TUNEA1)`? However, expect a change of the results from tunenewt (which must have been wrong before, when compiling with crlibm).

kyrsjo:
The oh-shit warning was fixed. Tests are not affected.

It is still failing tunefft and tuneffti with nagfor tough... Very strange.

JamesMolson:
Ok, so currently (with the above patches) on lxplus:
32 bit gfortran passes
64bit does not.

32bit nagfor also fails.
64 bit ifort fails.

All failures are in the fma_sixtrack file for the TUNEFFTI output.

Nagfor seems to get all the comparisons for TUNEFFTI as bad, but only some are bad for gfortran and ifort.


but, on my main desktop (fedora23) 64bit gfortran does pass okay.

JamesMolson:
64 bit clang/gfortran on osx is good too.

kyrsjo:
James: It should now work 100%. Nagfor/32 and gfortran/32 agrees on my machine (to a new .canonical), could you test gfortran/64 on LxPlus?

Miriam&James: Please review my changes. I think I fixed a few bugs (especially the "calling exp(real) as exp(complex)", that was a nasty one). Let me know when I can bump the date (and version number probably) and we merge.

Issue #126, opened 2016-11-24
Beam Beam block trap error


kyrsjo:
Following up  #106 , we should modify the code of the beam beam input block to check that the xstr is present. If not present, set xstr=xang as done for old versions.

kyrsjo:
Fixed in #145

PR #145, opened 2016-12-12
Detect missing xstr in BEAM block

kyrsjo:
If missing, print WARNING and set xstr=xang.


kyrsjo:
I tested it using `SixTest/javier`, where the xstr is present. Only this and `javier_bignpart` actually contains the 2nd line of the BEAM block.

kyrsjo:
This fixes #126

kyrsjo:
@rdemaria : Javier and I agree that we should STOP and print error message if old file format is encountered. Agreed?

rdemaria:
No, I prefer the old style passes with warning because of many old tools...

kyrsjo:
OK, will do.

By the way, I have a small doubt about the manual; it says:
"Angle of the position of the slices in the boosted frame [rad] (i.e. X = Z sin(xstr ) cos(xplane),
Y = sin(xstr ) cos(xplane) ). In absence of crabbing user should make sure xstr=xang; in case
the xstr flag is not set then xstr=xang is assumed and a warning is printed (since version 4.5.45)."

I guess that should really be not just X=Z * sin etc., but also Y = Z * sin etc.?

rdemaria:
I guess I forgot a Z...

On Mon, Dec 19, 2016 at 1:17 PM, Kyrre Ness Sjøbæk <notifications@github.com

kyrsjo:
Can you push to my branch? TIL...

kyrsjo:
Ah, it's the "allow edits from maintainers" setting. Huh, TIL again.

rdemaria:
What is til?

kyrsjo:
TIL = Today I Learned :)

PR #148, opened 2016-12-13
Dump beam mean and sigma

kyrsjo:
New dump format for writing out <x>,<x'> etc. and <x*x>,<x*xp>, etc.

(not yet ready)


mfittere:
1. you need all the second moments of the beam for any further analysis, explicitly:
`<x_ix_j>, i,j=1,...,6`
where the average is taken over the step size indicated, e.g. I want to dump my distribution every 10^4 turns and I track e.g. 10^6, then the average is calculated every 10^4 turns over 10^4 turns.
2. for any further analysis of amplitudes etc, I think one then only needs the normalization matrix (to be checked in detail) -> dump it in the header of each file!

rdemaria:
I think it would be useful to output also the number of particles since
they might change during tracking

kyrsjo:
Hi all,

I calculate all the 2nd moments of the beam explicitly, just take a look in the diff (where I fill the xyz2 6x6 matrix). I also plan to write out the particle number (napx), see the (not yet complete) header.

Miriam, if you have some insight into the normalization matrix, I'll be happy to add that. And we need to check that the DUMP file opening etc. happen after the LINE block has been calculated?

amereghe:
Hi all,
If not too complicated, I would leave the possibility to compute averages every 10^4 turns, but only on the current turn...
Thanks!

kyrsjo:
Hi Alessio,

sorry if I misunderstand you, but: My plan was to just have it trigger at the interval set in the DUMP block (so every N turns, where N is very commonly set to 1), and when it triggers and format=3, I'll calculate the averages. Which I guess is what you want?

But: Does people also want to keep accumulating averages, then every N turns normalize, output, and zero it out? I guess this can be done also; for the input maybe one can make it accept "negative" N and this means "sum up over N turns, then output"?

One needs to put a bit of thought into how to normalize then, in the face of particle losses and numerical errors. The "cleanest" would probably be to calculate the averages every turn (normalized to the number of particles), then add it to a "global" accumulator. Then every N turns, this global accumulator is normalized by N, written out, and zero'ed.

amereghe:
Hello Kyrre,
I may have misunderstood @mfittere but I think that one of her remarks was to have averages every 10^4 turns over 10^4 turns, whereas I was raising the point that one may want averages every 10^4 turns over 1 turn...
Thanks!

mfittere:
I agree with @amereghe that we should have the option to NOT average and to average.

mfittere:
For the normalization there are two matrices in SixTrack:
1) from the 6d optics. Here is the piece of code I use to store it for the fma analysis:
```
            if(fma_flag) then
              if(ic(i)-nblo.gt.0) then !check if structure element is a block
                if(ldump(ic(i)-nblo)) then !check if particles are dumped at this element
                  dump_tas(ic(i)-nblo,ii-1,ii-1)=angp(1,ii-1)
...
```
2) the normalization matrix used for generating the initial coordinates `tas(npart,6,6)` (there is `ta` and `tas`, I would have to check which one to use once we are ready)

kyrsjo:
Hi, I will leave the averaging over many turns for later, since at least the naive implementation of this would require storing (6*6+6)*nele doubles somewhere in memory. This is ignoring the ALL element.

An alternative would be do this on disk, eg. write every turn, then at the end of the simulation "compress" the file. That would also be useful for collimation. On the other hand, one could also easily do this in an external program...

kyrsjo:
Regarding normalization: Yes, this would be nice; especially if we want to directly output e.g. emittance etc. - which becomes meaningfull with hugenpart.

But for now, I just want to make something functional (and use it for a study)... Then we can add extra bells and whistles later.

kyrsjo:
OK should be done now, could you take a look? Basically I'm writing out this:
```
4 – Beam mean and sigma:
Header: # napx turn s[m] <x>[mm] <xp>[mrad] <y>[mm] <yp>[mrad] <z>[mm] <dE/E>[1]
<x^2> <x*xp> <x*y> <x*yp> <x*z> <x*(dE/E)> <xp^2> <xp*y> <xp*yp> <xp*z> <xp*(dE/E)>
<y^2> <y*yp> <y*z> <y*(dE/E)> <yp^2> <yp*z> <yp*(dE/E)> <z^2> <z*(dE/E)> <(dE/E)^2>

A number of lines describing which elements are used and the current dump period is added one
per relevant line in DUMP block.
Data lines: As described in the header; one per turn (and for collimation, one per pack of par-
ticles). The for the “product” quantities, the units are the product of the units of the “normal”
ones.
```
I also fixed some stupidities with some other headers, which resulted in updating a few lines in a lot of canonicals.

If you agree, I'll merge this tomorrow morning, then fix #220 .

rdemaria:
Do we want to use canonical variables for the sigma matrix?

On Mon, Feb 20, 2017 at 6:22 PM, Kyrre Ness Sjøbæk <notifications@github.com

mfittere:
I would vote for canonical variables

mfittere:
I don't understand this: "A number of lines describing which elements are used and the current dump period is added one
per relevant line in DUMP block.", can you give an example?
Do I understand it correctly, that you calculate the moments for the number of particles tracked? E.g. I am tracking 32 particles, so I will calculate the moments over 32 particles?

kyrsjo:
I can make format 4 a bit simpler then: Only dump the means, which is what I actually need for a physics study I'm working on. Then push the sigma's (including normalization) into format 5. Format 4 will anyway be a useful little file for quick checks (especially combined with ALL), and I don't want to make it dependent on optics calculations etc. for it to work at all.

Regarding the manual:
If there are multiple single elements that write to the same DUMP file, I write a header like with the element name etc., then the header describing the columns. After that, for every element that wants to dump to the same file, I then write another header with the element name etc., like the first header line.  Then comes the data. This is true for format 2 as well.

rdemaria:
to use canonical variable you don't need the optics, it is just to use x,
px, y, py, sigma, psigma=Delta E/(beta0 P0 c) instead of the variable used
in format 2. To go to action/angle variable you then need the normalization
matrix that is optics dependent, which I am not even sure it is easily
accessible for all elements.




On Mon, Feb 20, 2017 at 9:27 PM, Kyrre Ness Sjøbæk <notifications@github.com

kyrsjo:
Ah, sorry; I misread the comment. Agreed that canoncial would be nice; I was indeed thinking of action angle (or rather, normalized) coordinates, which is not that easy... Will implement canonical!

I'm anyway pushing the full (canonical) matrix out to format 5, and having format 4 just dump the means.

kyrsjo:
Ok, now it should be done. Could people take a look? Apparently I have a hard time not mixing up y with yp etc. today...

I ended up adding 3 new formats:
 * Format 4: Only dump means
 * Format 5: Means and beam matrix
 * Format 6: Canonical means and beam matrix

I also cleaned up the basic DUMP file opening, sanity-checking, and header-writing code quite a bit. This resulted in that all headers for format 2 changed slightly... Thus the massive number of modified files.

Optics and averaging over multiple turns are not included now - for optics there are complications for 4D and needing the LINE block (probably); for averaging over multiple turns there are some issues (altough possible to fix at the cost of some code spaghetti OR dynamic memory allocations / pointers) in the basic architecture of DUMP.

SixTrack/sixtrack.s
```diff
@@ -26127,8 +26095,82 @@ C     Convert r(1), r(2) from U(0,1) -> rvec0 as Gaussian with cutoff mcut (#sig
...
+     &   '<yp^2> <yp*z> <yp*(dE/E)> '//
+     &   '<z^2> <z*(dE/E)> '//
+     &   '<(dE/E)^2>'
+                
+             else if ( dumpfmt(i).eq.6 ) then ! FORMAT 6
+                write(dumpunit(i),'(a)')
+     &               '# napx turn s[m] ' //
+     &   '<x>[m] <px>[1] <y>[m] <py>[1] <sigma>[m] <psigma>[1] '//
+     &   '<x^2> <x*px> <x*y> <x*py> <x*sigma> <x*psigma> '//
+     &   '<xp^2> <px*y> <px*py> <xp*sigma> <xp*psigma> '//
```
rdemaria: change xp with px

kyrsjo: Thanks, will do! Otherwise it looks OK?

kyrsjo:
OK, done. I'll bump the version number and merge!

PR #152, opened 2016-12-20
Documentation 

androula:
Would be good if in the documentation it was mentioned explicitly that SPECIAL ELEMENTS go in fort.3


kyrsjo:
Hi, you should try to merge *from* androula/SixTrack/:doc_update_web *to* SixTrack/SixTrack:master. :)

kyrsjo:
See #155 .

PR #154, opened 2016-12-20
FMA fixes: DUMP max turn = -1, 4D tracking

kyrsjo:
Hi Miriam @mfittere,

could you look at these modifications to FMA, especially the 4D tracking stuff?

Andri @androula crashed the program hard by running FMA on Thick4D results; I think I fixed it by basically just skipping the 3rd mode...

### DIFF
--- a/SixTrack/sixtrack.s
+++ b/SixTrack/sixtrack.s
@@ -57666,11 +57666,12 @@ c$$$            endif
 +if cr
 +ca crcoall
 +ei
-      integer :: i,j,k,l,m,n !for do loops
-      integer :: fma_npart,fma_tfirst,fma_tlast !local variables to check input files
-      logical :: lopen              !flag to check if file is already open
-      logical :: lexist             !flag to check if file fma_fname exists
-      logical :: lread              !flag for file reading
+      integer :: i,j,k,l,m,n                    ! for do loops
+      integer :: num_modes                      ! 3 for 6D tracking, 2 for 4D tracking.
+      integer :: fma_npart,fma_tfirst,fma_tlast ! local variables to check input files
+      logical :: lopen                          ! flag to check if file is already open
+      logical :: lexist                         ! flag to check if file fma_fname exists
+      logical :: lread                          ! flag for file reading
       character(len=getfields_l_max_string) :: ch,ch1
       character filefields_fields
      &     ( getfields_n_max_fields )*( getfields_l_max_string )
@@ -57748,6 +57749,12 @@ c$$$            endif
      &action='write',form='formatted')
       call fma_error(ierro,'cannot open file fma_sixtrack for writing!',&
      &'fma_postpr')
+
+      if (idp.eq.0 .or. ition.eq.0) then
+         num_modes = 2          !4D tracking
+      else
+         num_modes = 3          !6D tracking
+      endif
       
 !     write the header
       write(2001001,'(a)') '# eps1*,eps2*,eps3* all in 1.e-6*m, '//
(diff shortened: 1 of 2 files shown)


kyrsjo:
(I'm running the fma test now, it takes a while tough. Probably because I compiled without optimizations, duh...)

kyrsjo:
FMA test is a success! @mfittere , please check it and let me know. If it's OK, you can then either push the date update yourself (to my branch! as @rdemaria showed was possible. I didn't know...) and click The Big Green Button, or let you let me know and I'll do it.

mfittere:
I never tested the FMA for 4D. In general you can calculate the tunes without the normalization, but then you obtain Qx,Qy and all sidebands as the modes are not decoupled. With normalisation you get instead the eigenmodes Q1,Q2. The correct treatment is to use Q1,Q2.

If there is the need to run the FMA analysis also for 4D, I have to check the normalization matrix and do some tests. This might be very time intensive or might be also rather quick, hard to tell. I only do that if there is really need to run 4D instead of 6D. If we do that, I need some time and fix also the other issue along the way. This week I cannot, so it will be after Christmas.

SixTrack/sixtrack.s
```diff
@@ -57998,7 +58009,7 @@ c$$$            endif
 !     for fma_norm_flag = 0 use physical coordinates x,x',y,y',sig,dp/p
 !         fma_norm_flag > 0 use normalized coordinates
             do l=1,napx ! loop over particles
-              do m=1,3 ! loop over modes (hor.,vert.,long.)
+              do m=1,num_modes ! loop over modes (hor.,vert.,long.)
```
mfittere: This doesn't work as you then have to use the 4D matrix instead of the 6D matrix, otherwise the normalization is not correct. I would suggest to implement this properly and use for 4D tracking the 4D matrix and 6D tracking the 6D matrix.

mfittere:
All fine with me! Check that the FMA test passes (I think this you did already?) ...

kyrsjo:
Yeah, I checked. :)

androula:
Great!
I don't mind at all 6d. I don't know why sixtrack chooses 4d, I have SYNC in my fort.3... 

GEOM
PRINTOUT
NEXT
TRAC
600  0 32 0 2 0 1
 1 1 0 0 2
0 0 1 1 1 20000 2
NEXT
INITIAL COORDINATES
 2 0 0 1
1.0
1.0
1.0
1.0
0.0
0.0
1.0
1.0
1.0
1.0
0.0
0.0
7000000.0
7000000.0
7000000.0
NEXT
ITERATION ACCURACY
50 0.10E-13 0.10E-14
10 0.10E-09 0.10E-09
10 0.10E-04 0.10E-05
 0.10E-07 0.10E-11 0.10E-09
NEXT
BEAM
2.2E+11 3.5 3.5 0.0755 1.13E-04 1 1 1 1
NEXT
SYNC
35640 .0003225 16 0 26658.864 938.2796 1 0
1 1
NEXT
DUMP
ipmymarker.1 1 661 2 start-dump.txt
cc.1         1 662 2 cc1-dump.txt
cc.2         1 663 2 cc2-dump.txt
mymmq.1      1 664 2 mmq1-dump.txt
mymmq.2      1 665 2 mmq2-dump.txt
NEXT
LINE
ELEMENT 0 2 1 2.5 2.5
NEXT
FMA
start-dump.txt TUNEABT 1
NEXT
ENDE

mfittere:
Hi Androula,

so there is no need for 4D? Then we will leave it like this as it might 
be some effort to implement it for 4d too. Did you check that you have 
cavities?

Miriam

PR #157, opened 2016-12-24
Merge wire branch

mfittere:
Hey,

I had to create a new pull request as I couldn't change the remote branch. This is now Andrei's and Kyrre's branches merged. The wire tests run, I haven't tested any other compilers than gfortran. I will now start merging my part.


JamesMolson:
What is with all the extra folders in wire_thin6din

mfittere:
You mean the tests for the wire?

JamesMolson:
Yes, it seems to be a duplicate of everything.


Also unless I need something else in addition to this branch as is, the tests fail for me:

```
 MF: wirekick - wire_current(ix),wire_lint(ix),wire_lphys(ix),wire_flagco(ix),wire_dispx(ix),wire_dispy(ix),wire_tiltx(ix),wire_tilty(ix),wire_flagco(ix)   0.0000000000000000      
  0.0000000000000000        0.0000000000000000                0   0.0000000000000000        0.0000000000000000        0.0000000000000000        0.0000000000000000                0
ERROR: in wirekick -  wire_flagco defined in WIRE block must be either 1 or -1!
bez(195) = wire_1          
wire_flagco(195) = 0




         ++++++++++++++++++++++++
         +++++ERROR DETECTED+++++
         ++++++++++++++++++++++++
         RUN TERMINATED ABNORMALLY !!!
```

Each of the 4 tests dies like this.

mfittere:
They fail at the moment, as I just merged my branch and I have to modify the tests. They passed before

JamesMolson:
ah :)

mfittere:
I modified how the wire is defined, so now I am changing all fort.3 and fort.2 files of the tests.

mfittere:
Note: only wirethin6d works for now. I will do the rest of the tests tomorrow + remove all the printout and debug messages.

Kyrre, you will for sure still complain about the variable names and definition. There is still work to do, feel free to improve the code, I just didn't have enough time to do it all. Maybe wait with any changes until I have the tests.

Documenation is also still to missing!

JamesMolson:
```
    Start 38: wire_thin6d
1/1 Test #38: wire_thin6d ......................   Passed    1.57 sec
```
That is 64bit gfortran, I'll try with ifort and nagfor which usually will show if there are any write statements doing odd things.

mfittere:
I haven't even tried to compile it yet with any other option than 
gfortran. Changing write statements is ok, but please leave everything 
else the same. The code might be still in a fragile state. And don't 
remove any debugging comments either yet, I might still need them!

JamesMolson:
nagfor is good, ifort fails to run

```
         OOOOOOOOOOOOOOOOOOOOOO
         OO                  OO
         OO  SIXTRACK-INPUT  OO
         OO                  OO
         OOOOOOOOOOOOOOOOOOOOOO

          TITLE:ltrelens2016_coll%1%s%62.31_60.32%0_1%3%45           
         PROGRAM MODE : FREE FORMAT INPUT --READ FROM EXTRA GEOMETRY STRENGTH FILE--

forrtl: severe (64): input conversion error, unit -5, file Internal Formatted Read
Image              PC                Routine            Line        Source             
SixTrack_4545_crl  0000000000755486  Unknown               Unknown  Unknown
SixTrack_4545_crl  000000000077CDCF  Unknown               Unknown  Unknown
SixTrack_4545_crl  000000000077AD0D  Unknown               Unknown  Unknown
SixTrack_4545_crl  0000000000518298  Unknown               Unknown  Unknown
SixTrack_4545_crl  000000000052F2DB  Unknown               Unknown  Unknown
SixTrack_4545_crl  000000000040050E  Unknown               Unknown  Unknown
SixTrack_4545_crl  000000000082D44B  Unknown               Unknown  Unknown
SixTrack_4545_crl  00000000004003E9  Unknown               Unknown  Unknown
```

mfittere:
Thanks for checking! Yeah, I expected that there might be some errors ...

mfittere:
The current version passed the wire tests for gfortran, so I think we have first version to work with. In the following a list of things still to do.
**Variables:**
1. move variables out of big common block and put them in separate one, then rename wireco2 to wireco
2. wire_flagco corresponds to `windex`. Change `windex1(ix)` to have same values as wire_flagco or even entirely replace it by wire_flagco
3. rename variables if possible with wire_* named variables with more meaningful names (e.g. clowbeam + others)
**Comments:**
1. remove debug messages + write statements also commented ones
**Compilers**:
1. check for different compiles (I only tried ./make_six gfortran and ./make_six gfortran cr)
**Testing**:
1. check that thin4d,thck4d,thck6d really use the correct tracking routine
2. Check

kyrsjo:
Regarding the read90, could you change the build script to build a static binary? Then we'll avoid this problem in the future.

kyrsjo:
User manual:
 * Please describe the wire in subsubsection 3.2.1 Single Elements, as is also done for BEAM etc., and reference subsubsection 3.3.8 Wire in it.
 * 3.3.8, Description: "The wire module serves..." -> "The wire *block* serves..."
 * Table 3.7, tilt_x / tilt_y: Strange upside down exclamation mark?

kyrsjo:
Also user manual:
 * Please describe `flag_co` - what does +1 and -1 mean
 * Are you really using the `ibeco` variable from the BEAM block, as stated in the manual?

kyrsjo:
In block `+cd wirekick`:
 * In the `+if crlibm` / `+if .not.crlibm` blocks, there is a mixture of `_rn` trig functions and non-`_rn` functions. The precence of `_cr` functions in the `+if .not.crlibm` causes non-crlibm builds to fail, such as the collimation version.
 * Please indent the loops and Fortran if's.
 * one+dpsv(j) can also be written as dpd(j); it is already pre-calculated.

Other things:
 * I see that you are actually using a variable from another block. Does this mean that the BEAM block MUST be present for the WIRE block to be valid? Please (1) point this out in the WIRE block docs, and (2) in the BEAM block docs mention that `ibeco` also has an effect on WIRE.
 * Lots of little changes in indentation everywhere in the code. How did this happen? Please avoid this. I will try to figure out where it came from and fix it the right way (i.e. avoid that the change ever happened, otherwise `git blame` might say that you wrote half the code in some commit that only undid the change you did to the indentation earlier...
* In general, if you spot something else you want to change while writing e.g. WIRE, do not fix this in the wire branch. Rather create new branch from master, fix the issue there, make a pull request back to master (this is very quickly approved for small things), and merge master onto your dev branch.

kyrsjo:
Some questions about the version history:

Did you do a manual merge here?
https://github.com/mfittere/SixTrack/commit/7af0a08f9d8a62bea413b6817cdf3b28e507240c
That might be a problem, as it could screw up version control... Looking at the diff, it looks like it's basically a version bump from 4.5.29 to 4.5.45, including the code changes we did in that time period. I already did the version bump in https://github.com/kyrsjo/SixTrack/commit/e892f5a674c25e0cef6d1018e12725c47aa38d61 , so why? This is then followed by https://github.com/mfittere/SixTrack/commit/57ceb68a09e6b0c3997db1a514af5cfd26c1be94 , which is not a really a merge commit but which seems to repeat every change in sixtrack.s?

The history here really looks like a big mess, I don't want to merge it with how it looks. One way to fix it may be to scrap the history of this branch, copy the current source files as-is into a new branch starting from whatever version is closest, and before committing it remove all "extra" changes (spaces etc. in unrelated code). That will preserve the version control history of the various changes that has happened since 2.4.29, at the cost of loosing the history of the wire; making it look like one big commit followed by a few smaller ones to fix what is left to fix.

mfittere:
To answer your questions:

- I fixed the flag_co, the weird ! for tilt_x in the manual and added the wire AND the elens to the list of single elements. I added a comment for the ibeco in the beam block description and there is already a remark in the WIRE block description.
- yes, we are really using the ibeco flag from the beam block. The reason is that you want the wire and the beam-beam element to behave always in the same way. So I think it is the safest to just define one flag for the desired behaviour than introducing an additional one in the wire block which you have to again remember to assign the right value to. Anyway, these flags are always set to 1 anyway.
- what do you mean by that???

> In general, if you spot something else you want to change while writing e.g. WIRE, do not fix this in the wire branch. Rather create new branch from master, fix the issue there, make a pull request back to master (this is very quickly approved for small things), and merge master onto your dev branch.
- merging: the merge was a mess. When I tried to merge your branch into mine, it actually took the old sixtrack files and merged the new ones on top, so that all the !hr03 etc. which you had deleted where included again. Wanting to finish this and as an act of desperation, I just merged manually and then checked the changes. This is probably also the source of the different indentation and the resulting git blame issue. Probably, I should have checked out your branch into a new branch *new_wire* and then merged my wire branch onto the *new_wire* branch. The wire doesn't have too many changes, so maybe the least time intensive and in addition keeping the history right:

1. After doing all our changes, I merge the latest version of the SixTrack/master to the wire_merge_all branch
2. I check out the latest SixTrack/master to a new branch, e.g. 'wire_master'
3. 'copy' the changes manually to the sixtrack.s and the test files + documentation (that shouldn't take too long with just using meld)
4. do the full testing
5. merge this 'wire_master' branch which should have the right indentation + git blame final

- good spot with the crlibm. We will fix it.

- update on where we are: 

1. Andrei checked the tests again, as we changed the optics calculation and th fort.90 and fort.10 changed considerably. Good news: everything still works as it should
2. Andrei also updated the physics manual with a picture and better explanation. We will add all this tomorrow (today he is not here)

kyrsjo:
* Regarding the merging - I think we both have the same idea here. When creating the new branch, please just drop the SixTest/*testname*in folders.
 * Regarding the "what do you mean with this" - it was more a general remark that if you spot a bug in master while working on a feature branch, as a general rule it is better to fix this in master (i.e. create a new branch from master, fix the bug there, merge it to master, then merge master to your feature branch). This way, pull requests are smaller, and bugfixes gets into master faster.

Great work so far!

mfittere:
Do you have a better idea than to literally copy the files and then merge them manually with meld? Because I don't have any better idea which would get rid of all the small indents etc. and would fix the git blame? I already removed all the *in folders by just pulling from the new master, the wire ones are still missing - will do with the update of the tests.

We will finish everything today and then we'll test and then it is hopefully done.

kyrsjo:
No, that was basically the idea... I guess one could do a bunch of git magic to rewrite history, however that seems harder. Maybe if one could do a squashing of the unneccessary commits with `git rebase`, like we did to fix a similar mistake when DYNK was originally commited (sixtrack.s was commited with DOS line endings, changing every single line in the whole file...)?

mfittere:
Ok, I'll just do the manual one. Probably git rebase would work, but I think I am faster with just doing the manual diff.

kyrsjo:
Can we close this PR, since we now have #174 ?

mfittere:
yes, close it!

kyrsjo:
Close since we now have #174 .

PR #162, opened 2017-01-15
cmake updates.

JamesMolson:
This just adds some labels to the tests, so you can run ctest -L blabla and only get a subset of the tests. Each test can have multiple labels.

So far I have:

fast: for tests < ~60 seconds run time
medium: say < ~30 minutes run time
slow: everything else, i.e. prob1 and prob3 currently.


JamesMolson:
Ok, so this update should let you build BOINC on linux. Maybe other places too.

PR #163, opened 2017-01-16
Cleanup crlibm

kyrsjo:
Move our own codes out of the crlibm folder and into the roundctl folder, fix compilation on Windows by always using the local fpu_control.h. This contains inline assembly which sets up the FPU.

TODO: Fix so that it builds with make_six...


JamesMolson:
Hmm, not building as-is.

kyrsjo:
Works! On both Linux and mingw64.

kyrsjo:
.. and mingw32. Note that I had to install `mingw32/mingw-w64-i686-cmake` from pacman, otherwise it did not find the Platform file. Mac also worked, as long as static is disabled.

PR #165, opened 2017-01-17
SixTest: Delete all "TESTNAMEin" folders

kyrsjo:
This deletes all "TESTNAMEin" folders in SixTest, which are just copies of the TESTNAME folders.

The point of these folders was to make it easier to "reset" a test after running SixTest, however this can just as well be accomplished by simply
```
rm -rf TESTNAME
git checkout -- TESTNAME
```


JamesMolson:
Purge it with fire.

Issue #169, opened 2017-01-18
Test "lostnumxv" fails with CR


kyrsjo:
Both fort.90 and fort.10 are DIFF.

kyrsjo:
Note this commit:
https://github.com/SixTrack/SixTrack/commit/8af413b17b4920a4f4ab3fc0302b0a7520d24b40
It's from PR #137

JamesMolson:
a comment in sixtrack.s:

```
! If CR we have to worry about turns printed in fort.6
! If lost should be OK, otherwise we need to use nnuml instead
! of the numl in numxv/nnumxv???? Eric.
! where we reset [n]numxv to nnuml UNLESS particle lost
! Now we shall try using that fix at start of tracking
```

kyrsjo:
I ran this test in version v4.5.14 (the oldest tag we currently have), with the same result - It works for the normal version, DIFF for cr version.

JamesMolson:
Looking in the input file we have
```
TRACKING PARAMETERS-----------------------------------------------------
1000 0 30 2.500 6.500 0 1 0 0 50 50
1 1 0 1 2
0 0 1 1 1 50000 2
NEXT
```

Looking in the fort.90, tracking seems to stop after 50 turns, but up until that point tracking is the same.

What are the last 4 entries on line 1, the 0 0 50 50.

No other input fort.3 files seems to have these (grep ^TRAC SixTest/*/fort.3 -A1).
The manual says these are:
niu(1), niu(2)
numlcp
numlmax

but no description is provided.

Looking in sixtrack.s we have:
```
! Two new options for Turns to checkpoint and max turns/job.
      if(iclr.eq.1) read(ch1,*,round='nearest')                         &
     &numl,numlr,napx,amp(1),amp0,ird,imc,niu(1),niu(2),numlcp,numlmax
```
setting max turns/job to 50 when the canonical expects 1000 might cause problems.


Later in the code:
```
+if cr
!  does not call CRPOINT if restart=.true.
!  (and note that writebin does nothing if restart=.true.
          if(mod(numx,numlcp).eq.0) call callcrp()
          restart=.false.
```

and for numlmax
```
! and now reset numl to do only numlmax turns
      nnuml=min((numlcr/numlmax+1)*numlmax,numl)
      write (93,*) 'numlmax=',numlmax,' DO ',numlcr,nnuml
```
etc etc, so this test will only every do 50 turns in CR mode.


[...]

PR #170, opened 2017-01-18
Fix building of DATAMODS together with CR

kyrsjo:


kyrsjo:
I also added a canonical fort.90 to test lostnumxv, see #169 .

PR #171, opened 2017-01-18
Boinc submodule

kyrsjo:
Adding the SixTrac/boinc repository as a submodule pointing to a known-good commit; fix various things in CMakeLists.txt such as flag conflict detection and making it possible to build BOINC without API.


JamesMolson:
looks good!

Issue #39, opened 2016-01-13
Fort.3 ZIPIT block for BOINC


kyrsjo:
The block should contain a list of files to put in Sixout.zip. This would enable collimation etc. to run on BOINC. If no ZIPIT block present, we may want to have a "default list" of files? In case of no BOINC, the block shall not be valid (In DATEN, detect no BOINC flag and exit if ZIPIT block is present).

See: http://indico.cern.ch/event/476291/contribution/0/attachments/1209033/1762820/SixTrack_20160108.pptx

Assignments:
- @kyrsjo : DATEN modification and new global arrays
- @ericmcintosh / @kyrsjo : Update code to select the files to zip

rdemaria:
We could also think to disentangle the boinc library which contains a zip library, with another external zip library like http://www.libarchive.org/ and have the feature implemented regardless on boinc.

kyrsjo:
Moving discussion from email thread, in order to make the information more accessible:

LibArchive looks nice; another alternative is libzip; however I haven’t found any Fortran interface for it so I may have to write some wrapper code. If I end up looking at writing a pile of wrappers, I’ll contact the authors (tk PIGTAIL giga DOT or DOT at and dillo ÆTT nih DOT at) first in case someone has done this before (not unlikely) but not published it anywhere easily found from Google. If we do write a wrapper, I would also find it useful to actually distribute them, separately from SixTrack.

It seems that BOINC uses some version of info-zip to do it’s zipping. I looked at that, however they only distribute binaries (i.e. “zip” and “unzip” commands) and the sources for that, so some brain surgery would be needed. They released the compression library as zlib, which was then built on to make libzip.

It also seems that the BOINC client can be configured to do “all the work”, possibly making zip support in SixTrack redundant? https://boinc.berkeley.edu/trac/wiki/FileCompression

ericmcintosh:
Hi Kyrre, look forward to discussing this tomorrow.
I am a bit familiar with the BOINC stuff which I am using.
However I think your provision of a Fortran wrapper is
a good idea, rather easy to do as well.

I must also say I don't think we should be doing this in
SixTrack; it should be done in a script or .bat file.
Lots to talk over with you.     Eric.

kyrsjo:
I thought having it built into the binary would be a requirement for running it on BOINC?

We'll talk to tomorrow.

rdemaria:
Users need to choose the files to be returned after a sixtrack simulation
[...]

PR #173, opened 2017-01-24
Zipit

kyrsjo:
This links with the "libarchive" library, making the ZIPIT block #39 possible, and will replace the BOINC zip library.

Not yet ready to commit, however I don't mind some feedback.


kyrsjo:
I bumped into one problem for static linking:
```
[100%] Linking Fortran executable SixTrack_4545_libarchive_crlibm_fast_tilt_cmake_gfortran_static_64bit
../libarchive_build/libarchive/libarchive.a(archive_write_disk_set_standard_lookup.c.o): In function `lookup_gid':
/home/kyrsjo/code/SixTrack/SixTrack/libarchive/libarchive/archive_write_disk_set_standard_lookup.c:134: warning: Using 'getgrnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
../libarchive_build/libarchive/libarchive.a(archive_write_disk_set_standard_lookup.c.o): In function `lookup_uid':
/home/kyrsjo/code/SixTrack/SixTrack/libarchive/libarchive/archive_write_disk_set_standard_lookup.c:204: warning: Using 'getpwnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
[100%] Built target SixTrack
```

Looking at the `nm` output, these seem to be "weak" symbols:
```
 $ nm SixTrack_cmakesix_LIBARCHIVE_defaultcompiler_defaultbuildtype/SixTrack_4545_libarchive_crlibm_fast_tilt_cmake_gfortran_static_64bit  | grep grnam_r
00000000006f84f0 T __getgrnam_r
00000000006f84f0 W getgrnam_r
00000000006f84f0 T __new_getgrnam_r
0000000000700b20 T __nscd_getgrnam_r
$ nm SixTrack_cmakesix_LIBARCHIVE_defaultcompiler_defaultbuildtype/SixTrack_4545_libarchive_crlibm_fast_tilt_cmake_gfortran_static_64bit  | grep pwnam_r
00000000006f8770 T __getpwnam_r
00000000006f8770 W getpwnam_r
00000000006f8770 T __new_getpwnam_r
0000000000700460 T __nscd_getpwnam_r
```

I'm not sure if this is fine or if it is a problem...

rdemaria:
No clue... Have you tried to ask Laurent or test on Windows or other Linux?

JamesMolson:
Ah, I see you are adding it as a submodule, so we get the latest version. Everybody should be aware that there were lots of security vulnerabilities in previous versions of libarchive that were just fixed.

rdemaria:
Did you have a look to this as an alternative for instance?

https://github.com/kuba--/zip

kyrsjo:
For Windows, I need to get rid of boinc zip. Which is actually looking to be quite easy - just change one function call in `sixtrack.s` to point to SixTrack/libArchive_Fwrapper.c -> f_read_archive instead of `boinc_unzip()`.

Regarding submodule: I'm actually pinning it to a specific version (which is pretty much the newest + my patches to get it to compile on Windows).

Regarding the kuba--/zip, no I was not aware of that project. I already have libarchive working with SixTrack on Linux, and with my test program on both Linux and Windows. However somehow, it doesn't want to link when using CMake on Windows...

kyrsjo:
Regarding libarchive security: AFAIK, we'll be treating trusted data here, so I'm not too worried. Anyway I'm using a pretty small subset of it - note the string of -DOPTION=OFF that I'm passing to the libarchive cmake...

JamesMolson:
mingw64 windows link errors?

SixTrack/buildLibraries.sh
```diff
@@ -0,0 +1,20 @@
...
+git submodule update
+
+### BOINC
+
+### libArchive
+rm -rf libarchive_build
+mkdir libarchive_build
+cd libarchive_build
+
+cmake -DENABLE_BZip2=ON -DENABLE_CAT=OFF -DENABLE_CPIO=OFF -DENABLE_EXPAT=OFF -DENABLE_INSTALL=OFF -DENABLE_LIBXML2=OFF -DENABLE_LZMA=OFF -DENABLE_NETTLE=OFF -DENABLE_OPENSSL=OFF -DENABLE_TAR=OFF -DENABLE_CNG=OFF -G "Unix Makefiles" ../libarchive -LH
```
JamesMolson: Should this not be passed ${CMAKE_GENERATOR} as an argument, assuming  -G "Unix Makefiles" is always the case might cause future problems.

kyrsjo: Hmm, that may be a good idea. Only thing is that this isn't really running within cmake, it's just a shell script to make things reproducible, kindof like cmake_six (which it may eventually be integrated into).

Another possibility would be to somehow put it into the main cmake file, similarly to what was done with crlibm?

CMakeLists.txt
```diff
@@ -851,6 +860,32 @@ if(BOINC)
...
+		find_library(Bcrypt STATIC Bcrypt)
+		find_library(iconv STATIC iconv)
+		target_link_libraries(SixTrack libarchive bz2 Bcrypt pthread iconv)
+	else()
+		set_target_properties(libarchive PROPERTIES IMPORTED_LOCATION ${LIBARCHIVE_BUILD_DIR}/libarchive/libarchive.a)
+		target_link_libraries(SixTrack libarchive)
+
+		find_library(pthread STATIC pthread)
+		find_library(z STATIC z)
+		target_link_libraries(SixTrack z pthread)
```
JamesMolson: I'm needing this there:

                find_library(bz2 STATIC bz2)
                target_link_libraries(SixTrack bz2 z pthread)

kyrsjo:
OK, now I can compile BOINC + LIBARCHIVE + CR on both Linux and Windows (MINGW64), and it works as expected (unzipping Sixin.zip at the start of the run). Hooray!

The problem was that on Windows I have to link it as `-Llibarchive_build/libarchive -larchive` while Linux was happy with just linking the `libarchive.a` file directly as if it was an object file (which is what we do with `crlibm` on all platforms).

This is because Windows needs a "trampoline" to link the static library to another external library (here `zlib`) - in the folder `libarchive_build/libarchive` we have both `libarchive_static.a` and `libarchive.dll.a` - one is the actual code, the other is the trampoline.

I found this post to be quite helpful:
http://stackoverflow.com/questions/3704374/linking-error-lnk2019-in-msvc-unresolved-symbols-with-imp-prefix-but-shoul

JamesMolson:
With FreeBSD:
boinc + cr + api + libarchive
```
[100%] Linking Fortran executable SixTrack_4545_libarchive_cr_boinc_api_crlibm_fast_tilt_cmake_gfortran_static_64bit
//usr/lib/libc.a(gdtoa_strtod.o): In function `strtod':
gdtoa_strtod.c:(.text+0x1580): multiple definition of `strtod'
SixTrack/roundctl/libroundctl.a(dtoa_c.c.o):/home/molson/SixTrack/SixTrack/roundctl/dtoa_c.c:(.text+0x0): first defined here
collect2: error: ld returned 1 exit status
```

standard build is fine.

kyrsjo:
OK, I think we have something reasonably solid! It builds correctly with and without BOINC and API, on Linux, Windows (x64), and FreeBSD. Will check Windows (x32) and Mac.

Then I need to update the build manual, run tests, silence the C-code (it's a bit chatty to stdout), and then I think we can merge!

The actual ZIPIT block (for output) I will do in a new branch; however I think disentangling the libraries were probably more work than writing a fort.3 block + some hooks...

Since there's been a huge amount of new features lately, a BOINC release is around the corner, and with this branch *everything now finally builds correctly* - maybe it's time to bump the version - maybe even from 4.5.45 to 4.6.0 ?

JamesMolson:
I'm checking this on a mac now.

rdemaria:
Very good! For the new version I would stick with:

Given a version number MAJOR.MINOR.PATCH, increment the:
    MAJOR version when you make incompatible API changes,
    MINOR version when you add functionality in a backwards-compatible manner, and
    PATCH version when you make backwards-compatible bug fixes.

So definitely 4.6.0. I think we should have 5.0.0 when collimation branch is merged...

kyrsjo:
If we're using that scheme, we should have bumped the minor a BUNCH of times the last couple of years... And I guess with API changes are = input file changes?

Otherwise agree!

rdemaria:
I know, we should have 4.60.0... I think the "semantic" scheme is very useful, and it is a good to start. I agree API is input file at the moment.

kyrsjo:
And there, Win32 builds as well. Just had to fix a bug in boinc and another bug in libarchive. Once fixed, SixTrack built on the first try :D

kyrsjo:
@JamesMolson is working on the OS X build; there seems to be two problems:
 * Building with API
 * Building an actually static executable - gfortran from brew REALLY wants to dynamically link the libquadmath dylib even if a static version is available. The "fix" is to delete the dylib, as described here: http://stackoverflow.com/questions/17590525/correct-way-to-statically-link-in-gfortran-libraries-on-osx

See also PR #177 for OsX support.

JamesMolson:
there is a symlink to the real .dylib that you can just rename, rather than deleting it.

kyrsjo:
Thanks, this is great!

Could you add one more line in the doc for how to fix the libquadmath issue, more than just "There have been issues with dynamic linking of libquadmath.dylib and other items in your /usr/local folders.", but actually describe the hack?

Once this is done, it can be merged.

kyrsjo:
Left to do:
 * Update build manual
 * Make sure we're building the libraries with the right flags (optimization etc.)
 * Run tests
 * Bump version number :)

kyrsjo:
I ran CTEST with the BOINC exec on Linux ( `SixTrack_cmakesix_API_BOINC_BUILD_TESTING_CR_LIBARCHIVE_defaultcompiler_defaultbuildtype` ), all passed except for `lostnumxv` (expected to fail with CR) and prob1/prob3 (not expected to fail).

The reason `prob1` and `prob3` failed is that once the `KillTimes` array is "used up", it stops restarting the process. I suggest that it should keep restarting (using the last time in the array) until the simulation is finished OR some kind of limit is reached. One way to fix this would be to just make the KillTimes array (from the `SixTest/CMakeLists.txt` -> `CR_KILL_TIME` right?) longer...

Another thing: Isn't the `KillSixTrack` function somewhat redundant now?

JamesMolson:
Yes it is a bit redundant, it can be removed if the kills are but into the main()

For prob1/3, maybe your 2^n idea is the way to go?
Or a constant every few seconds.

JamesMolson:
Did you also have speed issues when running the boinc build?

kyrsjo:
Let's leave the performance issue + mac until a later merge (new branch). I'll finish the docs and one of us finishes the ctest stuff, then I'll bump the version number and merge. This branch will then continue with implementing the zipout part of zipit, which will be merged later.

JamesMolson:
Which ctest stuff are you talking about here?

PR #174, opened 2017-01-25
Wire merge with SixTrack master

mfittere:
I did a manual merge of the wire_merge_all branch. All tests for the wire pass, now I still need to test it with the new makefiles and the test suite. Manuals are also updated including also the Twiki pages.

**Open issues:**

1. There are the following lines in `sixtrack.s` of which I don't know which version is correct:
sixtrack/master:
```
!--SKEW ELEMENTS
  320   kzz=-kzz
        goto(330,340,350,360,370,380,390,400,410,420),kzz
        goto 440
```
miriam/wire_merge_all:
```
  320   kzz=-kzz
        goto(330,340,350,360,370,380,390,400,410,420,                   &
     &       440,440,440,440,440),kzz
        goto 440
```
same for:
sixtrack/master:
```
!--SKEW ELEMENTS
  220   kzz=-kzz
        goto(230,240,250,260,270,280,290,300,310,320),kzz
        goto 350

```
miriam/wire_merge_all:
```
!--SKEW ELEMENTS
  220   kzz=-kzz
        goto(230,240,250,260,270,280,290,300,310,320),kzz
        goto 350
```
2. the optics calculation in 6D for wire_thck6d fails


kyrsjo:
Regarding the first GOTO, they should be equivalent - if kzz (after sign is flipped) is in the range 1-10 (i.e. "nonlinear elements"), then goto `330,340,350,360,370,380,390,400,410,420`, if it is bigger then goto 440. Only difference I can see is that in the wire_merge_all code, there are special hardcoded cases for kzz in the range 11-15 as well, which all goes to 440. I would prefer the variant in master, as it's clearer.

The second GOTO: I think those code pieces look identical? Did you copy/paste the same code twice, or is it just time for me to go home?

mfittere:
It appears two times in the code. I also thought that it does exactly 
the same, but wanted to check with you that there is no hidden reason I 
am missing. I will then take the master version as you suggested.

kyrsjo:
Great. Regarding "two times", I meant that the second two gray boxes are identical, and you're asking me if they are different.

I take it they should be different in the same way the two upper gray boxes in your post are different from each other?

mfittere:
Yes, sorry about that. It was late yesterday :-D!

mfittere:
The crlibm is still not working. I get differences between crlibm version and not crlibm version ...

kyrsjo:
Great. It's late here too, just wanted to make sure that *I* wasn't seeing double :)

Regarding CRLIBM, that's expected - however CRLIBM is the same on different platforms and compilers.

mfittere:
I thought that for the tests you should at least get the same results with and without crlibm? I was just trying the collimation version thin4d and thin6d. How do you then check the .not.crlibm part?

kyrsjo:
The .not.crlibm part is difficult to check, precisely because it is NOT reproducible. I.e. you can check the physics manually like you did for the e-lens, but it won't stay the same forever afterwards and everywhere (which makes for easy automatic black-box checks).

But because of collimation, we can't really drop the .not.crlibm version either. It's a bit of a mess really... Ideally we could maybe write all code as if crlibm was there, and then have "trampoline" functions that just redirects sin_rn -> sin etc. whenever crlibm isn't there. However if we do this, we have to be very carefull with performance.

mfittere:
That's really bad. Because you can always do a copy paste error when generating the .not.crlibm part. Why would it degrade the performance? You think that it has to choose the function at each step would make such a time difference?

I checked the dump files of the thin6d not crlibm version and at least there are only very small differences in the last digit. So it should be fine.

Tthe dump files are strange as I get negative turn numbers??? Here:
```
 # DUMP format #2, bez=op0             , number of particles=          64
 # dump period=           1 , first turn=           1 , last turn=          -1
 # ID turn s[m] x[mm] xp[mrad] y[mm] yp[mrad] z[mm] dE/E[1] ktrack
      -63        1   6815.26943  1.006760057E+00 -7.744159896E-03  2.447027357E+00 -2.568668660E-03  0.000000000E+00  1.432803961E-16       31
      -62        1   6815.26943  1.006754046E+00 -7.744057573E-03  2.447027344E+00 -2.568668620E-03  0.000000000E+00  1.432803961E-16       31
      -61        1   6815.26943  7.181878380E-01 -2.229870082E-03  2.492883084E+00 -2.243325368E-03  0.000000000E+00  1.432803961E-16       31
      -60        1   6815.26943  7.181818278E-01 -2.229767762E-03  2.492883064E+00 -2.243325318E-03  0.000000000E+00  1.432803961E-16       31
      -59        1   6815.26943  4.292755866E-01  3.290448005E-03  2.537221081E+00 -1.915758297E-03  0.000000000E+00  1.432803961E-16       31
      -58        1   6815.26943  4.292695777E-01  3.290550316E-03  2.537221054E+00 -1.915758235E-03  0.000000000E+00  1.432803961E-16       31

```

kyrsjo:
Yes, it's not ideal. However basically rewriting all kick modules + another piece of infrastructure.. isn't such a high priority really.

Performance wise, I don't think it would be much worse than crlibm, depending on how this kind of stub is actually linked. You might still end up with an extra function call tough, which *might* be expensive. Also, I would not be surprised if many Fortran compilers can do lots of tricks with normal math functions, inlining or vectorizing them.

Negative turn numbers... THAT I have never seen before. I will check!

mfittere:
So the x,y,z etc. are the same, it is only that the turn numbers are negative. Thanks for checking! I am not trying to get the new cmake and ctest working and will do the final tests as everything else is now finished :-)!!!

JamesMolson:
```
1/5 Test #40: wire_thin4d ......................   Passed   12.40 sec
2/5 Test #38: wire_thck4d ......................   Passed   13.70 sec
3/5 Test #39: wire_thck6d ......................   Passed   16.13 sec
4/5 Test #41: wire_thin6d ......................   Passed   30.34 sec
5/5 Test #42: wire_thin6d_ibeco0 ...............   Passed   30.92 sec

100% tests passed, 0 tests failed out of 5
```

Seems to be good on freebsd

kyrsjo:
CMAKE is actually ridiculously easy. Just say `./cmake_six` instead of `./make_six gfortran`. It will even build in parallel, which saves you a bunch of waiting on make...

If you want to add flags, it works in the same way; i.e. to build a collimation version it's `./cmake_six COLLIMAT -CRLIBM` instead of `./make_six gfortran collimat`. Note that while make_six automatically disables crlibm when it sees collimat, cmake_six require you to do it manually so that you know what's going on (it will actually tell you to do it).

Switching compilers works in the same way as `make_six`. Try to run `./cmake_six help`!

If you're running on LxPlus, please load the newest cmake before starting. To do this, simply run:
```
export PATH=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64/bin:$PATH
export CMAKE_ROOT=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64
```

mfittere:
I did actually the following setup. Let me know if this is the right way to go:
```
# load newer version of cmake
export PATH=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64/bin:$PATH
export CMAKE_ROOT=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64

# ifort and icc/icpc
# 64-bit
source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh
# 32-bit
#source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh ia32

# nagfor
source /afs/cern.ch/sw/fortran/nag/usenag.bash 6.0

```

mfittere:
... and thanks for the explanation!!!
I already compiled with gfortran, gfortran cr . It worked fine!

mfittere:
I fail with ifort:
`cmake_six ifort BUILD_TESTING`
giving the error message 
```
  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:15 (ENABLE_LANGUAGE)


-- Configuring incomplete, errors occurred!
See also "/afs/cern.ch/user/m/mfitter1/opt/SixTrack/SixTrack/SixTrack_cmakesix_BUILD_TESTING_ifort_defaultbuildtype/CMakeFiles/CMakeOutput.log".
See also "/afs/cern.ch/user/m/mfitter1/opt/SixTrack/SixTrack/SixTrack_cmakesix_BUILD_TESTING_ifort_defaultbuildtype/CMakeFiles/CMakeError.log".
CMake Error: your Fortran compiler: "ifort" was not found.   Please set CMAKE_Fortran_COMPILER to a valid compiler path or name.

ERROR in CMAKE;
Most likely you specified a bad argument!
Please run cmake_six help
```

mfittere:
If it helps, I usually used this for the ifort:
```
source /afs/cern.ch/sw/IntelSoftware/linux/setup.sh 
source /afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2011/composerxe-2011.2.137/bin/ifortvars.sh ia32 
source /afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2011/composerxe-2011.2.137/bin/iccvars.sh ia32 
```

mfittere:
same for nagfor:
`cmake_six nagfor BUILD_TESTING`
gives
```
CMake Error: your Fortran compiler: "nagfor" was not found.   Please set CMAKE_Fortran_COMPILER to a valid compiler path or name.
CMake Error: Internal CMake error, TryCompile configure of cmake failed
-- Check for working Fortran compiler: nagfor  -- broken
CMake Error at /usr/share/cmake/Modules/CMakeTestFortranCompiler.cmake:54 (message):
  The Fortran compiler "nagfor" is not able to compile a simple test program.

  It fails with the following output:

   

  

  CMake will not be able to correctly generate this project.
Call Stack (most recent call first):
  CMakeLists.txt:15 (ENABLE_LANGUAGE)

```

kyrsjo:
Does the `nagfor` or `ifort` commands do anything?

From the build manual:

> 2.1.2 ifort and icc/icpc
Another popular set of compilers are Intel’s ifort and icc. To load these from
AFS, simply run their setup script as
source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh
if you want yo produce a 64-bit executable, or
source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh ia32
for a 32-bit executable.
However note that if using icc/icpc, it is not possible to compile static 64-
bit executables, since Intel does not provide a static version of the libcilkrts
library for 64-bit.
If running the Intel compilers on LxPlus, it is recommended to first load a
newer version CMAKE. This can be accomplished through the commands:
export PATH=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64/bin:$PATH
export CMAKE_ROOT=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64

And also:
> 2.1.3 nagfor
To load nagfor version 6.0 on LxPlus, simply run
source /afs/cern.ch/sw/fortran/nag/usenag.bash 6.0
before any compilation commands. Both 32- and 64-bit executables are sup-
ported, as well as both static- and dynamic linking.

kyrsjo:
By the way, you're loading an ancient version of ifort, in 32-bit mode. 64-bit mode is default in the CMake setup. If you want 32-bit mode, please add "32BIT -64BIT" to the `cmake_six` command line.

mfittere:
It didn't find the compilers. I had to put '' around the path. Now I get the following error:
Scanning dependencies of target test_copy_elensidealthck6d
```
make[2]: *** [test_copy_notilt] Segmentation fault
make[1]: *** [SixTest/CMakeFiles/test_copy_notilt.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
Scanning dependencies of target test_copy_fma
make[2]: *** [test_copy_thick6ddynk] Segmentation fault
make[1]: *** [SixTest/CMakeFiles/test_copy_thick6ddynk.dir/all] Error 2
[ 75%] Built target test_copy_dynk_globalvars
[ 75%] Built target test_copy_tilt
[ 75%] Built target test_copy_lostnotilt
[ 75%] Built target test_copy_thick6dblocks
[ 75%] Built target test_copy_crabamp
[ 75%] Built target test_copy_bb_ntwin1
[ 75%] Built target test_copy_s316
make[2]: *** [test_copy_wire_thck6d] Segmentation fault
make[1]: *** [SixTest/CMakeFiles/test_copy_wire_thck6d.dir/all] Error 2
[ 75%] Built target test_copy_thick4
[ 75%] Built target test_copy_frs60
[ 75%] Built target test_copy_lost
[ 75%] Built target test_copy_bbe571ib0
/bin/sh: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
[ 75%] Built target test_copy_lostevery
[ 75%] Built target test_copy_bbe51
[ 75%] Built target test_copy_bbe52
[ 75%] Built target test_copy_distance
[ 75%] Built target test_copy_exact
[ 75%] Built target test_copy_dipedge
[ 75%] Built target test_copy_lostnumxv
[ 75%] Built target test_copy_prob1
[ 75%] Built target test_copy_bb
[ 75%] Built target test_copy_eric
[ 75%] Built target test_copy_prob3
[ 75%] Built target test_copy_javier
[ 75%] Built target test_copy_frs
[ 75%] Built target test_copy_elensidealthck6d
[ 75%] Built target test_copy_elensidealthin4d
[ 75%] Built target test_copy_elensidealthck4d
[ 75%] Built target test_copy_wire_thin6d
[ 75%] Built target run_astuce
[ 75%] Built target test_copy_wire_thck4d
[ 75%] Built target test_copy_wire_thin4d
[ 75%] Built target test_copy_wire_thin6d_ibeco0
[ 75%] Built target test_copy_fma
[ 75%] Built target test_copy_elensidealthin6d
make: *** [all] Error 2

ERROR in MAKE;
Could be a bug...
.. or maybe you need to install some libraries?

```

kyrsjo:
In which file do you get negative ID numbers? What is the build command?

kyrsjo:
Did you load the newer cmake version?

kyrsjo:
Is this on LxPlus?

By the way, when running the tests (thin6d and thck4d), I get the following warning:
`Note: The following floating-point exceptions are signalling: IEEE_UNDERFLOW_FLAG IEEE_DENORMAL`
I ran them simply by going to the folder, symlinking in the binary, and `./SixTrack_blabla > fort.6`

mfittere:
I go the same warning and ignored it as I thought it is normal.

JamesMolson:
Oh that looks like the random segfault that was happening on lxplus, I think it was afs being silly since it doesn't happen anywhere else.

mfittere:
I loaded the new cmake, I did exactly that:
source /afs/cern.ch/sw/IntelSoftware/linux/17-all-setup.sh
export PATH=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64/bin:$PATH
export CMAKE_ROOT=/afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64

mfittere:
Hm, I thought I got the negative numbers with the collimation version. BUT, this shouldn't even work as I don't have a collimation block! I have to check on that, very bizarre.

mfittere:
I have it with the negative numbers!
test `wire_thin4d`
and compiled with
./make_six gfortran collimat

mfittere:
... and it shouldn't even run as there is no collimation block!

mfittere:
Any answers for the error in the cmake???

JamesMolson:
It was make itself segfaulting

make[2]: *** [test_copy_wire_thck6d] Segmentation fault


it seems to randomly happen and only on lxplus and I can't work out why due to the random nature.

just re-run the command and it will most likely work.

mfittere:
`cmake_six ifort CR BUILD_TESTING` still fails

It is strange. For each configuration (ifort, ifort cr, nagfor, nagfor cr) I have to run it twice to compile as it would generate some files the first time which it then uses for the second run.

JamesMolson:
um, I just tried nagfor + build_testing and it worked first time,

let me look at ifort. Its still using make -j so you might be hitting some resource limit or something like that. Let me check.

and running it twice is strange, because using the cmake_six script it should rm the folder each time.

JamesMolson:
hmm, I just tried what you said:

./cmake_six ifort CR BUILD_TESTING

and it worked perfectly the first time, so I'm not sure what is going on.
I did drop the make command down to -j4 though...

JamesMolson:
I should also note that CR + ctest is somewhat broken right now, and I'm going to fix it tomorrow so wait until then to play with that.

mfittere:
OK! Anything else? I try with the -j ...

Sent from my mobile phone.




On Wed, Jan 25, 2017 at 4:51 PM -0600, "James Molson" <notifications@github.com<mailto:notifications@github.com>> wrote:


I should also note that CR + ctest is somewhat broken right now, and I'm going to fix it tomorrow so wait until then to play with that.

mfittere:
Summary of the compiler tests:
- the following ones worked:
```
cmake_six gfortran BUILD_TESTING
cmake_six ifort BUILD_TESTING
cmake_six nagfor
```- the following failed:
`cmake_six nagfor BUILD_TESTING
`
with the error message:
```
/bin/sh: line 1: 27161 Segmentation fault      /afs/cern.ch/sw/lcg/contrib/CMake/3.7.0/Linux-x86_64/bin/cmake -E create_symlink /afs/cern.ch/user/m/mfitter1/opt/SixTrack/SixTrack/SixTrack_cmakesix_BUILD_TESTING_nagfor_defaultbuildtype/checkf10 /afs/cern.ch/user/m/mfitter1/opt/SixTrack/SixTrack/SixTrack_cmakesix_BUILD_TESTING_nagfor_defaultbuildtype/SixTest/exact/checkf10
make[2]: *** [test_copy_exact] Error 139
make[1]: *** [SixTest/CMakeFiles/test_copy_exact.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
...
make: *** [all] Error 2

ERROR in MAKE;
Could be a bug...
.. or maybe you need to install some libraries?
```
- the CR version I didn't retest, let me know once the ctest is working again also for the cr.

**Comment by JamesMolson** [comment]
The segfault is only on lxplus and it is random and only when simlinking a lot of files at the same time. It isn't anything to do with the cmake settings or compilers or anything else.

**Comment by mfittere** [comment]
for cmake_six gfortran I get the following failed test:
```
 6/42 Test  #6: SelfTest_read90 ..................***Failed  Required regular expression not found.Regex=[At line .* of file .*read90.f90 \(unit = 190\)
Fortran runtime error: Cannot open file 'fort.190': No such file or directory
]  0.00 sec
```
Everything else seems to have passed (prob3 is still running)

**Comment by kyrsjo** [comment]
Regarding the build errors with `cmake_six` on lxplus: I think the main problem here is AFS, which apparently doesn't like it when we try to run parallel builds with `make -j`. I fixed it for the zipit branch here:
https://github.com/SixTrack/SixTrack/pull/173/commits/7a22e04c4a4d67540aa7d8a0e13bbf75c99e09a1

You can change your cmake_six to build with `make` instead of `make -j` or `make -j 8`; however please do not commit this as it will cause a conflict when we merge the changes in ZIPIT with the changes in in wire-merge-master.

**Comment by kyrsjo** [comment]
Hi, if you merge the current upstream master into this branch, your cmake_six problems should be all fixed. cmake_six now contains AFS workarounds :)

If you want to develop on the mac, take a look at the very short howto in the build manual - look in the build_manual2 branch (+ I'll send you the pdf). Basically, you need to (1) install brew and (2) install 6 packages.

**Comment by kyrsjo** [comment]
Hmm, where do we have a conflict?

**Comment by mfittere** [comment]
I will do the merge of the newest version and check the merge conflicts. I then push the new version.

**Comment by mfittere** [comment]
It just didn't like the new wire block and your comment. All fixed now.

**Comment by kyrsjo** [comment]
Thanks! The SelfTest failures with ifort were fixed in #194. The gfortran one was fixed somewhat earlier. The collimation block oopsie is fixed in #193 . The segfaults were due to AFS. cmake_six is now aware of this; for ctest don't run with the "-j" (parallel) option.

Regardning the conflict: I guess you mean my new comment where I point out that "this is the ENDE"? (omnious music...)

**Comment by mfittere** [comment]
yes, exactly. It is weird that it marked it as conflict. Usually it understands these things ;-). What about the cr versions? I remember that we were almost fine for the non-cr, but there were issues for the  cr.

**Comment by mfittere** [comment]
Summary tests:
- gfortran: all tests passed
- ifort: thick6ddynk difference in dump, fort.90 and fort.10 pass
- gfortran cr: lostnumxv fails as expected
- ifort cr: lostnumxv fails as expected

**Comment by kyrsjo** [comment]
I confirm the problem with DUMP files for thick6ddynk on 32-bit, version 17, upstream sixtrack version bbfcb44c255cf235d7d624a71cbbcdf6628265f3 . A diff reveals that the problem only occurs intermittently in the last digit of the dump files, while fort.90 etc. are fine. This again indicates that the problem is in writing the dump files, not the tracking.

Inspecting the subroutine `dump_beam_population`, I see we have several straight (standard fortran) write statements with doubles. The double->ascii conversion thus depends on the compiler, and apparently ifort32 does this slightly differently. Furthermore, it is not enclosed in `enable_xp`/`disable_xp`. To fix this would mean to put a bunch of `dtostr` in this function whenever crlibm is active. This problem is already known please see #144 .

For the wire, please just go ahead and ignore this problem.

**Comment by kyrsjo** [comment]
The CR error I think is rather due to running on AFS... It seems to have gotten a lot more flaky the last few weeks. I suspect IT is trying to make it crappy in order to make people leave :)

Timing: In general we're in a feature freeze until #191 is done; after this we will make the BOINC release, and then we can start merging things again. But if you are planning to use Wire on BOINC, we can merge it before the release.

**Comment by kyrsjo** [comment]
Regarding the conflict, this is just that I moved the position of the `tools` section (astuce and dafor) of the build manual somewhere else in #186 .

**Comment by kyrsjo** [comment]
Thanks for lots of good tests and documentation! And thanks for having not-too-enormous DUMP canonicals - however the files `OPBEAM_DUMP.canonical` are all significantly shorter than the other dumps. Why?

File: SixTrack/sixtrack.s [hunk 1]
```diff
@@ -7393,269 +7437,302 @@ cc2008
...
+     &cos(atan(yv(1,j)/sqrt(((1d0+dpsv(j))**2-yv(1,j)**2)-              &
+     &yv(2,j)**2))-tx)                                                  
+          xi = xi*(cos(tx)-sin(tx)*tan(atan(yv(1,j)/                    &
+     &sqrt(((1d0+dpsv(j))**2-yv(1,j)**2)-yv(2,j)**2))-tx))              
+          yv(1,j) = sqrt((1d0+dpsv(j))**2-yv(2,j)**2)*                  &
+     &sin(atan(yv(1,j)/                                                 &
+     &sqrt(((1d0+dpsv(j))**2-yv(1,j)**2)-yv(2,j)**2))-tx)
+
+          xi = xi-(((yi*sin(ty))*yv(1,j))/                              &
+     &sqrt((1d0+dpsv(j))**2-yv(1,j)**2))/                               &
```
kyrsjo: FYI, you may loose some numerical precission by writing a^2-b^2, (a+b)*(a-b) is better.

SixTrack/sixtrack.s
```diff
@@ -18492,6 +18828,16 @@ c$$$         endif
          !Moved from daten()
          crabph4(ix)=el(ix)
          el(ix)=0d0
+!--Wire
```
kyrsjo: You want to always force ed/ek/el to be 0 for these elements in the tracking, regardless of what the input file says? It would probably be better to just check (if they should really always be 0 in fort.2), and spit out an error if the input is invalid.

mfittere: It already checks that the length is zero and puts an error message if not.

SixTrack/sixtrack.s
mfittere: That's a good point. I will change that.

mfittere: And yes, we definitely want to run on boinc! So I would like to get it into this version!

SixTrack/sixtrack.s
kyrsjo: OK, I was just a bit confused about what this code was doing in initialize_element.

SixTest/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ if(BIGNPART OR HUGENPART)
 endif()
 
 if(NOT COLLIMAT)
-	SET(SIXTRACK_TESTS bbe51 bbe52 bbe571ib0 bb bb_ntwin1 crabamp dipedge distance dynk_globalvars elensidealthin4d elensidealthin6d eric exact frs60 frs javier lostevery lost lostnumxv lostnotilt notilt prob1 prob3 s316 thick4 thick6dblocks tilt elensidealthck4d elensidealthck6d fma thick6ddynk ${SIXTRACK_TESTS})
+	SET(SIXTRACK_TESTS bbe51 bbe52 bbe571ib0 bb bb_ntwin1 crabamp dipedge distance dynk_globalvars elensidealthin4d elensidealthin6d eric exact frs60 frs javier lostevery lost lostnumxv lostnotilt notilt prob1 prob3 s316 thick4 thick6dblocks tilt elensidealthck4d elensidealthck6d fma thick6ddynk wire_thck4d wire_thck6d wire_thin4d wire_thin6d wire_thin6d_ibeco0${SIXTRACK_TESTS})
```
JamesMolson: Is the lack of space here:
```
wire_thin6d_ibeco0${SIXTRACK_TESTS}
```
between the 0 and the $ okay? Currently ${SIXTRACK_TESTS} starts as empty but it might not stay that way forever.

JamesMolson:
Ah, just the version number to bump now?

mfittere:
added it! Thanks for spotting!

- James, I didn't test your ctest again (my apologies).
- prob1/prob3 with cr never managed to run through on lxplus without me being logged out or any afs problems (just as a side note). How do you manage that it runs on lxplus?

JamesMolson:
I always ran in /tmp/ and screen so I avoided afs/kerberos entirely.

mfittere:
so where do I have to bump the version number? Just change it in sixtrack.s
      data version /'4.6.4'/
      data moddate /'06.02.2017'/

JamesMolson:
Yes :)

mfittere:
Done! I press the red green button?

mfittere:
I press the big green button now guys?

kyrsjo:
Awesome! :)

PR #175, opened 2017-01-25
Fix missing { typo in CMakeLists.txt

JamesMolson:


kyrsjo:
Strange that it didn't make CMake panic...

PR #176, opened 2017-01-25
Rename strtod function.

JamesMolson:
to dtoa_strtod.


kyrsjo:
Thanks for renaming `strtod` -> `strtod_cr` and `dtoa` -> `dtoa_cr`; now we avoid conflicts with system libraries.

PR #183, opened 2017-01-30
Build/test system updates

JamesMolson:
This fixes the test suite on freebsd (don't use clang), removes some now un-needed functions, and makes the BOINC build link correctly on OSX.


SixTrack/cmake_six
```diff
@@ -70,8 +70,8 @@ fi
 echo ""
 
 #Get options
-compilers=(defaultcompiler gfortran ifort ifort-icc ifort-icc-icpc nagfor) # Valid compilers. defaultcompiler = system default compilers
-compilers_CMAKE=('' '-DCMAKE_Fortran_COMPILER=gfortran' '-DCMAKE_Fortran_COMPILER=ifort' '-DCMAKE_Fortran_COMPILER=ifort -DCMAKE_C_COMPILER=icc' '-DCMAKE_Fortran_COMPILER=ifort -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc' '-DCMAKE_Fortran_COMPILER=nagfor')
+compilers=(defaultcompiler gfortran ifort ifort-icc ifort-icc-icpc nagfor gcc g++ gcc-g++ gcc-g++-gfortran clang clang++ clang-clang++) # Valid compilers. defaultcompiler = system default compilers
```
kyrsjo: Consistency: Normally the fortran compiler comes first, then c, then c++. I.e. it would be better to list it as `defaultcompiler gfortran gfortran-gcc-g++ gcc g++ gcc-g++ ifort ifort-icc ifort-icc-icpc nagfor clang clang++ clang-clang++`

This array also shows up in the "help" command.

kyrsjo:
The People are happy, and will merge the PR. Thanks! :)

PR #192, opened 2017-01-31
Stop if RIPP block encountered

### LINKED ISSUE #182: RIPP block NOT VALID
If RIPP block is present in `fort.3`, currently it just prints a warning and continues tracking. It should STOP and tell the user how to convert the fort.3 (there is a Python script).

kyrsjo:
This fixes #182


mfittere:
very good move! Should have done this in the first place ;-). We should remember for the future that if features are replaced by new ones, we should abort and post a message instead of just letting it run.

kyrsjo:
Agreed. It would have saved some grief about the mac version as well.

Once you accept the review, I can merge :)

PR #194, opened 2017-01-31
CTest SelfTest for ifort

kyrsjo:
Now the SixTest helper binaries (1) check that their inputs exists and (2) give the same error messages on gfortran and ifort.

Further, we check for the output in the right way, i.e. gfortran prints "STOP 1" at the last line, while ifort prints just "1" where 1 is the exit status.


kyrsjo:
@JamesMolson if you're happy with it, just go ahead and merge.

PR #190, opened 2017-01-31
Update Sixin.zip for prob1/prob3 to remove RIPP block, which was removed in v4.5.20

kyrsjo:

PR #198, opened 2017-02-01
Check if the BOINC input zip files for tests match the standard versions.

JamesMolson:
This should prevent #190 being an issue in the future. Currently must be manually run.


SixTest/CheckTestInputForBOINC.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+for i in $(ls -d */); do
+#echo ${i%%/};
+if [ -f ${i}Sixin.zip ]; then
+	mkdir ${i}/zip
+	cp ${i}/Sixin.zip ${i}/zip
+	cd ${i}/zip
+	unzip -qq Sixin.zip
+	for j in $(ls fort.*); do
```
kyrsjo: Sixin.zip may not only be fort.*!

kyrsjo: see for example `javier_bignpart` (no Sixin? huh.), `dynk_globalvars`, or `thick6ddynk`. They both have extra_inputs.txt.

kyrsjo:
Perfect!

PR #199, opened 2017-02-01
ctest on windows

JamesMolson:
Horrible and a work in progress but it works in some cases.

TODO:

1. Deal with \r\n line endings on windows when checking files (for extra checks).
2. CR runs + kills.


SixTest/bin/SixTestWrapper.cpp
```diff
@@ -56,13 +73,20 @@ struct KillInfo
 * 4: bool for STF enabled
 * 5: CR enabled
 * 6: CR kill time
```
kyrsjo: Document argument 7 & 8 ?

SixTest/bin/SixTestWrapper.cpp
```diff
@@ -91,25 +115,25 @@ int main(int argc, char* argv[])
...
 		STF = true;
 	}
 
-	if(atof(argv[5]) != 0)
+	if(atof(argv[7]) != 0)
 	{
 		CR = true;
-		KillTimes = ParseKillTimes(argv[6]);
+		KillTimes = ParseKillTimes(argv[8]);
 	}
```
kyrsjo: Extracting the executable names into a string here would be safer, in case the argv indices change.

kyrsjo:
This seems to work (tests passing); now we only need extra_checks and CR support :)

kyrsjo:
Pre-'\r' fix, I ran everything except prob1/prob3 on Win64, all pass except those who has extra_checks. Now running on Win32...

kyrsjo:
Slight problem: ctest on windows does NOT kill the currently running SixTrack process(s) when I try to kill it with control+c...

JamesMolson:
Oh, a '\r' check on them?

kyrsjo:
Yes, for canonicals used by check_extras. It could also be built into the script... But let's get this release done now.

kyrsjo:
Awesome with the check of the Zipin files. You may want to also consider moving the shell script to the `SixTest/bin` folder?

kyrsjo:
Also, I see it looks like you fixed the Windows CR issue. Can you push (cleanup if neccessary) the fix? Agreed that the timings look a a bit (but not totally) random...

kyrsjo:
Since you're adding stuff to the binary name: On Windows, could you strip out the ".exe" in the compiler name? Also, for the arch, are you sure that this makes sense? I just built the executable `SixTrack_462_crlibm_fast_tilt_cmake_Linux_gfortran_static_x86_64_32bit`...

PR #200, opened 2017-02-01
finishing solenoid code finally

vikasnt:


kyrsjo:
Thanks Vikas, and congrats on PR number 200 :) The PR looks perfect.

For new readers, please note that this is basically a continuation of issue #4 .

Now, you had some questions in the email you sent me, could you please post them here? It's infinitely easier to keep track of...

vikasnt:
​Hey,

Yes!! Where is my award money :D,
Anyway back to code

In this commit I have

1) replaced ejf0v(j))/ejfv(j)  by  oidpsv(j)
2) divided kx and kz by 2 at start of block, and then multiplied back at end of block.

Now, 2 things remains before finishing this code:


1)
From your last comment in Issue #4 ,

(After discussing with Riccardo on my end) These may be due to that the canonical momentum Px includes the vector potential of the field \vec A. So when you enter the solenoid, the field changes, meaning that the canonical momentum also changes. However, as discussed in section 3.5 in http://www.worldscientific.com/doi/pdf/10.1142/9781783262786_0003 [1], this dependence may cancel due to the effect of the fringe field. This should be discussed in the physics manual


But in WEPCH043(google gives first link to file) on 2nd page you can see
------------------------------------------------------------------------------
The implementation follows closely the formulae given in [8], where the canonical equations of motion are derived directly from the Hamiltonian. We only report the relevant formulae here and note some important facts:

The hard-edge model is used. Fringe fields of the solenoid are taken into account, however they are fi- nite ’hard-edge’ steps of the B-field;

the solution of the equations of motion are obtained from the expanded Hamiltonian, and it that sense the solutions or not ’exact’;

and although this is a thin element, both the normal- ized magnetic strength ks and its product ks · L with the length L have to be known and used in the equa- tions. This makes the solenoid different from e.g. the multipoles. 

-------------------------------------------------------------------------------
So the original implementation takes into account fringe effect. ie
There is no documentation for this code's validity/purpose at the start and end(reverse of this)of the block.
              yv(1,j)=yv(1,j)-xv(2,j)*strackx(i)
              yv(2,j)=yv(2,j)+xv(1,j)*strackx(i)
And as we have seen earlier this screws up final form of equation big time(we should ask the original author if we can, instead of wasting more time unless you have more insight).

2) x'(in code) vs px(in Sixtrack manual)--> I have done the calculation and as we had agreed earlier they don't match. SO we need to decide which is correct to use(x'=Px/P or px=Px/Po). Unless we know relation between px variable in reference papers and (Px,P,Po) used in Manual and Twiki, we can't be sure whether/which transformation is needed(i.e should the final code implementation be written in px=Px/P or px=Px/Po or simply px=Px, although currently code is in px=Px/P while manual uses px=Px/Po).Only one notation will be correct and whoever wrote the manual might help us out?
All the reference documents start with something like this
------------------------------ 
If zi, zf denote initial and final (canonical) coordinates, where z ∈ [x,px,y,py,σ,pσ],
--------------------------------


3) Lastly document the correct code in manual ,and add missing L in map of Solenoid with the 1/2 factor in strackx and strackz.
​

vikasnt:
Also, 

In Manual,

` Px,Py,Ps: canonical momenta coordinates in a straight or curved reference frame. Note that for a circular trajectory: vs = (1+hx)s ̇, Px = m x ̇+qAx, Py = m y ̇qAy and Ps = m s ̇(1 + hx)2 + q(1 + hx)As.
px = Px/P0, py = Px/P0, ps = Ps/P0 normalized momenta coordinates`.

And earlier at end of 2) I wrote from reference paper WEPCH043-

`If zi, zf denote initial and final (canonical) coordinates, where z ∈ [x,px,y,py,σ,pσ],`

so px in paper is canonical which should be same as Px in Manual. If this is correct, then the map with x'=px/P(in code) or px=Px/Po(in Manual) are both wrong and we should instead use Px(in Manual) for Solenoid map?

kyrsjo:
This does not compile? I get the error:
```
track.f:2556:65:

       crkve=yv(1,j)-(((xv(1,j)*strackx(i))*strackz(i))*oidpsv(j) !hr02
                                                                 1
Error: Expected a right parenthesis in expression at (1)
track.f:2557:65:

       cikve=yv(2,j)-(((xv(2,j)*strackx(i))*strackz(i))*oidpsv(j) !hr02
                                                                 1
Error: Expected a right parenthesis in expression at (1)
track.f:2558:64:

             yv(1,j)=crkve*cos_rn((strackz(i)*oidpsv(j))+        &!hr02
                                                                1
```
(etc.)

I will try to fix it...

kyrsjo:
OK, now compilation works. I would really want to sit down and check the equations a bit before merging this tough...

JamesMolson:
Being done by @tpersson in a new branch. Closing this one.

PR #209, opened 2017-02-07
ZIPF block (formerly ZIPIT)

kyrsjo:
This block makes it possible to compress the output data after a successfull SixTrack run.

Please review, and when done I'll bump the version number and recompile six.pdf.

Only issue I have is that I'm not sure how to make a ctest for this... Ideally I would like to uncompress Sixout.zip and then compare it. I tested it with the FMA test, and it seems to work great. Note that it requires LibArchive to work.


CMakeLists.txt
```diff
@@ -1074,7 +1078,7 @@ foreach(loop IN ITEMS ${FORS})
 	if(NOT BNLELENS)
 		ast_masks_deck(nwrtbnl ${loop})
 	endif()
-
+
```
JamesMolson: tab

SixTrack/sixtrack.s
```diff
@@ -1419,6 +1419,16 @@ C     Block with data/fields needed for checkpoint/restart of DYNK
 !
 !-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-
 !
++cd zipf
+      integer zipf_maxfiles,zipf_numfiles
+      parameter (zipf_maxfiles=64)
```
JamesMolson: Surely when say debugging and you want *everything* back 64 might be a little low?

kyrsjo: Hm, good point. I'll bump it up to 256!

kyrsjo:
OK the testing seems to work 100% now, at least on Linux. I still need to check it on Windows...

kyrsjo:
OK, so testing is DONE, works 100%. I even enhanced SixtestWrapper so that it checks that we get the expected ammount of extracted files.

Could you please take a look @JamesMolson ? If / when you're happy, let me know and I'll bump the version number and merge.

kyrsjo:
Ugh, please wait for merge. On 32-bit Linux I have a segfault in `libArchive_Fwrapper.c::f_write_archive_` line 95 which is the first free, i.e. `free(outname_c);`.

This also causes SixTestWrapper to hang while waiting for a backtrace that never arrives, UNLESS I run it in Valgrind.

kyrsjo:
OK, I think I fixed it; I wrote a `\0` to one position past the end of the `outname_c` array. On some systems i've seen position -1 been used to store the array size (later used for `delete[]`, which then crashes); I guess other systems use one-past the length of the array for some things some cases? Will retest in the 32-bit the VM tomorrow.

Meanwhile, any code scrutiny is very welcome! If I haven't heard anything, and the tests pass, I'll bump the version and merge tomorrow around lunchtime.

However, the behaviour of SixTestWrapper (hanging at 0 CPU) when there was a segfault in the C-part of SixTrack concerned me. Not that this is a bug in ZIPIT.

kyrsjo:
Yay, it worked!

kyrsjo:
I'm done, please review!

rdemaria:
Thanks! Give me a day for tomorrow...

PR #215, opened 2017-02-09
Dynk randu

kyrsjo:
Todo: add test


kyrsjo:
Adding the ONOFF function as well, for pulsed operation of e-lens.

To test (by @jfwagner ), by running about ~20 turns on (say) an e-lens with very large radius:
- Does it work as expected in the normal case, e.g. period of 2, or 5, or 10; and on-time of 1, or 3, or 9? Look at dynksets.dat
- Does it catch errors? Zero or negative on-time, period smaller than 2, on-time larger than or equal to period, negative on-time and/or period?
- What happens on negative time (use the shift in SET)? Is the behaviour described in the manual correct (always on for t <= 0)?

Also: Actually, I think allowing on-time to be zero or equal to the period could be useful for debugging simulations...

kyrsjo:
OK, I allowed p1=0 (always off) and p1=p2 (always on) - edge cases that can be useful for testing. Please check these too...

kyrsjo:
How is the testing going?

PR #218, opened 2017-02-09
Use /usr/bin/env bash in CheckTestInputForBOINC.sh

JamesMolson:


kyrsjo:
Also in `compare_stf.sh`, and lots of other scripts in the SixTest folder.

Not that it really fixed the problem of that test not really working on Windows (as far as I can remember).

PR #222, opened 2017-02-14
Disable CNG library on Windows, it doesn't work correctly on WinXP

kyrsjo:


kyrsjo:
Unfortunately, this seemed to have no effect whatsoever :/

kyrsjo:
It seems we're not the only ones who have problems with LibArchive's CMAKE linking to every library it can find, whether it is used or not:
https://github.com/libarchive/libarchive/issues/619

kyrsjo:
The user Boinc127 on the forum may have solved the problem:
https://lhcathome.cern.ch/lhcathome/forum_thread.php?id=4119&postid=28896#28896

Will remove the offending line from the CMakeLists and push a new exec later today!

Open source is great :)

kyrsjo:
Hm. This was one problem - problem number 2 was that we were linking the system libarchive provided by MSYS, not the one compiled by me. I tried to fix that, with partial success: It builds and links, but it doesn't run, failing because of some missing library:
```
 error while loading shared libraries: ?: cannot open shared object file: No such file or directory
```

The ldd output is as before:
```
$ ldd.exe SixTrack_cmakesix_32BIT_API_BOINC_BUILD_TESTING_CR_LIBARCHIVE_NO64BIT_defaultcompiler_defaultbuildtype/SixTrack_465_libarchive_cr_boinc_api_crlibm_fast_tilt_cmake_Windows_gfortran.exe_static_AMD64_32bit.exe
        ntdll.dll => /c/Windows/SYSTEM32/ntdll.dll (0x7ffea4ec0000)
        ??? => ??? (0x76ff0000)
        wow64.dll => /c/Windows/SYSTEM32/wow64.dll (0x76fa0000)
        wow64win.dll => /c/Windows/system32/wow64win.dll (0x76f30000)
        wow64cpu.dll => /c/Windows/system32/wow64cpu.dll (0x76f20000)
```

kyrsjo:
OK, now I realize: The problem is that libarchive_static.a (which I build) looks for `__imp_crc32` , `__imp_inflateEnd` etc. from zlib. These are defined in the `.dll.a` for zlib - i.e. the trampoline to link DYNAMICALLY to zlib -- so when I try to link STATICALLY to zlib, it doesn't find the symbols (because libz.a defines the symbols `crc32`, `inflateEnd` etc. *without* the `__imp_`).

Somehow this still works when compiling `libArchiveTester` (which is c++), but for compiling SixTrack it only links if I explicitly point it to the .dll.a. In both cases the result is a dynamic link to `zlib1.dll`.

Now, when I linked the system libarchive, I didn't have this problem (instead, I it linked Bcrypt...). So it seems that the problem isn't really in how I compile SixTrack, but libarchive, where the static version wants to link the dynamic version of zlib, which gfortran does not expect...

kyrsjo:
Seems we're not the first to stumble accross this bug:
https://github.com/libarchive/libarchive/issues/331

The files linking to problematic things are `archive_read_support_format_zip.c` and `archive_write_set_format_zip.c`, functions `__imp_inflateEnd`, `__imp_crc32`, `__imp_inflateReset`, `__imp_inflate`, `__imp_inflateInit2_`, `__imp_deflate`, `__imp_deflateEnd`, and `__imp_deflateInit2`.

PR #226, opened 2017-02-24
Fix collimat arrays

kyrsjo:
Fixing collimation array bugs


kyrsjo:
OK now? I would want to bump the version number for this.

Also, some response from the collimation team would be nice...

JamesMolson:
looks good to me

@rbruce1 @amereghe  ?

PR #229, opened 2017-03-16
Building/testing on aarch64

JamesMolson:
By default gcc seems to create fma instructions on armv8/aarch64. This disables them. Without this the tests will fail.


kyrsjo:
Tested it on LxPlus with GCC 4.4.7, it works now. Merge at will!

PR #233, opened 2017-03-22
Cleaning up write statements

kyrsjo:
Replacing the
```
+if cr
       write(lout,*)
+ei
+if .not.cr
       write(*,*)
+ei
with simply
```
       write(lout,*)
```
where the value of `lout` is varying.


kyrsjo:
That should be it? I deleted about 8000 LOC... Now testing that the output is still the same!

kyrsjo:
Checked normal version, CR version, collimation version, and DA version, comparing STDOUT from each version before and after this patch. This shows that the outputs are identical (except some extra spaces that were removed in COLLIMAT, and some outdated output from daten when parsing DUMP blocks).

Will bump the date and merge now.

PR #236, opened 2017-03-28
SCATTER block

kyrsjo:
Framework for scattering routines (do not merge yet).


kyrsjo:
Regarding the fact that HUGENPART currently does not compile in this branch: It seems we've broken the limit of how much RAM we can use in common blocks; there is a limit at 2GB.

Please see:
https://en.wikipedia.org/wiki/.bss#BSS_in_Fortran
(I wrote that section on Wikipedia after the last time I had to learn more than I really wanted about how Fortran manages it's memory).

Please also see issue #250 . Note that CMake is smart enough to read the Fortran sources to look for modules, and build the files containing modules before the code that depends on the modules. I therefore propose to put the code in scatter.f90 into an actual module, including the global COMMON blocks, which should not be common blocs but just allocatable module variables... The same should probably also be done with DYNK... But that's a separate PR, to be done after this (in order to profit from the improvements in the build system made in this branch).

SixTrack/scatter.s90
```diff
@@ -181,6 +182,8 @@ end function
 subroutine scatter_profile_getParticle(profileIdx, x,y, xp,yp,E)
 
     implicit none
+
+    ! Input/Output
```
kyrsjo: I would rather call that arguments, or to be technically correct, dummy variables.

I/O implies something being sent out of the "computer", i.e. to disk, terminal, network, etc. ...

SixTrack/scatter.s90
```diff
@@ -199,8 +202,9 @@ double precision function scatter_generator_getCrossSection(profileIDX, generato
 +ca stringzerotrim
 +ca comscatter
 
+    ! Input/Output
```
kyrsjo: ditto.

SixTrack/scatter.s90
```diff
@@ -239,6 +243,7 @@ subroutine scatter_generator_getTandXi(generatorIDX, t, xi)
     
     double precision scatter_generator_getPPElastic
 
+    ! Input/Output
```
kyrsjo: ditto

kyrsjo:
TOADD (probably after merging this):
 * Possibility of loading pre-calculated distributions of e.g. t, xi from file as another "generator"
 * Possibility to enable SCATTER only for one given turn

kyrsjo:
Getting closer to done; TODO:
 * GENerator ppbeamelastic should accept cross section as optional argument in the input file -> Veronica
 * Write documentation -> Veronica
 * CRlibm -> Kyrre
 * CR -> Kyrre
 * Fortran module, dynamic allocation -> Kyrre
 * SixTest (after CRlibm)

vkbo:
ToDos:

- [x] GENerator ppbeamelastic should accept cross section as optional argument in the input file -> Veronica
- [ ] Write documentation -> Veronica

kyrsjo:
Merged!

I think you can just check off stuff in my comment above yours?

SixTrack/scatter.s90
```diff
@@ -400,16 +400,17 @@ subroutine scatter_parseGenerator(getfields_fields,getfields_lfields,getfields_n
     case("PPBEAMELASTIC")
 
         scatter_GENERATOR(scatter_nGENERATOR,2) = 10 ! Code for PPBEAMELASTIC
-        if(getfields_nfields .ne. 8) then
-            write(lout,'(a)') "SCATTER> ERROR, GEN PPBEAMELASTIC takes 8 arguments:"
-            write(lout,'(a)') "GEN name PPBEAMELASTIC a b1 b2 phi tmin"
+        if(getfields_nfields .lt. 8) then
```
kyrsjo: Please make it scream if it is .gt. 9 as well...

SixTrack/scatter.s90
```diff
@@ -435,6 +436,13 @@ subroutine scatter_parseGenerator(getfields_fields,getfields_lfields,getfields_n
         ! PPBEAMELASTIC b2
         scatter_fdata(tmpIdx+4) = round_near(errno,getfields_lfields(8)+1,getfields_fields(8))
         if(errno .ne. 0) call rounderr(errno,getfields_fields,8,scatter_fdata(tmpIdx+4))
+
+        if(getfields_nfields .eq. 9) then
+            ! PPBEAMELASTIC corss-section
```
kyrsjo: Spelling

SixTrack/scatter.s90
```diff
@@ -400,16 +400,17 @@ subroutine scatter_parseGenerator(getfields_fields,getfields_lfields,getfields_n
...
+            write(lout,'(a)') "GEN name PPBEAMELASTIC a b1 b2 phi tmin (cross-section)"
             call prror(-1)
         end if
 
         ! Request space to store the arguments
         tmpIdx = scatter_nfdata + 1
         scatter_GENERATOR(scatter_nGENERATOR,3) = tmpIdx
-        scatter_nfdata = scatter_nfdata + 5
+        scatter_GENERATOR(scatter_nGENERATOR,4) = 0      ! Index of scatter input, if present
+        scatter_nfdata = scatter_nfdata + getfields_nfields - 3
```
kyrsjo: Add comment that getfields_lfields is 8 or 9

SixTrack/scatter.s90
vkbo: It's a typo in comment. It makes no difference whatsoever. Just fix it when you see it.

kyrsjo:
OK, HUGENPART should now work correctly!

kyrsjo:
CRLIBM done; @vkbo could you please check that I didn't miss anything? You'll get the idea from the diff...

For CR, I think we should add the test first, then add the CR code, and finally check that we can reliably get it to work.

kyrsjo:
Only CR left now. @vkbo , could you please take a look at the test, that it makes sense? I extended it to 1000 turns and did some more stuff with DYNK.

PR #238, opened 2017-03-29
FCC test

kyrsjo:
Adding a FCC lattice test to SixTest; Requires big- or hugenblz. Also some restructuring of SixTest/CMakeLists.txt.


JamesMolson:
This is enabled all the time but should only be on with bignblz or above for now.

kyrsjo:
Thanks! I don't know how I missed that - I restructured the CMake to work like that, but then I forgot...

PR #239, opened 2017-03-31
astuce++

JamesMolson:
astuce re-written in C++

It can now take multiple input .s files.

It can build sixtrack and all tests pass, but the generated .f files are not exactly identical.


JamesMolson:
This also removes limitations in the sizes of everything, decks, includes, line lengths, deck names etc etc etc, everything is now unlimited.

kyrsjo:
I have checked and confirm that except for (1) the old sixtrack adding trailing whitespace in empty lines and removing them after non-empty lines (which may cause bugs!), and the order of blocks, the produced source files are identical. This was checked using `SixTrack/beauty/splitFortranFile.py` .

Functionality is therefore now CONFIRMED OK.

The tests were done on all fortran files generated by the standard build, and the beamgas.f90 generated by the BEAMGAS (COLLIMAT -CRLIBM) flag.

SixTrack/tools/astuce/astuce++.h
```diff
@@ -0,0 +1,118 @@
...
+	void PrintInputSettings();
+
+	//Read the input .s files into a buffer
+	void ReadInputFiles();
+
+	//Does the first parse
+	void ParseSourceFile();
+
+	//Extracts +cd flags
+	void ExtractFlags();
```
kyrsjo: This fills CallStorage. Please rename to "ExtractCallsBlocks" or something like that, rather than flags which is really something else.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+	else
+	{
+		std::cerr << "Could not open " << InputFileName << std::endl;
+		exit(EXIT_FAILURE);
+	}
+
+	char c_buffer[4096];
+	while(infile->good())
+	{
+		infile->getline(c_buffer, 4096);
```
kyrsjo: Check if bufffer was actually big enough. There is a flag.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+					size_t old_step = 0;
+					while(step!=std::string::npos)
+					{
+						step = defines.find(",", step+1);
+						Flags.insert(FixCase(defines.substr(old_step,step-old_step)));
+						old_step = step+1;
+					}
+				}
+			}
+			else if(itr->substr(0,2) == "E " || itr->substr(0,2) == "e ")
```
kyrsjo: Here you look for "E ", higher up you look for "EX" and "DF". Can there be lines "DF" without a space before the actual flags?

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+		{
+			*LogStream << it->first << " ";
+			it++;
+		}
+		*LogStream << std::endl;
+		*LogStream << "Total \"calls\": " << CallStorage.size() << "\n" << std::endl;
+	}
+}
+
+void Astuce::ExtractDecks()
```
kyrsjo: This code is *really* similar to ExtractFlags()...

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+void Astuce::DoExpandIF(std::map<std::string, std::list<LineStorage> >::iterator Storage_itr)
+{
+	std::string BlockName = Storage_itr->first;
+
+	if(OutputFlag)
+	{
+		*LogStream << "Processing: " << BlockName << " with " << Storage_itr->second.size() << " lines." << std::endl;
+	}
+
+	EnabledStates.push(true);
```
kyrsjo: A comment about how this is used could be useful here.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+	//If a parent block is disabled, all children should also be disabled.
+	//Done with a std::stack<bool> and push()/pop()/top()
+
+	//Does this block have a +if statement?
+	if(line->Text.substr(0,3) == "+if")
+	{
+		//Disable this line
+		line->Enabled = false;
+
+		bool EnabledState = false;
```
kyrsjo: Very close name to to the member "EnabledStates"; consider renaming.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+			FlagNames.push_back(SplitSubString);
+		}
+
+		for(size_t n=0; n < FlagNames.size(); n++)
+		{
+			bool NOT = false;
+			bool AND = false;
+			bool OR = false;
+
+			if(FlagNames.at(n).substr(0,5) == ".AND.")
```
kyrsjo: Valid pre-C++11? On the other hand, if GCC4.4.7 is happy with it...

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+		//Find the name of this statement
+		size_t NameStart = line->Text.find_first_not_of(" ", 4);
+		size_t NameEnd = line->Text.find_first_of(" ", NameStart+1);
+		std::string FlagName = FixCase(line->Text.substr(NameStart,NameEnd-NameStart));
+
+		size_t Position = 0;
+		size_t NextOR = 0;
+		size_t NextAND = 0;
+		std::vector<size_t> Splits;
+		std::vector<std::string> FlagNames;
```
kyrsjo: Also close names, however here both are local variables -> probably OK.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+			bool NOT = false;
+			bool AND = false;
+			bool OR = false;
+
+			if(FlagNames.at(n).substr(0,5) == ".AND.")
+			{
+				AND = true;
+				FlagNames.at(n).erase(0,5);
+			}
+			if(FlagNames.at(n).substr(0,4) == ".OR.")
```
kyrsjo: .AND..OR. can't be valid ASTUCE -> this could probably be an `else if`. The "if .not." is however OK , since .OR..NOT. and .AND..NOT. is valid astuce.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+		while(Lines_itr != Storage_itr->second.end())
+		{
+			*LogStream << BlockName << "\t" << Lines_itr->Enabled << " " << Lines_itr->Text << std::endl;
+			Lines_itr++;
+		}
+
+	}
+}
+
+void Astuce::ProcessIfBlock(std::list<LineStorage>::iterator line)
```
kyrsjo: Nice :)

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+	EnabledStates.push(true);
+
+	std::list<LineStorage>::iterator Lines_itr = Storage_itr->second.begin();
+	while(Lines_itr != Storage_itr->second.end())
+	{
+		ProcessIfBlock(Lines_itr);
+		Lines_itr++;
+	}
+
+	EnabledStates.pop();
```
kyrsjo: If everything is OK, this *should* leave the stack exactly empty right?

It might be a good idea to check that this is the case, in case of missing `+if`. I guess if you have extra `+if`, the pop will fail and you'll have an exception, but one should also check the case that the stack is too long.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+
+		//AND (&) the result of the above with the current stack top
+		EnabledStates.push(EnabledState & EnabledStates.top());
+	}
+	else if(line->Text.substr(0,3) == "+ei")
+	{
+		//Disable this line
+		line->Enabled = false;
+
+		EnabledStates.pop();
```
kyrsjo: Here also check for empty stack and maybe print an error...

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+
+void Astuce::ExpandCA()
+{
+	bool Replaced = true;
+	while(Replaced)
+	{
+		//Repeat until no +ca exists
+		Replaced = false;
+
+		//First step, go through existing +cd blocks and check for +ca
```
kyrsjo: This would be a description of the whole while loop? Replacement inside decks is below.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+				CallStorage_itr = CallStorage.find(FlagName);
+				if(CallStorage_itr != CallStorage.end())
+				{
+					//Delete the +ca statement line
+					Lines_itr = Storage_itr->second.erase(Lines_itr);
+
+					LineStorage Title;
+					Title.Enabled=true;
+					Title.Text="!!! START " + FlagName;
+///					Storage_itr->second.insert(Lines_itr,Title);
```
kyrsjo: Comment why not putting these in (makes FOX explode)

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+
+		DeckStorage_itr = DeckStorage.find(*Decks_itr);
+		if(DeckStorage_itr != DeckStorage.end())
+		{
+			std::list<LineStorage>::iterator Lines_itr = DeckStorage_itr->second.begin();
+			while(Lines_itr != DeckStorage_itr->second.end())
+			{
+				if(Lines_itr->Enabled == true)
+				{
+					OutputFileBufferStream << Lines_itr->Text.substr(0,80) << "\n";
```
kyrsjo: Put a comment that you are cropping to 80 chars.

Eventually, disabling this could be a keyword in the .ast file...

kyrsjo:
end subroutine pedantery :) Overall good, my comments are on readability of the sources.

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -0,0 +1,708 @@
...
+		DeckStorage_itr = DeckStorage.find(*Decks_itr);
+		if(DeckStorage_itr != DeckStorage.end())
+		{
+			std::list<LineStorage>::iterator Lines_itr = DeckStorage_itr->second.begin();
+			while(Lines_itr != DeckStorage_itr->second.end())
+			{
+				if(Lines_itr->Enabled == true)
+				{
+//					OutputFileBufferStream << Lines_itr->Text.substr(0,80) << "\n";
+					OutputFileBufferStream << Lines_itr->Text << "\n";
```
kyrsjo: Was it not this that made FOX explode?

JamesMolson: No, that was:

```
Title.Text="!!! END " + CallName;
Storage_itr->second.insert(Lines_itr,Title);
```

Basically a FOX block getting a comment inside it.

kyrsjo: Ah, no, that was the headers. But you removed those.

SixTrack/tools/astuce/astuce++.h
```diff
@@ -0,0 +1,118 @@
...
+	void PrintInputSettings();
+
+	//Read the input .s files into a buffer
+	void ReadInputFiles();
+
+	//Does the first parse
+	void ParseSourceFile();
+
+	//Extracts +cd flags
+	void ExtractCalls();
```
kyrsjo: Thanks for renaming :)

SixTrack/tools/astuce/astuce++.cpp
JamesMolson: Could add the space to the search I suppose. for "EX " and "DF "

SixTrack/tools/astuce/astuce++.cpp
JamesMolson: Yes, the difference being where the storage used and the search term.

SixTrack/tools/astuce/astuce++.cpp
JamesMolson: Fixed

SixTrack/tools/astuce/astuce++.cpp
JamesMolson: Yes, lxplus g++ is happy with it.

kyrsjo: OK, fine.

SixTrack/tools/astuce/astuce++.cpp
kyrsjo: Would merging these codes into one be easy? I.e. pass the function the search term and a pointer to the storage? If it's a hassle then don't bother.

SixTrack/tools/astuce/astuce++.cpp
kyrsjo: Great :) I saw you did it.

kyrsjo:
Thanks for the changes. All good now except for one thing: What happens if the +if and +ei doesn't match up?

kyrsjo:
Uhm, this does not build?
```
14:13:37 kyrsjo@pcbe15843 ast_mask $ git status
HEAD detached at jamesMolson/cmake
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	../../../Doc/building_sixtrack/building_sixtrack.pdf
	../../../SixTest/scatter_bbelastic/

nothing added to commit but untracked files present (use "git add" to track)
14:13:44 kyrsjo@pcbe15843 ast_mask $ cat track.ast
sixtrack.s
trackn.f
df vvector,crlibm,tilt,fast
e tra_thck
e tra_thin
e dynktrack
ex
14:13:50 kyrsjo@pcbe15843 ast_mask $ cat track.ast.log 
Opened /home/kyrsjo/code/SixTrack/SixTrack/SixTrack_cmakesix_defaultcompiler_defaultbuildtype/ast_mask/track.ast
Bad entry in input: ex
Exiting
```

SixTrack/tools/astuce/astuce++.cpp
```diff
@@ -405,6 +406,13 @@ void Astuce::DoExpandIF(std::map<std::string, std::list<LineStorage> >::iterator
 
 	EnabledStates.pop();
 
+	//Check we are now empty
```
kyrsjo: What happens if you pop once too much, i.e. if there is an extra +ei there?

kyrsjo:
Two problems to be fixed:
- If a "+cd" is missing (I just added a space in front of the line, I tried with `rhicelens`), the error message points to the name of a deck (`NWRTBNL` in this case)
- If a "+dk" is missing (I just added a space in front of the line, I tried with `cor_ord`), there is no error message.

JamesMolson:
Should all be fixed now?

kyrsjo:
It now breaks when building collimation (-CRLIBM to cmake_six), because the plato_seq.ast file looks like this:
```
plato_seq.s
plato_seqn.f
df
e plato_seq
```

The error message is:
```
Opened /home/kyrsjo/code/SixTrack/SixTrack/SixTrack_cmakesix_NOCRLIBM_defaultcompiler_defaultbuildtype/ast_mask/plato_seq.ast
Bad entry in input: df
Exiting
```

Requests:
- Allow empty df
- Always print error message to stderr, it will be printed by the makefile
- Always print the name of the crashing ast file

JamesMolson:
okay, try and break it now

kyrsjo:
Challenge accepted!

kyrsjo:
Tried to build a bunch of stuff. With the exception of DATAMODS (which was my own stupid mistake), it works.

Now trying with corrupted input files...

kyrsjo:
I found more ways to break it!

The code looking for `+ca blocks` (I played with "+cd version" if this matters):
- If you put something in front of it - any character, including a `!` (i.e. a comment), a `+if`, spaces, random characters: It still reads the ca block and builds the fortran file "correctly". It really needs to be more picky...
- If you put something after it, anything, even a comment, or a space, it takes this as part of the name. Comments and spaces are NEVER part of the name, and should be stripped out.

The '+if' parser is pretty robust, but it breaks when you put a comment straight after it without any space. It seems that the comment becomes part of the name?

Decks on the first line works. Good. But I would assume that the decks have the same problems as the call blocks?

If there are two decks with the same name, only the first is taken, the second is discarded. The same is true for call blocks (+cd).

JamesMolson:
When we are looking for +ca, at this stage the code has been split by lines and the search is:

```
if(Lines_itr->Text.substr(0,3) == "+ca")
```

So I don't quite see how adding a space or anything else before that is going to trigger that code, hmm.


I'll add a search for ! to the terminator for names, instead of just a space.
There are many forms of comments though...

I can put in a check at the map insert too.

kyrsjo:
Regarding the +ca, I wonder if the fact that I was playing with the very first line of code may have changed things?

Regarding comments, the only one that matters would be the free-form style which is triggered by exclamation mark. The old style ignores the whole line by putting something in the first column, which will anyway make it a non-astuce line since astuce commands have to start with a + in column 1. Also + in column 1 would not be a valid FORTRAN comment AFAIK.

kyrsjo:
Getting there, all the usual builds now work, but I can still break it... I'm building with `./cmake_six -CRLIBM debug` as this is pretty fast...

If I define a two blocks with the same name like this:
```
+cd version
      character*8 version
      character*10 moddate
      integer itot,ttot
      dta version /'4.6.14'/
      data moddate /'04.04.2017'/
+cd version!asdf
      character*8 version
      character*10 moddate
      integer itot,ttot
      data version /'4.6.14'/
      daa moddate /'04.04.2017'/
```
(Fortran syntax errors are deliberately inserted in order to produce a recognizable error), I get two different call blocks out according to the astuce++ log file (useful log file!):
UMLALID1 VERSION VERSION!ASDF WIRE

Also if I define a block `version` and another block `version ` (with a space), I get the first one and no warnings, as they are considered to be different. If I only have `version `, It fails because it can't find `version` (no space).

If I then try to call `version ` and define `version `, it also fails with the following message:
```
Searched for VERSION and it was not defined anywhere!
The following +cd blocks exist:
"" "ACDIP1" "ACDIPKICK" "AL
```
If I define `crcoall !asdf`, it finds the block `!asdf` but not `crcoall`, and compilation fails.
Defining `version` and calling `version ` works as expected. Calling with `+ca version !adf` (space then comment) also works as expected, as does `+ca version!adf` (comment without space).

It thus looks like the calls are correctly stripping the strings of any trailing spaces, while the `+cd`s are not. Also, in the the error message `Searched for VERSION and it was not defined anywhere!` the name `VERSION` should be in `"VERSION"`, but it is not. 

Further, if I put a space in front of ` +cd version`, it still finds it; it should not. Ditto with the comment symbol `! +cd version`. A +cd command should always be leftmost on the line.
Note that this fault is specific to the first +cd of the file; doing the same to `+cd crcoall` gives the expected behaviour.

If you could do the same for +cd as you do for +ca, that would be great. Alternatively, protest and exit if there are unexpected stuff on the input line.

The `+if` almost works as expected, I was only able to break it in one way:
Putting the statement directly after the +if, as in `+if.not.cr`. This always behaves as `if false`, and disables the code inside the +if. The state of the `cr` flag does not matter. I tried it with this block:
```
+cd crcoall
!     Standard output unit
!     For CR version, this is the "buffer file" fort.92;
!     Otherwise write directly to "*" aka iso_fortran_env::output_unit (usually unit 6)
+ifcr
      integer lout
      common /crflags/lout
+ei!adf
```

No problems found with `+ei` 👍 

Multiple identical +cd definitions were caught 👍 

Multiple identical +dk definitions were caught  👍  (as long as now comments or spaces on the same line)

Two decks with a spurious space after resulted in a multiple definition of deck `""` warning.

Spurious space after a deck command in the ast (`e beamgasK `) was put into the name it looked for.

No space between deck command and deck name `ebeamgasK` gave an error message 👍 

Spurious space after file name `sixtrack.s` was ignored 👍  However it was not ignored after `beamgasn.f `...

Extra comma at the end of the flag list  seems to be ignored 👍 

Extra newlines at the end of the file ignored 👍  but no newline(s) after `ex` are strictly neccessary 👍

kyrsjo:
Good.

Problems still remaining:
- `+if .not. FLAG` always evaluates to true
- Ignoring any characters in front of the first +cd
- Space *in front* of the .s file name in the ast mask results in the wrong error message (can't find a deck in the files) instead of "cant find the input file". Similarly, putting a space in front of the ".f" file produces a file " track.f", but that's OK.

I'll try to fix the remaining problems then give you a PR...

kyrsjo:
Hi James, this PR should fix the remaining issues with astuce++:
https://github.com/JamesMolson/SixTrack/pull/7

JamesMolson:
Merged! All finished now?

kyrsjo:
Yes! Bump the last digit of the version number & date in sixtrack.s and MERGE!

... And please put a meaningfull commit comment on the merge (more than just "merged PR 239 branch cmake"), so that it is easy to recognize the PR when looking at the branch plot... Put the version number and a quick comment.

JamesMolson:
I'll let you click merge :)

PR #245, opened 2017-04-07
Collimation cleanup and rework and file splitting

JamesMolson:
So people can see the current state - do not merge yet!


JamesMolson:
Ok, so I have pulled out a lot of the collimation code from trauthin and thin6d into subroutines.

These can then be removed and put into their own file. From my testing it generates the same result as before, but I've only tested a standard loss map case.

@kyrsjo @amereghe @rbruce1  - Thoughts so far? It is mostly copy/paste + some cleaning + common block changes, but now a file split will be trivial.

kyrsjo:
I think you mentioned that now only the Geant4 branch counts (which also contains the modifications in this branch), which is tracked in #247 . Can we close this PR?

JamesMolson:
Yes

PR #246, opened 2017-04-13
New EXP version of BEAM block

kyrsjo:


rdemaria:
This is comment from 24/10/2016 for a possible improvement of the equations.

The boost then rotates the weak and cancel px assuming half crossing angle xang. So if in the Sixtrack frame before entering in the beam-beam one has x=0, z=a, px=px_closedorbit, in the boosted frame one has x=a*tan(xang)*cos(xang) and px=0. Same happen to the strong but this time x=a*sin(xstr) although I think it should be x=a*tan(xstr)*cos(xang). (I hope signs are correct). After the beam-beam kicks, the closed orbit is added again.

kyrsjo:
This now passes all fast current tests. Medium tests passed at least before setting ed,ek = zero in initialize_element.

rdemaria:
What do you mean by the last sentence?

On Thu, Apr 13, 2017 at 4:20 PM, Kyrre Ness Sjøbæk <notifications@github.com

kyrsjo:
TODO: If the extra single-element parameters of the are set, treat this as an error (ala WIRE).

JamesMolson:
Does dup2() work on windows, bsd, etc?

kyrsjo:
While I remember: Let's make sure that if the order of the elements in the BEAM block matters, this is actually checked when reading. Both for the new- and the old version.

kyrsjo:
James: It's in the "if not windows" thingy. Windows will still get it the old way (so functionality isn't 100% equivalent).

BSD: I cannot check, but it looks like it? https://www.freebsd.org/cgi/man.cgi?query=dup2&apropos=0&sektion=0&manpath=FreeBSD+4.3-RELEASE&format=html

If not, disabling it for that OS should be easy. I will anyway test this branch on Mac etc. + try various branches before merging, as the changes are kind of extensive...

kyrsjo:
Before merging: Increase the number of turns in the test & check that they still match.

kyrsjo:
TODO: In initialize_element, wrap the BEAM code that's there now in an if(lfirst), at least if the expert flag is on...

kyrsjo:
TODO: I don't think we have any tests with BEAM and ibbc = 1 (aka `Linear coupling considered in 4D and 6D (0 = off; 1 = on).`). This is quite a different code path in some places, so we should test it before- and after this change, for the "traditional" block.

kyrsjo:
TODO: The fort.13 input for `beam-HO_6D-newstyle-loadParticles` gives incorrect results because fort.13 is written (as fort.12) then read WITHOUT using CRLIBM. This causes it to be inaccurate in round-trip. Either remove this test, or fix the file (which will probably mean making a reader and writer(s) using CRLIBM).

kyrsjo:
OK, so in *priniciple*, it all works now. Some polishing is still remaining, however that should be much easier. I will do that polishing, either tonight or tomorrow morning.

Riccardo: If you could start reviewing, especially the manual, that would be really great! For the manual, please feel free to either commit to my branch or make an offshot and give me a PR to this branch. Please don't commit the PDF file... For the code, if you could just add annotations on GitHub, that would be great; I will be editing it a little bit still.

James: If you could check that I did not break SixTestWrapper on BSD, that would be awesome. If I did, a PR to fix it (even if that is just an ifdef to skip that code) would be very welcome! For the record, it does work correctly on Linux and Mac.

TODO (Kyrre) - mostly general code polishing:
- [ ] If in the new type of BEAM block, CHECK that ed/ek/el etc. are all set to zero
- [ ] Make the input file generator formatting a bit nicer...
- [ ] Fix or remove test beam-HO_6D-newstyle-loadParticles
- [ ] In initialize_element, wrap the BEAM code that's there now in an if(lfirst), at least if the expert flag is on...
- [ ] Increase the number of turns in the test & check that they still match. Commit new (longer) canonicals.
- [ ] Add test with BEAM and `ibbc = 1`. Run on both the current MASTER and the new branch, and check that the result is the same. Not a physics test, but at least we know that we didn't screw up anything in the old code by introducing BEAM-EXP.
- [ ] Run all the tests
- [ ] Run most tests on most platforms
- [ ] Bump version in sixtrack.s and six.tex, commit six.pdf, and merge to master!

For the future, we can think about fixing the physics as mentioned by Riccardo in https://github.com/SixTrack/SixTrack/pull/246#issuecomment-293836906 .

JamesMolson:
```
SixTestWrapper.cpp:375:55: error: use of undeclared identifier 'S_IRUSR'
                        int fd_6 = open("fort.6",O_WRONLY|O_CREAT|O_TRUNC, S_IRUSR | S_IWUSR);
```

kyrsjo:
James: Thanks. That's on FreeBSD, the mode setting is crashing? Seems I need to include sys/stat.h to get those DEFINEs. Could you please try again now?

Relevant docs:
https://www.freebsd.org/cgi/man.cgi?sektion=2&query=open
https://www.freebsd.org/cgi/man.cgi?query=chmod&sektion=2&apropos=0&manpath=FreeBSD+11.0-RELEASE+and+Ports

kyrsjo:
I checked and digged a bit into the code: It turns out that the order of beam elements in the BEAM block does not matter, since the data are stored in parbe() and ptnfac(), which are indexed by the single element index. Later (in umlalid1), the array imbb(i) is used to index from the *structure* element index to the BEAM parameter table (of length nbb), i.e. bbcu etc.

So there is no need to check for that the ordering is correct in the input file - it doesn't matter :)

kyrsjo:
OK, I'm done; please check that things are sane; feel free to give me pull requests (to this branch) for changes or request them in the diff view.

Once you are happy, let me know and I will bump the version and merge, then close couple of issues...

PR #247, opened 2017-04-14
Geant4 collimation

JamesMolson:
Base is #245 

Allows passing particles to geant4 in collimators in order to use the physics within geant4.

In progress...


kyrsjo:
Thanks!

FYI it's a good idea to make note of the version (now 3.7.0) in the commit message for the version bump itself and especially for the merge commit! Makes things a LOT easier...

kyrsjo:
Also to put a short "change log" in the merge commit; this makes the tags easier to write... We should REALLY update the tags tough, they are lagging far behind.

PR #258, opened 2017-04-22
Improving Doc makefiles

amereghe:
In particular, improving the one in physics_manual and creating a couple of new ones for building_sixtrack and user_manual

...very very dirty ones: nothing subtle, simply something for cleaning tex files and automatically compile...

### DIFF
--- a/Doc/building_sixtrack/makefile
+++ b/Doc/building_sixtrack/makefile
@@ -1,0 +1,11 @@
+
+manName=building_sixtrack
+
+all: $(manName).pdf
+
+clean:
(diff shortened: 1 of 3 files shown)


kyrsjo:
Thanks. Are there any reason why clean should not remove the .pdf? I find that to be expected behaviour, i.e. clean all generated files...

rdemaria:
I see one wants to keep the pdf only and remove all the temporary latext
files. One can add a cleanall target that removes also the pdf. What do you
think?

amereghe:
I put the cleanall in the physics manual, not in the build / user manual - can do that and re-pull

PR #259, opened 2017-04-25
Small fixes to manual (multipoles) etc.

kyrsjo:


kyrsjo:
@androula  Could you look over what I did to the manual? It's basically just formatting... Also, `yp` is really `y'` right?

PR #270, opened 2017-05-18
Fix beambeam nosync

### LINKED ISSUE #269: Beam-beam new style breaks with no synchrotron motion
When ition in the SYNC block is equal to 0, SixTrack breaks with the error:

         ++++++++++++++++++++++++
         +++++ERROR DETECTED+++++
         ++++++++++++++++++++++++
         RUN TERMINATED ABNORMALLY !!!

         BEAM_BEAM: EITHER NORMALIZED EMITTANCES OR THE RESULTING SIGMA VALUES EQUAL TO ZERO

This does not happen with the equivalent old style input.

rdemaria:
Fixes #269


rdemaria:
Test pass by taken the sigma matrix calculated by old-style and ition=0 necessarily different from the one with ition=1

PR #271, opened 2017-05-24
Include ZLIB in the distribution

kyrsjo:
This patch explicitly includes zlib in the distribution. This fixes the inability to compile with LIBARCHIVE on LxPlus.

Note: While this does build on my workstation and on LxPlus, I have not yet checked it on Windows, and expect that I have to do some adjustments there...

@amereghe This should solve your issue!


rdemaria:
Do we wait for windows tests to merge?

kyrsjo:
Yes. I suspect windows compilation is broken... I'll try it on Friday!

PR #261, opened 2017-05-03
Fix beambeam boost

rdemaria:
Fixes #260

PR #282, opened 2017-06-06
Fix canonical output for the javier_bignpart test.

JamesMolson:


kyrsjo:
We know; but the file is rather big (90 MB?) and GitHub as a "soft" quota + this would go into history and everyone will have to download it for all eternity. We are anyway reviewing a lot of the bb physics, which will probably rustle these tests quite a bit, so let's wait...

kyrsjo:
Also, we could probably reduce the number of turns in this test; the current 10k is a bit much (and takes 4 minutes to run on my machine). Now tracking down where it broke...

JamesMolson:
It was https://github.com/SixTrack/SixTrack/pull/261


The javier test was updated, but not the bignpart one.

kyrsjo:
Doh, of course. Thanks.

But yeah; let's rather make the file shorter before commiting it. No reason we have to write every turn for 10k turns and 128 particles; we could reduce both the number of turns and the output frequency.

Also, I suspect this may have messed up your long-running cmake branch...

JamesMolson:
Simple enough to fix that branch :)

kyrsjo:
I wonder what happens to this PR if you do a force push to delete the commit?

JamesMolson:
Lets find out!

JamesMolson:
Nothing it seems...

kyrsjo:
I can even still open the commit in GitHub? Strange. Maybe the branch pointer is gone, but the commit is still there? However the "reopen" button is now grayed out.

Interesting.

PR #291, opened 2017-06-19
Fix ppc64 tests

JamesMolson:


kyrsjo:
No objections. Debug build works too?

However I find it strange that `-fno-inline-functions` has an effect, and it would be useful to find out which expensive optimization is causing the problem.

JamesMolson:
Debug: http://abp-cdash.web.cern.ch/abp-cdash/viewTest.php?buildid=40796

kyrsjo:
See it. Please go ahead and merge!

PR #294, opened 2017-07-04
Check machine length

### LINKED ISSUE #151: Machine length issues
Check machine length "tlen" in sync block VS calculated from `cadcum`. If they are off by more than epsilon (or bigger tolerance?), then print warning.

Also: In cadcum, print the machine length.

This made problems for @Pingul, creating weird z-oscillations on top of synchrotron oscillations, when the machine length in SYNC was off by 2 cm!

kyrsjo:
Fixes #151 

Not completely done yet; I need to do something about the tests.

Currently the threshold is set to 1 um.

 Broken tests and diff [m] (did not check prob1/prob3):
 * distance 0.0002000043
 * bbe52    0.0002000036
 * bbe51 0.0002000036
 * bbe571ib0 0.0002000036
 * dynk_globalvars 0.0192001398
 * eric 0.0002000122
 * frs60 0.0002000036
 * frs 0.0002000036
 * s316 0.0002000122
 * thick6dblocks -56.0000000000
 * elensidealthck6d -56.0000000000
 
Note the two last ones... Also there seems to be the same numbers showing up quite often.


rdemaria:
Can we reassign tlen to dcum(iu+1)

and recompute

halc2=harm/tlen
hsy(3)=(two*pi)*halc2

hsyc(j)=((two*pi)*ek(j))/tlen

instead of issuing an error?

kyrsjo:
If so, the tlen + derived quantities will be different in the initialization and in the tracking.

The "ideal" solution to this is probably to move the initialization of SYNC to a later point in the code, after all elements are initialized (so that the el setting is really always length and never phase of accelerating cavity / crab cavity etc.) but before the optics etc. is initialized.

Also, not having some kind of "tolerance" but actually always changing tlen (and derived quantities) would mean updating almost all 6D test cases... Having it as an error at least traps the issue.

I could make the tolerance settable, to allow the user to say "whatever, I will use that which comes out of my converter, so that my batch jobs never fail", but still catch cases of unknown bad input files?

rdemaria:
I don't understand the first comment in parentheses. Can't we say that the
machine length information is discarded from version 3.7?

I would like to avoid to invalidate input files for an issue we don't
understand much and at the same time not to offer a real improvement.

kyrsjo:
The comment in the parenthesis is referring to that the cavity (acc and crab) phase is read into the el array in fort.2, and not pulled out of there before in initialize_elements. So summing up the el's, as is done in cadcum, won't work.

Furthermore, the number of elements iu isn't really avaliable before in subroutine `ord`.

kyrsjo:
There is also the issue about backwards compatibility; by forcing the user to take action, there are no surprises.

However we could add a flag to ignore the check, allowing old "technically invalid" files to be used with minor modifications.

kyrsjo:
In general, maybe we could add a STRICT flag at the beginning of fort.3, which would turn on strict input checking - that TLEN matches, number of fields in the input lines of fort.3 etc? This way we could conserve compatibility with old input files, while new files would be properly checked.

rdemaria:
I think it is a bit silly to ask the user to write a quantity SixTrack is able to compute at a certain stage in particular for an issue that is not confirmed. At the moment if you run out of time I would leave it open or we implement the correct logic that the machine length is defined by lattice. I really think other solutions make things unnecessarily more complicated.

kyrsjo:
Following the discussion with Joel @Pingul and followup on what was the actual problem in that case, I've changed it from a fatal error to just a warning.

I think we can merge it now (just bump the version number).

JamesMolson:
What was the actual problem then? :)

rdemaria:
The effect seen was due the closed orbit not being added in the initial conditions resulting in very large betatron oscillations and not due to an inaccurate machine length.

PR #315, opened 2017-08-21
Fixed some code formatting that broke syntax highlighting

Jadzia626:
Beginning a parenthesis twice in a +if block broke syntax highlighting. These repeated code lines were also inconsistent in the source. Parentheses are now closed properly in all four blocks.


Jadzia626:
Also submitted the issue to the repository for the syntax highlighter that broke on this.
For reference: https://github.com/dparkins/language-fortran/issues/101

PR #316, opened 2017-08-23
New flag for line 3 in collimation block for reading a normalized dis…

jfwagner:
…tribution which is then transformed in SixTrack internally to physical coordinates. Code needs to be modified with TAS matrix (the inverse normalisation matrix T).

### DIFF
--- a/SixTrack/collimation.s
+++ b/SixTrack/collimation.s
@@ -149,6 +149,7 @@
       myalphay = talphay(1)
       mybetax  = tbetax(1)
       mybetay  = tbetay(1)
+
 !07-2006      myenom   = e0
 !      MYENOM   = 1.001*E0
 !
@@ -391,6 +392,12 @@
      &           myenom, mynex, mdex, myney, mdey,
      &           myx, myxp, myy, myyp, myp, mys,
      &           enerror, bunchlength )
+         elseif(do_thisdis.eq.6) then
+            call readdis_norm(filename_dis, 
+     &           mynp, myalphax, myalphay, mybetax, mybetay,
+     &           myemitx0_dist, myemity0_dist, myenom, 
+     &           myx, myxp, myy, myyp, myp, mys, 
+     &           enerror, bunchlength)
          else
             write(lout,*) 'INFO> review your distribution parameters !!'
             call prror(-1)
@@ -7591,6 +7598,84 @@ c$$$     &           myalphay * cos(phiy))
       call prror(-1)
       
       end
+
+!
+!========================================================================
+!
+      subroutine readdis_norm(filename_dis, 
+     &           mynp, myalphax, myalphay, mybetax, mybetay,
+     &           myemitx, myemity, myenom, 
+     &           myx, myxp, myy, myyp, myp, mys, enerror, bunchlength)
+
+!     Format for the input file:
+!               x, y   -> [ m ]
+!               xp, yp -> [ rad ]
+!               s      -> [ mm ]
+!               DE     -> [ MeV ]
+!
+      implicit none
+
++ca crcoall
++if crlibm
++ca crlibco
++ei
++ca collpara
++ca dbmkdist
+
+      character*80   filename_dis
+      double precision enerror, bunchlength
+      
+      logical lopen
+      integer stat
+      
+      double precision normx, normy, normxp, normyp, normp, norms
+
+      save
+
+      write(lout,*) "Reading input bunch from file ", filename_dis
+
+      inquire( unit=53, opened=lopen )
+      if (lopen) then
+         write(lout,*) "ERROR in subroutine readdis: "//
+     &        "FORTRAN Unit 53 was already open!"
+         goto 20
+      endif
+      open(unit=53, file=filename_dis, iostat=stat,
+     &     status="OLD",action="read")
+      if (stat.ne.0)then
+         write(lout,*) "Error in subroutine readdis: "//
+     &        "Could not open the file."
+         write(lout,*) "Got iostat=",stat
+         goto 20
+      endif
+
+      do j=1,mynp
+         read(53,*,end=10,err=20) normx, normxp, normy,
+     &     normyp, normp, norms
+         myx(j)  = normx  * sqrt(mybetax*myemitx)
+         myxp(j) = sqrt(myemitx)*(-normx*
+     &     myalphax/sqrt(mybetax) + normxp/sqrt(mybetax))
+         myy(j)  = normy  * sqrt(mybetay*myemity)
+         myyp(j) = sqrt(myemity)*(-normy*
+     &     myalphay/sqrt(mybetay) + normyp/sqrt(mybetay)) 
+        myp(j)  = myenom * (1d0 + normp * enerror)
+        mys(j)  = bunchlength * norms
+      enddo
+      
+ 10   mynp = j - 1
+      write(lout,*) "Number of particles read from the file = ",mynp
+
+      close(53)
+
+      return
+      
+ 20   continue
+!      call abend('I/O Error on Unit 53                              ') !ABEND is for the CR version
+      write(lout,*) "I/O Error on Unit 53 in subroutine readdis"
+      call prror(-1)
+      
+      end
+
 !
 !========================================================================
 !


kyrsjo:
Hi Joscka, regarding the normalization - I just spoke to Stefano, and he pointed out that for the distributions that are generated internally in the collimation part, there already is the code to normalize them.

It may be that it is easier for you to just look into what they are doing there, than it is to use the TAS matrix.

Cheers,
Kyrre

JamesMolson:
Also surely in any region with dispersion, this isn't going to match the distribution into the lattice.

kyrsjo:
You mean the collimation normalization? True. The TAS 6x6 matrix should however work fine...

jfwagner:
Right, the generator of distributions in SixTrack "makedis_.." doesn't take in account the dispersion. I first followed how it is done there taking only "tbetax,tbetay,......". In order to have the full transformation I would like to use now the TAS matrix in SixTrack , so that it is complete. I'll come up with a new pull request then.

kyrsjo:
Joscka, I don't think you need to make a new PR (unless we should have both "collimation style" normalization and "SixTrack style normalization" as two separate options. Even then I would consider making it one single PR.

For the one you have now, you showed me a plot with two scatter plots that looked different. The plots were however somewhat hard to read. Did you calculate the Sigma matrices and compare those?

To check if dispersion really is what makes the difference, try to send in a distribution with no energy spread.

Also, in your code, I see that you're using alpha, beta, and emittance for x and y, plus `enom`, `enerror` and `bunchlength`. Could you try to just write those out from inside your function, then write the same equations in a new Python script? That will be an easier way to debug your equations.

jfwagner:
Where and how do I add the tas matrix correctly for the routine? Is it in a common block?
For the variable "ia", what do I have to set? I understood it stands for a particle.

SixTrack/collimation.s
```diff
@@ -7591,6 +7599,156 @@ c$$$     &           myalphay * cos(phiy))
...
+      character*80   filename_dis
+      double precision enerror, bunchlength
+      
+      logical lopen
+      integer stat
+      
+      double precision normx, normy, normxp, normyp, normp, norms
+      double precision myemitz
+
+      save
```
kyrsjo: Remove save

SixTrack/collimation.s
```diff
@@ -7591,6 +7599,156 @@ c$$$     &           myalphay * cos(phiy))
...
+      
+ 10   mynp = j - 1
+      write(lout,*) "Number of particles read from the file = ",mynp
+
+      close(53)
+
+      return
+      
+ 20   continue
+!      call abend('I/O Error on Unit 53                              ') !ABEND is for the CR version
```
kyrsjo: Delete the "abend" line

kyrsjo:
TODO:
 - [ ] Documentation - Ask @amereghe for best place to put it (Twiki)?
 - [ ] Two small comments in the code
 - [ ] In `subroutine daten`, when reading collimation block: If we get `do_thisdis`.eq.6 and iclo6.eq.0 -> abort, since we're not sure about the meaning of the TAS matrix in that case. Could you just in case make a post here, where you put both the iclo6=0 and iclo6!=0 TAS matrix for comparison.
 - [ ] Check and post here the proof that collimation works fine with iclo6 .ne. 0, @rbruce1 should agree.
 - [ ] @mfittere : Could you take a quick look that the equations and comments about the units look OK to you?

Then: Merge whatever is in upstream/master, bump the version number, and we'll merge this into master. Thanks!

jfwagner:
TAS matrix for iclo6=2
  2.4490304227567807     
  2.47780511296652568E-019
  2.48051493399166102E-005
  5.93122320542262313E-006
  3.87188223445383334E-006
   1.3333137678185949     
 -5.98709954997833841E-005
  0.40832485869749385     
 -1.02250112775557723E-006
  3.94985062792144036E-006
  1.27863652757680526E-006
  7.30280710570202912E-002
 -2.37001983414170955E-005
  5.92610125569660572E-006
   2.4504254017805711     
 -4.30407096366375667E-018
 -1.38471798405412065E-005
  0.60825884956855658     
 -1.02052616505373720E-006
 -4.13335093391638196E-006
 -7.56967286058323803E-005
  0.40809240721010737     
  5.46118317896008609E-007
  4.86203918816133923E-002
 -4.70780852368622455E-003
  1.43244587426834252E-002
 -3.13603995059776128E-003
  6.53127062278691106E-003
   26.311297990609848     
 -2.16311039430162942E-015
  1.30119614841343685E-010
 -9.38563656006667675E-011
  5.82151309839940623E-011
  1.99375703487504188E-010
 -6.20227999259797376E-008
  3.80064869009651626E-002
 closed orbit
 clo6v  0.75009695492214201     
 clop6v  4.74204932660131131E-006
 clo6v  2.94384720216672480E-005
 clop6v  0.29500180796499215     
 clo6v  1.37315946482529181E-002
 clop6v  1.83619243678748312E-006
 clo6v  0.75009695492214201     
 clop6v  4.74204932660131131E-006
 clo6v  2.94384720216672480E-005
 clop6v  0.29500180796499215     
 clo6v  1.37315946482529181E-002
 clop6v  1.83619243678748312E-006

jfwagner:
TAS matrix for iclo6=0
2.44948958e+00
-2.28193738e-16
 -1.08734187e-05
 -4.65247336e-07 
0.00000000e+00 
0.00000000e+00
-3.75306172e-07
 4.08248318e-01
 5.09808014e-08 
-1.78621037e-06
 0.00000000e+00 
0.00000000e+00
1.07216458e-05
 -4.73747855e-07
 2.44948959e+00
 2.46661717e-18
 0.00000000e+00
 0.00000000e+00
5.10331281e-08
 1.81245749e-06
 -2.35155861e-07
 4.08248316e-01 
0.00000000e+00
 0.00000000e+00
0.00000000e+00 
0.00000000e+00 
0.00000000e+00 
0.00000000e+00
 0.00000000e+00 
0.00000000e+00
0.00000000e+00
 0.00000000e+00
 0.00000000e+00
 0.00000000e+00
 0.00000000e+00
 0.00000000e+00

jfwagner:
I checked the file collgaps.dat for both cases, using the flag iclo6=0 , 1 and 2. The values are all the same, so everything is fine.

jfwagner:
The Twiki is updated:
https://twiki.cern.ch/twiki/bin/view/LHCAtHome/SixTrackCollimatVer#readdis_norm

kyrsjo:
A lot of places where the different options for do_thisdis is mentioned in the TWIKI does not mention the new option...

kyrsjo:
Also please wait with bumping the version number until I give you the go-ahead, and you have just merged from master. Right now the only reason there is no conflict is that I changed the lines with the version and date in sixtrack.s to *exactly* the same values as you did (i.e. if you merge now there won't be any version bump) ...


Regarding the TAS matrix, this is very strange RE what we did yesterday - @mfittere ? @jfwagner could you post your input files?

With "all the same" do you mean exactly the same down to the last digit? could you please post the collgaps file(s)? (you can add them as attachments).

jfwagner:
The collgap files are exactly the same.

kyrsjo:
Thanks for confirming that the collgaps files did not change. I assume this was a full normal LHC collgaps file with a bunch of different collimators at different settings?

I'm still confused about the TAS matrix tough, as Miriam and I came to the result that they did *not* change when we tried something similar (for the normalized StartDUMP, PR #322 ). If you could please post the input files so that I could take another look?

Regarding the version number and merging from master; I'll do that when I'll merge it.

kyrsjo:
Example input files I got from jwagner: 
[read_norm_distr_test_iclo6.tar.gz](https://github.com/SixTrack/SixTrack/files/1444238/read_norm_distr_test_iclo6.tar.gz)

kyrsjo:
I think you have forgotten adding the closed orbit itself (not just the crossing angle)?

kyrsjo:
Never mind, now I see both your comment and the code it is referencing...

kyrsjo:
I've noted that the 4D linopt, which is used for positioning collimators, is completely independent of the clo6d setting; the differences happen later in the execution of `program maincr`. So while changing from 4- to 6d may change the closed orbit a little bit, collimator positions will not change at all.

kyrsjo:
OK, from my point of view this is done. @rdemaria , could you take a look?

This should DEFINITIVELY be a version bump. I will merge this, but only after having spoken to Riccardo.

kyrsjo:
I just spoke to Riccardo; I will update the docs and merge this tonight.

For the future, when the DIST block comes it may be a good idea to review the structure of the TRACK and INIT block, including maybe making an alternative TRAC block (and removing the thin/thick6dua stuff). But let's do one thing at a time.

PR #319, opened 2017-08-28
update build manual

mfittere:
I think this is clearer ...


kyrsjo:
Thanks!

I added a few things and fixed the formatting - other places lstlistings is only used to get commands the user can type.

PR #321, opened 2017-08-31
Add a usefull little script for creating tags the right way

kyrsjo:


kyrsjo:
Usage:
`./addtag.sh v1.2.3 SHASUM`

Then add the release using the github interface...

PR #329, opened 2017-09-14
FMA allow for analysis over firstturn lastturn

mfittere:


kyrsjo:
Just to summarize the test changes:
* Tests `fma` and `fma_binary`: Added tests with "START_DUMP_1" and "IP3_DUMP_1" for a subset of the file 
* Tests `dump8_fma` and `dump7_fma`:  Added tests with "IP3_DUMP_2" and "START_DUMP_1" on a subset of turns.
* All FMA tests (NAFF was forgotten?): Updated fma_sixtrack.canonical to take 3 new colums into account

Rewiewing the code now. Will then run the tests, merge from master, run the tests again, and if all is good: bump version and merge.

Thanks for the work!

Doc/user_manual/six.tex
```diff
@@ -3318,7 +3319,8 @@ \subsection{FMA analysis}
 \subparagraph{fma\_flag\_norm}
 optional flag for calculating the tunes with physical ($x$,$x'$,$y$,$y'$,$s$,$dp/p$) or normalized coordinates in case physical coordinates are used in DUMP. The default is using normalized coordinates (\verb|fma_flag_norm=1|). For using physical coordinates explicitly set (\verb|fma_flag_norm=0|). See \textbf{Description} for the conditions under which normalization is available.
 
-
+\subparagraph{fma\_first\_turn, fma\_last\_turn}
+Turns used for FMA analysis. As the DUMP files are used as input for the FMA analysis \texttt{fma\_first\_turn} must be larger \texttt{first\_turn} in the DUMP block and  \texttt{fma\_last\_turn} must be smaller than \texttt{last\_turn} in the DUMP block. If \texttt{fma\_last\_turn = -1} the number of turns tracked is used as last turn number.
```
kyrsjo: Default value: 0,0 -> dumpfirst,dumplast

SixTrack/sixtrack.s
```diff
@@ -50382,7 +50437,15 @@ c$$$            endif
...
-            do k=1,fma_nturn(i) !loop over turns
+            ! If normalization within FMA, we now have to always write the full NORM_* file
+            ! Otherwise  one would overwrite the NORM_* file constantly if different FMAs are done
+            ! on the same DUMP file
+            if (dumplast(j) .eq. -1) then
+              dump_last_turn = numl
+            else
+              dump_last_turn = dumplast(j)
+            endif
+            do k=dumpfirst(j),dump_last_turn !loop over turns, use the dump files
```
kyrsjo: Problem: if dump_last_turn > fma_nturn_max, the array index k will be too big and we'll segfault, even if the DUMP itself is short enough to fit into the array. We need to use an offset either when writing or reading the array.

Also, we're reading the full dump (repeatedly), which means that even if the number of turns we're trying to do the FMA over is small enough, the full dump needs to be < fma_nturn_max.

SixTrack/sixtrack.s
```diff
@@ -50086,6 +50096,7 @@ c$$$            endif
      &epsnxyzv ! normalized emittances
       double precision :: tunelask,tuneffti,tunefft,tuneapa,tunefit,    &
      &tunenewt,tuneabt2,tuneabt,tunenewt1 !Define the functions to be called from PLATO
+      double precision :: dump_last_turn ! auxiliary variable for loop over turns
```
kyrsjo: Thou shall not use "double" variables as loop indices. I'm changing it to an int. This even made the compiler scream...

kyrsjo:
TODO: Update the Sixin.zip files.

SixTrack/sixtrack.s
mfittere: Did you fix this?

kyrsjo:
OK, I think I fixed the outstanding issues. Also avoiding writing the NORM dumps over and over and over again saved about 1 minute of the FMA test, when running on a SSD...

@mfittere if you could please review my changes and let me know if you're unhappy, if you are then please let me know otherwise I'll merge tomorrow.

kyrsjo:
To check: Do we now still produce the NORM dump if no normalization matrix is available etc.? We should not do that; it will be full of zeros...

PR #334, opened 2017-09-27
added error message for 6D flag in LIN BLOCK when using collimation v…

jfwagner:
…ersion.


kyrsjo:
Hi Joschka, many thanks!
However if I understand correctly, this checks just that the LINE block is NOT 6D; would it not be better to check that it IS in 4D?

Cheers,
Kyrre

PR #335, opened 2017-10-02
Dynk FreeForm and Memory Allocation

vkbo:
### Changes
* Reflow of dynk module to free form
* Moved checkpoint restart subroutines from sixtrack.s to dynk.s90
* Added dynamic allocation of memory for data arrays


rdemaria:
I would not be able to have a better judgement than the test suite, if this passes all tests I would merge.

What is the test coverage for DYNK?

vkbo:
Kyrre would be the best one to say, but I've tested the memory allocation and changes to the file read function with the globalvars test. This test reads 40001 float values from a file, and it works.

The checkpoint restart changes have also been tested extensively.

Kyrre also read through the reflowing of the code to free form, and found no mistakes. There's always the chance of a bug, but mostly what I've done there is to change indentation and change the line splitting which was incompatible with free form.

JamesMolson:
No version bump?

rdemaria:
Although it is major restructuring, I thought it would be fine without changing the version, since it does not introduce new features. Do you think it would be better to update the version?

kyrsjo:
Yes, it should absolutely have been a version bump.

It does introduce a new feature (you can now load big FILEs etc. without modifying the sources), it changes how SixTrack deals with C/R files (which are now incompatible, however the compatibility is checked by reading the version number in the header of the CR file...), in case bugs were introduced having a version (and date, that should anyway be bumped!) bump makes things possible to track down, and also having a useful commit message when merging this is usually a good thing...

vkbo:
Yes, most of the code doesn't change functionality. BUT there are changes to how restart files for DYNK are written, and the limit of how large files can be read are now a matter of memory, not the pre-defined size of data arrays.

rdemaria:
Good! Who does volunteer to bump the version? ;-)

kyrsjo:
I'm doing it now. Note that I haven't read the diff...

kyrsjo:
OK, see d6cef4a176ac619cf257a16e5977d4bcee6b99ef ; this combines the commit message and the version / date bumping.

Note that by updating the fort.3 in dynk_globalvars and not the one in the corresponding Sixin.zip, we probably broke one of the check scripts...

Regarding not version bumping - this rewrite is pretty much comparable to what @JamesMolson did with the collimation splitting, which triggered the bump from 4.6.x to 4.7.x - except that in that case, we were *explicitly* did not change *any* functionality.

rdemaria:
Thanks!

Issue #95, opened 2016-08-23
dipedge element


rdemaria:
Not document, nor tested. Not clear if non-linear effect is present.

kyrsjo:
Fixed in #340. Dipedge is now documented, and matches the MadX interface.
Testing for consistency already happens in the SixTest test "dipedge".

PR #340, opened 2017-10-05
Dipedge doc

tpersson:
Added documentation about the dipedge element to the manual. The implementation was correct in the codes so I just added it to the manual.


rdemaria:
Ok!

Can you put the definition of r_21 and r_34 (the equation used in your presentation would be enough)? Also a reference to the physics guide when you mentioned it would be useful, if not there already.

kyrsjo:
Perfect, many thanks! This is great news - we can now run with dipedge! We should however check how this affects the DA and collimation before turning it on by default in the MadX scripts...

Can you please add your name to the front page of the document, set the version to the current SixTrack version, and we shall merge it immediately!

kyrsjo:
Perfect, thanks - I merged it! You may now delete your branch from GitHub (if you like) by clicking the "delete branch" button just above.

Furthermore, this fixes issue #95 .

PR #343, opened 2017-10-16
fix fftw3 and boost include for NAFF

rdemaria:


kyrsjo:
Please hold off with this, as @skostogl has a potential patch to remove BOOST.

JamesMolson:
This can be closed now?

PR #345, opened 2017-10-19
Fix building and testing with nagfor

JamesMolson:


kyrsjo:
Thanks! If `flush` now works, we could replace all the `endfile; backspace;` stuff in sixtrack CR...

PR #346, opened 2017-10-20
Fix static linking with NAGLIB and astuce++

JamesMolson:


kyrsjo:
Any idea why this change was needed - AFAIK it used to work? New version of NAG?

PR #349, opened 2017-10-24
Scatter Update

vkbo:
* Added descriptive(-ish) headers to subroutines like in the DYNK module
* Added the input parsing and calculation for Gaussian profile


vkbo:
Don't merge quite yet. Need to update documentation.

SixTrack/scatter.s90
```diff
@@ -958,23 +1035,47 @@ module scatter
...
-        retval = scatter_fdata(scatter_PROFILE(profileIdx,3))
+        retval  = scatter_fdata(scatter_PROFILE(profileIdx,3))
     case (10) ! GAUSS1
-        ! TODO
+        beamtot = scatter_fdata(scatter_PROFILE(profileIdx,3))
+        sigmaX  = scatter_fdata(scatter_PROFILE(profileIdx,3) + 1)
+        sigmaY  = scatter_fdata(scatter_PROFILE(profileIdx,3) + 2)
++if crlibm
+        retval  = beamtot/(2d0*(pi*(sigmaX*sigmaY))) &
+                * exp_rn(-(x**2)/(2d0*sigmaX**2)) &
```
kyrsjo: Please avoid the ** operator - it does not optimize well compared to just saying `x*x` (as it is AFAIK really a call to `pow(x,float(2))`), and it may be somewhat implementation-defined...

SixTrack/scatter.s90
```diff
@@ -958,23 +1035,47 @@ module scatter
...
+                * exp_rn(-(y**2)/(2d0*sigmaY**2))
++ei
++if .not.crlibm
+        retval  = beamtot/(2d0*pi*sigmaX*sigmaY) &
+                * exp(-(x**2)/(2d0*sigmaX**2)) &
+                * exp(-(y**2)/(2d0*sigmaY**2))
++ei
+        write(lout,*) retval, x, y, sigmaX, sigmaY
```
kyrsjo: Forgotten debug statement?

vkbo: yes

SixTrack/scatter.s90
vkbo: ** and pow compiles to the same, so I don't really see your point. Do you have a source for this?

vkbo: As far as I can find out x**2d0 is faster than x**2 is faster than x*x for both ifort and gfortran, not to mention more readable.

rdemaria: Last time a check (~10 years ago) I saw by looking at the asm that the `**` operator was handled in a smart way by fortran compilers. This for exponent larger than 2 though. I am a bit puzzled that x*x is compiled differently x**2. 

Just for the sake of discussing, should not  `((x/sigmaX)**2)/2` just slightly more efficient?

vkbo: I just ran `-O1`, `-O2`, and `-O3` tests with both gfortran and nagfor (don't have ifort), and not only do they all optimise well, the `**` notation is also marginally faster.

And, yes, that may be more efficient I suspect. In any case, my computer performed 100M such exponents in ~200 ms, so I doubt this is much of a bottleneck.

kyrsjo: Interesting that the pow call is actually faster than just x*x, thanks for doing the test! I know that ** and pow are equivalent, as written in my initial comment.

However my main worry was not the speed, but whether the result would be the same on all compilers for all `x`.

vkbo: Ok, but you did write "optimisation", which I take to mean efficiency rather than precision. In any case, the speed test was also for Riccardo's comment.

If it is all compiled to the same assembly, which is at least true for ** and pow, it really does not matter.

kyrsjo: Yeah, I was worried that the optimization for efficiency might arrange things differently on different compilers, resulting in slightly different results. The fact that it's equivalent to POW means it may result in a library call...

kyrsjo:
OK, I see now that even the `**` seems to be used everywhere - I guess it's safe then.

If so, and you're ready, just bump the version number + date and we'll merge. Remember to merge from upstream first!

SixTrack/scatter.s90
JamesMolson: x**2 and x*x do a multiply

x**2D0 and pow(x,2) calls _pow_

PR #350, opened 2017-10-24
Boost removed

skostogl:
minima.hpp has been added, which contains the code from Boost concerning the Brent minimization algorithm.


SixTrack/naff/windows.cc
```diff
@@ -117,7 +116,7 @@ std::vector<std::complex<double>> taylor_window(const size_t N, const double SLL
   std::vector<std::complex<double>> out;
   std::vector<double> ww(N);
   double A = acosh(pow(10, SLL)) / pi;
-  double NBAR = boost::math::round((2 * pow(A, 2) + 0.5)+0.5);
```
kyrsjo: Cool!

One request, could you name it as something else than just `round`, rather `boost_round` or something like that? There is so much stuff going on with rounding in this code already, let's not confuse things (or confuse the linker)...

For the cmath stuff, the fact that these are overloaded C++ functions should make sure they are name-mangled and therefore unique at compile time.

kyrsjo:
Thanks a lot! Once you've done the scan we discussed + fixed the little naming things (I commented on), I'll version-bump and merge.

Then I'll handle the build system / FFTW mess, basically #343 ...

kyrsjo:
Another FYI - I see you're committing to master. Generally we're trying to keep the master branches of the forks as mirrors of the upstream master, i.e. with the same history. This makes starting work on a new branch easier.

This is fairly easy to fix - it basically requires
1. Making a new branch pointing to this commit:
`git checkout -b remove_boost`
2. Rolling back your "master" branch to the one it started as, i.e. 5d0985da8fec86a3eaaa92e0d29f4223a7080b8a :
```
git checkout master #Make master the active branch
git reset --hard 5d0985da8fec86a3eaaa92e0d29f4223a7080b8a
```
See https://www.atlassian.com/git/tutorials/undoing-changes for information about `git reset`.
You should now have one branch `remove_boost` pointing to your two "Boost removed" commits, and another branch `master` being a copy of `upstream/master`. Make sure this is the case before proceeding
3. Overwrite the current `master` branch pointer on github as well:
`git push --force`

Now you will have to recreate this pull request, unless you can somehow change it to be a pull request from skostogl:remove_boost instead of skostogl:master. In case of problems, just prod me, Veronica, or James.

Thanks!

kyrsjo:
We've decided to do the git reset stuff later (before Sofia starts working on something else in her fork), and rather just merge this "master".

Can I now go ahead and to the merge (including update of this branch + version bump)? If so, I will do it tonight.

Sofia, if you could please post the plots you sent me here for later reference, that would be great!

skostogl:
Comparison of footprint with and without BOOST

boost_comparison

kyrsjo:
Hi, I just checked, and on my machine the fma_sixtrack.canonical was totally unchanged, which is good news! I therefore reverted that change.

skostogl:
Yes Kyrre, sorry, I changed the fma_sixtrack.canonical according to my results and I forgot that we have some small discrepancies and I did not change it back. Thanks

PR #351, opened 2017-10-25
SixTrack 5.0 DEV [was: Add initial support for quad precision floating point math.]

JamesMolson:
Some initial support for setting floating point precision at compile time.

i.e.

single, double, quad.

In each subroutine you will need to add at the top:

"use module floatPrecision"

then when defining variables, do not do "double precision xxx", but instead do "real(kind=fPrec)  xxx"

where fPrec is set to real32, real64 or real128 in the module.

This currently defaults to real64, so the same as before.


JamesMolson:
Point number 2:

When specifying constant values, for example "5.2", doing it as so will make the number use single precision storage.

Specifying "5.2D0" will use double.

but specifying "5.2Q0" only works on some compilers.

The standard way to do this is to use "5.2_real128" etc.

rdemaria:
I have the impression that this will duplicate a lot of code. Would it be possible to write e.g. 5.2_dp and let astuce replace _dp with real32 or real128 when asked?

JamesMolson:
Yes, this is the case, so I can remove that duplicated block of variables.

rdemaria:
Same with crlibm, we keep crlibm branch only in the code and astuce will replace all special function with their _rn version.

JamesMolson:
Another note: calls to dble(foo) should be replaced with real(foo,PrecisionType)

JamesMolson:
All fast and medium 64bit tests seem to pass so far.

CMakeLists.txt
```diff
@@ -103,6 +103,15 @@ OPTION(AVX2 "Enable use of the Advanced Vector Extensions 2 (AVX2) instruction s
 
 OPTION(BUILD_TESTING "Enable the SixTest suite" OFF)
 
+OPTION(32BITM "Floats are 32bit (single precision)" OFF)
+OPTION(64BITM "Floats are 64bit (double precision)" ON)
+OPTION(128BITM "Floats are 128bit (quad precision)" OFF)
+
+OPTION(ROUND_NEAR "Always round to the nearest number after floating point operations" ON)
```
kyrsjo: These should depend on CRLIBM, otherwise they make no sense.

SixTrack/common_headers.s
```diff
@@ -1,6 +1,88 @@
...
+      parameter(c1m9 = 1.0d-9,c1m10 = 1.0d-10,c1m12 = 1.0d-12)
+      parameter(c1m13 = 1.0d-13,c1m15 = 1.0d-15)
+      parameter(c1m18 = 1.0d-18,c1m21 = 1.0d-21,c1m24 = 1.0d-24)
+      parameter(c1m36 = 1.0d-36,c1m38 = 1.0d-38)
+!     electron mass from PDG, 2002
+      parameter(pmap = 938.271998d0,pmae = .510998902d0)
+      parameter(crade = 2.817940285d-15, clight = 2.99792458d8)
++ei
++if 64bitm
+      double precision c180e0,c1e1,c1e12,c1e13,c1e15,c1e16,c1e2,c1e3,   &
```
kyrsjo: define as `real(kind=fPrec)`?

SixTrack/sixtrack.s
```diff
@@ -18524,15 +18526,16 @@ c$$$         endif
...
       implicit none
       integer maxf
       ! MAXF be kept in sync with maxf in various routines
       ! We maybe should use len(field(f)) here, like is done in splitfld...
       parameter (maxf=30)
       integer errno,f
       character*(*) fields(*)
-      double precision round_near,value
+      real(kind=fPrec) round_near,value
       fround=round_near(errno,maxf,fields(f))
```
kyrsjo: Hmm... `SixTrack/roundctl/round_near.c::double round_near_(int*,int*char*)` does never return a quad... It will always returns a double.

SixTrack/sixtrack.s
```diff
@@ -18687,9 +18692,10 @@ c$$$         endif
       return
       end
       
-      double precision function acos_rn(x)
+      real(kind=fPrec) function acos_rn(x)
+      use floatPrecision
       implicit none
-      double precision atan_rn,x,pi,pi2
+      real(kind=fPrec) atan_rn,x,pi,pi2
```
kyrsjo: Ditto for atan_rn - this bouncer function will always return a double...

kyrsjo:
Cool! However there is a trap - function calls! Fortran (unless you have INTERFACE or MODULE) does not check that the the arguments match when calling, so here be dragons...

For the sin/cos/etc. functions we can anyway just extend (and rename!) the current redirection function and use some #ifdefs - this will make the rounding up/down/near/without CRLIBM easy. The performance cost is minimal - we're anyway using a redirection for CRLIBM, and it will avoid a LOT of code duplication.

Then there is obviously the problem with the constants... How many are there really, outside of parnum? How many of them should really really be in parnum?

CMakeLists.txt
JamesMolson: Todo in the future :) need an updated CRLIBM first!

SixTrack/sixtrack.s
JamesMolson: Yes, all these functions need updating

JamesMolson:
In reply to your earlier comment:

```
Error: Type of argument 'a' in call to 'idint' at (1) should be REAL(8), not REAL(16)
```

```
Warning: Type mismatch in argument 'sca' at (1); passed REAL(8) to REAL(16) [-Wargument-mismatch]
```

gfortran does (seem) to catch them

CMakeLists.txt
kyrsjo: Not really! It should all be there.

rdemaria:
5.0 Dev ???

kyrsjo:
Yes, we kind of decided to clean up a lot of stuff... Take a look at the diff.

Basically, in addition to adding the 128bit stuff (which triggered a lot of cleaning), we wrote a wrapper for all the mathlib stuff. This means there are now no difference in the mathematical expressions for the crlibm- and non-crlibm version; you anyway call `sin_mb` etc. and then the bouncer handles crlibm or not etc.

Furthermore, we've converted parts of the code into freeform, read90 is being cleaned up, parts of the code are being updated so that it is compliant with Fortran2008, and various other fixes.

Let's have lunch tomorrow...

rdemaria:
I am sure I don't need to say but the strategy looks to me a bit risky without proper test and performance regression coverage. It would be better if the changes could be done in chucks of one week that can be merged upstream, then asses the impact and continue.

kyrsjo:
James just made me aware of the comments regarding the beam matrix; however I think this is now already done by DUMP? There are something like 10 different formats now, some of them writing one line per particle, others writing one line per turn with statistical information (suggested by Riccardo).

If something is missing (or people have postprocessing tools that expect a certain format and would like to avoid changing them), I think it is much better to just have one more format than to duplicate all the input reading, data structures, sanity checks, checkpointing etc.

amereghe:
Hi Kyrre, thanks for the remark. If you are sure that this is the case, then please ignore my comment, and proceed. I just wanted to make sure that we are not forgetting anything useful.
No worries for the output format - and absolutely war against double code!

JamesMolson:
One for @kyrsjo 

FMA is broken with BIGNBLZ/HUGENBLZ.

after some fun with git bisect

```
2c121e4cf6e441598f01fcd057766db4426e7c81 is the first bad commit
commit 2c121e4cf6e441598f01fcd057766db4426e7c81
Author: Kyrre Sjobak <kyrre.ness.sjoebaek@cern.ch>
Date:   Tue Oct 31 21:32:48 2017 +0100

    put FMA in a separate file

:100644 100644 3ae3443c257c500316c8c936495c05ef1efa7443 e22550867a84251c37e9b53986482da42caf2f4b M	CMakeLists.txt
:040000 040000 993d16ae53bb773096a4bdb7a7878c5e35902656 e14b07d8ee66915f12c71a26d521d0ad8dd77fc3 M	SixTrack
```

JamesMolson:
Currently broken things:

1. dump + CR
2. DA build
3. beamgas

Todo:

1. Kill the computed GOTOs and other warnings.
2. Add FLUKA DIST block
3. Clean up any leftover precision issues (collimation code really).
4. Add some collimation test cases for the online aperture check.
5. Test FLUKA code.

JamesMolson:
With that commit, no, with what I've done since (banishing real precision 0.5s, etc,) yes there are changes.

I still need to push that.

The lossmap looks the same, just the impacts on the collimators might change by a few femtometers, so I have changes in the impact and other files.

There is now a simple test case for all this, so its easy to see :)

I would consider the current single precision constant -> double precision variable to be a bug.

kyrsjo:
> real precision 0.5s

0.5 = 2^-1, should be exactly representable... It's the 0.1s (and 0.01 etc.) which are more of a problem. Anyway, it would be nice to see exactly where and why things changed. You mentioned that you cleaned up some mess where the proton mass were defined in several places, as different numbers?

> I would consider the current single precision constant -> double precision variable to be a bug.

Agreed. But it would be nice to single out the commits that change the results.

vkbo:
Still a bunch of computed gotos left. Some of them are really long.

I noticed some where replaced with a list of if statements `if(something.eq.N) goto NN`. That us of course a solution, but it doesn't improve readability. Still, maybe an option to just kill the warnings?

kyrsjo:
Some of the really really long ones are in the tracking blocks. One thing to watch out for there are gotos from inside the +ca blocks... Labelled loops may be one option there, a very nice feature of modernish FORTRAN allowing cycle and break of nested loops.

Regarding the "if (condition) goto XXX" replacements, please then rather leave the GOTOs, it's better to have a bunch of warnings that may at some point annoy us to fix it (or disable the warning with a compiler flag) than to brush the actual problem of not-so-readable logic under the carpet.

vkbo:
I encountered one subroutine where there were gotos from within ca blocks. These were only pointing to the end of the main loop, so I kept the label. Now that I think about it, I may have put it in the wrong place, so will need to double check that they all pointed from outside the main do loop (otherwise they will not trigger a cycle the way I implemented it).

As for the if replacements, I don't know who did those (haven't checked). I only replaced two computed gotos with if statements, and those only had two possible outcomes, so it made sense to do it that way.

JamesMolson:
Closing this, we will now merge into the dev branch.

PR #361, opened 2017-11-01
Energy kick for RF multipoles and fixes of the DA (FOX) part

tpersson:
Update for the RF multipoles:
- Update the RV variable in the FOX part. It was already updated in the tracking.

- 2nd order skew did not include misalignment in the FOX part. This gave strength behaviour. alignf added to it. 

- In the FOX part there was a 10e-3 too much, not present in the tracking part ( please check this in detail since this gave 10-3 everywhere)

- 1/(1+deltap) was multiplied also for the energy kick in tracking. I don't see the reason for this. 

- A 1/(1+deltap)  is now multiplied to actually update the angle in the FOX part. Already present in the tracking. 

- Move from deltap kick to energy kick. Please check this in detail. Very easy to make a mistake somewhere here...  

Things missing:
It is not optimized and some things could be cleaned up. The tests have NOT been updated. I think it would be good to do some proper testing and bench marking. My quick tests show a very small different to previous version except for the FOX part. 

The twiki is also not correct for the RF multipoles. I will update this.


SixTrack/sixtrack.s
```diff
@@ -3883,50 +3883,66 @@
...
+
 
 +if .not.tilt
 +if crlibm
-        yv(xory,j)=yv(xory,j) - crabamp*                                &!hr03
-     &sin_rn((((sigmv(j)/clight)*crabfreq)*2d0)*pi + crabph(ix))         !hr03
-      dpsv(j)=dpsv(j) -                                                 &!hr03
-     &((((((crabamp*crabfreq)*2d0)*pi)/clight)*xv(xory,j))*             &!hr03
-     &cos_rn((((sigmv(j)/clight)*crabfreq)*2d0)*pi + crabph(ix)))*c1m3   !hr03
+       yv(xory,j)=yv(xory,j) - crabamp*c1e3/e0f*                        &!hr03
```
rdemaria: I think you need to divide the voltage by 1/(1+delta) in yv update only because yv is Px/P and not Px/P0, while it is correct in the energy.

SixTrack/sixtrack.s
```diff
@@ -3883,50 +3883,66 @@
...
+     &sin_rn((((sigmv(j)/((clight*(e0f/e0))))*                          &!hr03
```
rdemaria: Look if e0f/e0 is not available already in another variable, or better put ((clight*(e0f/e0)))*crabfreq in a temporary variable like kcrab, this should make the code shorter and slightly more readable

SixTrack/sixtrack.s
```diff
@@ -3883,50 +3883,66 @@
...
+     &((((((crabamp*crabfreq)*2d0)*pi)/clight)                          &!hr03
+     &*xv(xory,j))*                                                     &!hr03
+     &cos((((sigmv(j)/(clight*(e0f/e0)))*crabfreq)*2d0)*pi              &!hr03
+     & + crabph(ix)))                                                    !hr03
 +ei
 +ei
       ejf0v(j)=ejfv(j)
-      ejfv(j)=dpsv(j)*e0f+e0f
-      ejv(j)=sqrt(ejfv(j)**2+pma**2)                                     !hr03
+      ejfv(j)=sqrt(ejv(j)**2-pma**2)                                     !hr03
```
rdemaria: Probably you have already done, these steps should be a carbon copy of the code after the normal cavity

SixTrack/sixtrack.s
```diff
@@ -20569,173 +20661,215 @@ C Should get me a NaN
...
-*FOX  Y(1)=Y(1) - CRABAMP*C1E3*
-*FOX  SIN(SIGMDA/C1E3/CLIGHT*CRABFREQ*2D0*PI + CRABPHT) ;
-*FOX  DPDA1=DPDA1 - CRABAMP*CRABFREQ*2D0*PI/CLIGHT*X(1)*
-*FOX  COS(SIGMDA/C1E3/CLIGHT*CRABFREQ*2D0*PI + CRABPHT) ;
+        if(kzz.eq.23) then
+*FOX  CRABAMP=ED(IX) ;
+           crabfreq=ek(ix)*c1e3
+           crabpht=crabph(ix)
+           
+*FOX  Y(1)=Y(1) - CRABAMP*C1E3/E0F*
```
rdemaria: same comment as before, my understanding is that Y(1) is Px/P and not Px/P0

rdemaria:
More general comment. For a proper test I would make a madx script with a lattice (crab with voltage and lag)+(any fodo), low energy, do some tracking with fairly large initial condition, convert to sixtrack (using the madx sixtrack command), use fort.13 to input the same initial condition and check the results. They should identical to machine precision.

kyrsjo:
* Add temporary variable - add wavenumber for the crabs
* Check parentheses -> @kyrsjo
* Do MadX <-> SixTrack check; put the madx script in SixTest/<name_of_new_test>/MadX -- keep the MadX script AND the output
* Manuals
* Twiki
* Update tests

JamesMolson:
```
Test #15: crabamp .............................***Failed    0.94 sec
Test #26: javier ..............................***Failed    9.01 sec
Test #53: dump4 ...............................***Failed    6.72 sec
Test #54: dump5 ...............................***Failed    5.84 sec
Test #55: dump6 ...............................***Failed    6.16 sec
Test #70: rfMultipoles_2ndOrder ...............***Failed    2.04 sec
Test #72: rfMultipoles_4thOrder ...............***Failed   22.34 sec
Test #71: rfMultipoles_3rdOrder ...............***Failed   24.83 sec
```

Maybe more tests in addition, prob1, prob2, javier_bignpart ?

kyrsjo:
A possibly very annoying comment: If the code inside "+if tilt" is anyway identical to the "+if .not.tilt", maybe you could consider dropping the +if's and maybe just put a comment that "tilt is ignored"? That would at least get the number of times you need to write or copy paste each equation down to two...

rdemaria:
Maybe it is better to implement the tilt. It is actually a big deal if you
look at the conversion madx to sixtrack. In madx track and twiss the tilt
is included, while in sixtrack it is not and in madx conversion the tilt
information it is used as a switch, so we cannot easily model tilted
cavities. In HL-LHC I have to install two elements crabH/crabV to survive.

In alternative let's fix the code as it is to have a reference for the
correct physics, then implement a generic rf-multipole with tilt (as in
madx) which will replace all these functions.

kyrsjo:
Eventually, yes that would be ideal, however maybe for now that should rather go into the issue tracker, to be fixed in a future pull request?

tpersson:
This commit has been tested against MAD-X for a single kick but with different offset in all 6D variables + the phase. I will take some of the automatically generated tests and add to the test suite and also update the documentation.

rdemaria:
It would be interesting to test this new code and the simple tests on the quad precision branch of James. Probably the best is anyway merge the code and the tests in master and then James can try to rebase.

SixTrack/sixtrack.s
```diff
@@ -4005,130 +4015,140 @@
...
+      rvv(j)=(ejv(j)*e0f)/(e0*ejfv(j))
       yv(1,j)=(ejf0v(j)/ejfv(j))*yv(1,j)
       yv(2,j)=(ejf0v(j)/ejfv(j))*yv(2,j)
-      rvv(j)=(ejv(j)*e0f)/(e0*ejfv(j))
+
       if(ithick.eq.1) call envarsv(dpsv,oidpsv,rvv,ekv)
                 enddo
 +cd ccmul3
 ! JBG RF CC Multipoles
+          print *, "test"
```
JamesMolson: Please remove this when you are done

JamesMolson:
So I tested what @rdemaria suggested. rfMultipoles_2ndOrder dropped to ~10^-27 (good).

a number of tests dropped by about 10^-3. Still 10^-13 at best. crabamp didn't seem to move.

rdemaria:
Good thanks!

tpersson:
The tests now runs on my laptop. I think all the things should be in place now.  I have added the scripts I used to benchmark with MAD-X. I have not added anything to the test system since the tests in there are basically doing the same thing and the goal was to have a more complete test where element by element  could be tested for a non-zero orbit. However, this already fails for a normal RF-cavitiy so this need some more investigation. 

So for me I this is fine to merge.

rdemaria:
Can you resolve the merge conflicts?

tpersson:
Maybe there is still away for me to do it but it says that only people with write access to this repository can merge.

tpersson:
Thanks James for the help with the tests. There is a problem with the fma but it doesn't happen every time and I don't think it is related to the changes. It should clearly be investigated but not directly related to this pull request.

SixTrack/sixtrack.s
```diff
@@ -3890,50 +3892,55 @@
           pi=4d0*atan(1d0)
 +ei
         crabfreq=ek(ix)*c1e3
-
+        crabamp=ed(ix)
```
kyrsjo: This variable `crabamp` no longer serves any purpose, why not delete it?

SixTrack/sixtrack.s
```diff
@@ -3890,50 +3892,55 @@
...
+                                 
         do j=1,napx ! loop over particles
-         crabamp=(ed(ix)/ejfv(j))*c1e3                                   !hr03
 
+         kcrab=(((sigmv(j)/(clight*(e0f/e0)))*crabfreq)*2d0)*pi         &!hr03
```
kyrsjo: This is actually the crab phase for a given particle, not the wave number; thus the name is a bit off. I would suggest to pull the unchanging calculation out of the loop, then just keep the sigmv(j)/kcrab+phase inside the loop...

SixTrack/sixtrack.s
kyrsjo: Was this checked?

Also, I would think that x**2-y**2 is not as accurate as (x-y)(x+y), since you are then subtracting or adding slightly less similar-but-large-numbers

SixTrack/sixtrack.s
```diff
@@ -3890,50 +3892,55 @@
...
+     & +  crabph(ix)                                                     !hr03 
 +if .not.tilt
```
kyrsjo: Just nuke the +if tilt / +if .not.tilt already OR implement the tilt; the as long as the code is actually the same, having the `+if` is just misleading...

kyrsjo:
Regarding the problem with the FMA, which part of the test is failing? And which test(s)? In which versions? I think there was a problem with CR of some DUMPs that got fixed in devel; many of the "fma" tests are just as much a test of DUMP.

SixTrack/sixtrack.s
rdemaria: implementing tilt requires also a review on the madx side, unfortunately. I would implement the tilt when we have a generic rf-multipole and review madx conversion at the same time. One can remove the "if tilt" at the sixtrack V merging stage.

rdemaria: I cannot comment the code below for some reason.  I am not sure (x-y)(x+y) is more accurate, since it is 2 sums instead of 1. I checked with mpmath and this is what I have got:

image

Anyway we would need to change consistently everywhere...

SixTrack/sixtrack.s
tpersson: True it is not a real wavenumber but not what I would mean with a phase in this context. It would be some kind of position dependent phase.. 

 Yeah that is sure possible to change.. Is it worth it?

SixTrack/sixtrack.s
tpersson: Sure I can delete it. It was left for almost like a comment for what ed(ix) actually is.

kyrsjo:
Regarding the difference in the energy update - after doing a more proper check, I see that it is the same as for the accelerating cavity, except for the order of the commands.

Also, thanks to Riccardo for checking the accuracy. Seems that was just blinded by "conventional wisdom", so there I learned something new :) I'll have a last look, possibly permutate the two lines for the energy update, then merge.

Thanks a lot!

kyrsjo:
FYI, the `+if .not.tilt` code doesn't compile:
```
/home/kyrsjo/code/SixTrack/SixTrack/SixTrack_cmakesix_BUILD_TESTING_NOTILT_defaultcompiler_defaultbuildtype/track.f:3258:6:

       ejv(j)=ejv(j)-(((((1d0/3d0)*(crabamp3))*(crkve**3-                &!hr13
      1
Error: Unclassifiable statement at (1)
```

Since this code will anyway soon be deleted, I will just merge now instead of holding things up even further.

JamesMolson:
Just merged this into my master and noticed:

```
 SixTest/rfMultipoles_2ndOrder/fort.90.canonical             |   Bin 149168 -> 0 bytes
 SixTest/rfMultipoles_2ndOrder/singletrackfile.dat.canonical |   Bin 4773376 -> 0 bytes
```

This is not good. I guess this test crashed and I copied in empty files as the new canonicals

JamesMolson:
```
 ERROR IN DAINV, INVERSION DID NOT WORK,I,J,PROD =            5           6   1.1444091796875000E-005   9.9999999999999995E-008   9.9999999999999996E-039
 ERROR IN ROUTINE DAINV
   ERR DAINV3
```

tpersson:
Humm.. This is of course not good but it I just remind that the second order RF multipole were not never taken into account in the FOX part before so it might very well be that this never worked.


rdemaria:
Tobias can you reproduce the failing example in march and see if the motion is stable?

tpersson:
I think that was before my time.. I am not sure which example that is..

SixTrack/sixtrack.s
```diff
@@ -20574,173 +20617,214 @@ C Should get me a NaN
...
+*FOX  CRABAMP3=ED(IX) ;
              crabfreq=ek(ix)*c1e3
              crabpht3=crabph3(ix)
-*FOX  Y(1)=Y(1) + 2*(1/2.)*CRABAMP3*((CRKVE*CRKVE)-
-*FOX  (CIKVE*CIKVE))*C1M3*
-*FOX  COS(SIGMDA/C1E3/CLIGHT*CRABFREQ*2D0*PI + CRABPHT3);
-*FOX  Y(2)=Y(2) - 2*CRABAMP3*(CRKVE*CIKVE)*C1M3*
-*FOX  COS(SIGMDA/C1E3/CLIGHT*CRABFREQ*2D0*PI + CRABPHT3);
-*FOX  DPDA1=DPDA1 - 2*(1/6.)*(CRABAMP3)*(CRKVE*CRKVE*CRKVE-
+*FOX  Y(1)=Y(1) + 2D0*(1D0/2D0)*CRABAMP3*((CRKVE*CRKVE)-
```
JamesMolson: 2 * (1/2) ?

SixTrack/sixtrack.s
```diff
@@ -21906,22 +22020,27 @@ C Should get me a NaN
...
-*FOX  Y(1)=Y(1) + 2*(1/2.)*CRABAMP3*((CRKVE*CRKVE)-
-*FOX  (CIKVE*CIKVE))*C1M3*
-*FOX  COS(SIGMDA/C1E3/CLIGHT*CRABFREQ*2D0*PI + CRABPHT3);
-*FOX  Y(2)=Y(2) - 2*CRABAMP3*(CRKVE*CIKVE)*C1M3*
-*FOX  COS(SIGMDA/C1E3/CLIGHT*CRABFREQ*2D0*PI + CRABPHT3);
-*FOX  DPDA1=DPDA1 - 2*(1/6.)*(CRABAMP3)*(CRKVE*CRKVE*CRKVE-
+*FOX  KCRABDA=(SIGMDA/(CLIGHT*(E0F/E0))
+*FOX  *CRABFREQ*2D0*PI + CRABPHT) ;
+
+*FOX  Y(1)=Y(1) + 2*(0.5D0)*CRABAMP3*((CRKVE*CRKVE)-
```
JamesMolson: 2 * 0.5 here as well...

tpersson: Yeah.. could very well be that I missed things like this one some places.. Feel free to fix it or merge it and I can have a look at the same time as I debug the 2nd order RF-multipole.

PR #383, opened 2017-12-01
File dump.s90 reflow and prettified

vkbo:
Now looks like actual Fortran Free Form (TM)
@kyrsjo


kyrsjo:
Thanks! Looks good now :)

I changed the name of the PR as it sounded like fort.90 (which is also kind of a dump).

PR #384, opened 2017-12-05
high precision dump test + root output + misc

JamesMolson:


kyrsjo:
Could you please merge the tests of the high-precision dumps in MASTER (also/instead)? I suspect https://git-scm.com/docs/git-cherry-pick could be useful..

PR #387, opened 2018-01-02
Convert common blocks to modules

kyrsjo:
This patch converts several of the big +cd blocks to modules, namely `parpro`, `common`, `commonmn`, and `commons`. Furthermore it removes `DATAMODS`, as this functionality (`allocate_thickarrays`) is now included in the new modules and always enabled. It also fixes a bug where part of fort.95 would be read twice when datamods is disabled (including once with the wrong format), and not at all when it is enabled.

This is a preparation for making the big arrays become dynamically allocated.


kyrsjo:
Hi, has anyone looked at this?

rdemaria:
If one would have been creative with ad-hoc definitions and astuce blocks, willingly or unwillingly re-scoping variables, would your changes affect the semantic?

JamesMolson:
Collimation does not build with this

JamesMolson:
Everything now seems to be working!

PR #393, opened 2018-01-15
FLUKA coupling fixes from Alessio

JamesMolson:


SixTrack/sixtrack.s
```diff
@@ -24425,69 +24425,8 @@ end subroutine runda
         call prror(79)
       endif
 
-      if(ithick.eq.1) call envarsv(dpsv,oidpsv,rvv,ekv)
```
rdemaria: Just not to forget, why is this removed, is it safe?

amereghe: That line is not actually removed - simply collapsed into one if statement, containing also blocksv, which is a replacemente of:

```
!-------------------------------------  START OF 'BLOCK'
[...] ! replaced with blocksv
!---------------------------------------END OF 'BLOCK'
[...]
 if(ithick.eq.1) then
!------ Compute matrices for linear tracking
    call envarsv(dpsv,oidpsv,rvv,ekv)
     call blocksv
    end if
```

PR #396, opened 2018-01-16
This commit gives the correct name of the rfMultipole tests and removes the crashing test (which crashed for other reasons)

tpersson:
The 2nd order did not contain any 2nd order RF multipoles (only crab cavities.) so it was removed. The problem that it didn't work simply came from a strong 2 strong in phase kicks of the crab cavity. This was probably fine before since then the FOX part was weaker by 10-3.


JamesMolson:
All done and ready to merge?

tpersson:
It should be fine to merge now. I forgot to add the changes to CMakeList.

PR #400, opened 2018-01-26
Mark SCATTERed particles as dirty for collimation

kyrsjo:
Here are the proposed changes to mark the particles as "dirty" for collimation, which should result in them being written to tracks2.dat etc.

I added one more parameter to scatter_thin (the structure element ID); it may be a good idea to also write this and/or the element s-position (`dcum` array AFAIK) to the SCATTER output file.


kyrsjo:
@JamesMolson  Did I forget something?

@vkbo  Does it work?

PR #407, opened 2018-02-01
Fix master building with CRLIBM disabled!

JamesMolson:


kyrsjo:
So this fixed collimation version. Did it also fix the Ubuntu crash problem? Is fort.6 still coming out OK in the CR tests?

PR #408, opened 2018-02-15
Increase number of characters in bez from 16 to 49

kyrsjo:


kyrsjo:
Still hardcoded stuff in the ROOT C(++) interface.

PR #418, opened 2018-03-21
Heavy ion tracking support

JamesMolson:


JamesMolson:
Ok, so everything except the tests for the elens work (as expected)

PR #421, opened 2018-03-21
New module: end_sixtrack

vkbo:
Moved subroutine prror, abend into module end_sixtrack.


kyrsjo:
Great! Would it make sense to put `abend` in the same module as `prror`?

Also, the common block `errout` definitively belongs in there:
```
+cd errout
!     Set the exit status in case of an error.
      integer errout_status
      common /errout/ errout_status
```

vkbo:
Ah, yeah, you mentioned another subroutine. I couldn't remember which one. I'll have a look at it.

Yes, I'll move the two variables as well.

PR #445, opened 2018-04-13
HDF5 and String Tools modules

vkbo:
This pull request adds a QUIET flag to fort.3 that disables the output of particle initial coordinates, and reduces the final particle output to one line per pair.

This PR also removed the old HDF5 module, and contains a starting point for a new module. This will be properly added in a separate PR, as of now it doesn't do anything.

There's also a change in CMake to use 5 digit version numbers for the executable, and added a SPAMMY flag for extra spammy debug output like the memory allocation info.

A new list for f90 files have been added to CMakeLists for all the files that do not require ASTUCE. the c pre-processor has also been enabled.


rdemaria:
Nothing against! James, any comments?

vkbo:
I reverted the changes to mod_alloc as it required more work than I could be bothered with right now. I cannot use the quiet flag as it generates a circular dependency.

James: Please fix your alloc spam!

vkbo:
Ok, let's hold off on merging this for now. I just started changing stringzerotrim and getfields_split. Just moved it to a module, but need to make some changes as well.

SixTrack/string_tools.f90
```diff
@@ -0,0 +1,371 @@
...
+  newLen = iB - iA + 1
+  
+  if(newLen <= 0) then
+    retString = string("")
+    return
+  end if
+  if(iA < 1)      iA = 1
+  if(iB > oldLen) iB = oldLen
+  
+  allocate(character(newLen) :: tmpString)
```
JamesMolson: Could you use the alloc() call and dealloc this before returning.

JamesMolson:
Could you use something like "verbose" instead of "spammy"

vkbo:
Added a deallocate()
SPAMMY is just a temporary flag. You can remove it when you remove your SPAM ... spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam spam sausages and spam.

Issue #443, opened 2018-04-09
Different proton mass in the wire


tpersson:
It looks to me like there are two proton masses. One called pma which is the input from the BEAM BLOC but there is also a pmpa which  looks some default value. In the WIRE map the pmpa is used (most other places pma is used. Is there a reason for this?

rdemaria:
Even if there was a reason,  would it be possible to use only pma?

tpersson:
As I see it we could use only pma, or rather the new nucm(j) which in the case of protons become pma.

amereghe:
I guess that, if pma is used to compute the kick given to a particle, let's go with mucm(j). Otherwise, if pma is used to compute some parameters, the value of which is referenced to protons, then we should maybe set a more proper interface...

JamesMolson:
I also push it into nucm(j)

There is also a value of the particle mass in the SYNC block.

It can also be set in the HION block, but if that does not exist I set it to default to whatever the default proton value is.

At some point I think there was a check if your set particle mass was within some tolerance of the default proton (or electron) value and if not, it would STOP.

rdemaria:
I would keep the value of proton mass in the BEAM command of fort.3. That
is:
- replace any pmap with pma
- in daten copy the value of pma to mucm(j) and mucm0 if no HION is found
- never use pma/pmap and use always mucm(j) and mucm0

tpersson:
Sure, I think that sounds reasonable.

There is a defined value in the common_modules.s90 and it is different depending on if fluka is used or not..


+if .not.fluka
  ! proton mass (MeV)
  real(kind=fPrec), parameter :: pmap = 938.271998_fPrec   !old
  ! real(kind=fPrec), parameter :: pmap = 938.2720813_fPrec  !2017

  ! electron mass (MeV) from PDG, 2002
  real(kind=fPrec), parameter :: pmae = 0.510998902_fPrec   !old
  ! real(kind=fPrec), parameter :: pmae = 0.5109989461_fPrec  !2017
+ei
+if fluka
!     A.Mereghetti and D.Sinuela Pastor, for the FLUKA Team
!     last modified: 08-12-2014
!     synch masses of proton and electron to values used by FLUKA
!     inserted in main code by the 'fluka' compilation flag
!     real(kind=fPrec), parameter :: pmap = 0.938272046e3_fPrec ! PDG 2014, Fluka-2011-dev
[...]

PR #447, opened 2018-04-18
Ions implemented for FOX

tpersson:
Ions implemented in the FOX part 
Fixes the proton mass in the wire that was not the same as in the rest of the code #443 
Initialize all the ions to the default ion chosen in the ion block. 

Numerical differences in many canoncial. They have NOT been updated. It would be good if someone else could have a look and see if this is just due to the fact that now the variable (mtc/(1+delta)) is the variable that is mainly used compared to delta before.


JamesMolson:
i.e. oidpsv goes to moidpsv

Also could you merge in dev again - the merge conflict is just a "use mod_hions" so easy to fix.

SixTrack/sixtrack.s90
```diff
@@ -10750,8 +10764,8 @@ subroutine daten
       iskew=0
       preda=c1m38
 !hion
-      zz0 = 1
-      aa0 = 1
+      zz0 = one
```
JamesMolson: these are integers, so using "1" is totally ok. "one" is a floating point variable and there will be a cast taking place with this.

rdemaria:
I think a way to test the DA part is to look at the tunes written in fort.6
after the closed orbit search. Normally fort.6 is not checked by the test
suits, but probably the tune in fort.10 are the one from this routine. Can
you check that tunes did not change?

tpersson:
Yeah that is perfectly fine. This how it was done for the tracking part. So to be consisten between the rwo implementations, or at least more consistent, I continued the same way.

Cheers
Tobias






-------- Original message --------

tpersson:
I will just have to fix the ion tests and then this can be merged.

tpersson:
All the tests should now pass and this fixes both the implementation of the ions and the proton mass in the wire #443. 

Should be possible to merge this now.

PR #454, opened 2018-04-26
Update to use the latest BOINC API version.

JamesMolson:


JamesMolson:
Seems to be broken on a mac, but its a simple fix I'll submit upstream.

PR #464, opened 2018-05-04
Move collimation variables into the collimation file and module.

JamesMolson:


JamesMolson:
This version went a bit wrong, and was doing in several other pull requests instead.

PR #472, opened 2018-05-21
Fix DYNK GLOBALVARS energy update to work with THICK and SIGMAV

kyrsjo:
Changes that I think should fix the remaining bugs in ENVARSV, after carefully reading the energy update code in `thck6dua` and `thin6dua`.

Note: In the old code, the algorithm was as follows:
* Store original e0 and e0f
* Update e0 and e0f
* For each particle:
  * if (sigma < 0): sigma = (beta0 / beta0_old) * sigma
  * Energy kick from the cavity && update of variables
  * if (sigma > 0): sigma = (beta0 / beta0_old) * sigma
* In thick only: Call `synuthck`

I'm a bit confused about why it does the sigma update the way it is done...

Things still to be done:
* Delete the now superfluous and commented-out code from `sixtrack.s`
* Delete variables `phas0`, `e0o`, `e0fo`, `xv1j`, `xv2j`
* Make variables `nde(1)` and `nde(2)` -- the number of turns at flat bottom and for energy ramping -- invalid in `daten`
* Review `nwr(1)`, `nwr(2)`, `nwr(3)`, `nwr(4)` in for marking invalid.
* Review and delete calls to the old code...

Once this tiny bit of diff is reviewed, I'll do the rest of the changes and update the test. Then we can merge.

-- Kyrre & Veronica

### DIFF
--- a/SixTrack/cmake_six
+++ b/SixTrack/cmake_six
@@ -38,16 +38,16 @@ if [ -x "$(command -v gawk)" ]; then
     fi
     }
     astuce_linter sixtrack.s90
-#    astuce_linter plato_seq.s90
+    # astuce_linter plato_seq.s90
     astuce_linter lielib.s90
     astuce_linter dabnew.s90
     astuce_linter beamgas.s90
-#    astuce_linter bouncy_castle.s90
+    # astuce_linter bouncy_castle.s90
     astuce_linter collimation.s90
     astuce_linter dynk.s90
     astuce_linter fluka.s90
     astuce_linter fma.s90
-    astuce_linter hdf5_sixtrack.s90
+    # astuce_linter hdf5_sixtrack.s90
     astuce_linter postprocessing.s90
     astuce_linter scatter.s90
     astuce_linter dump.s90
--- a/SixTrack/dynk.s90
+++ b/SixTrack/dynk.s90
@@ -3024,6 +3024,9 @@ subroutine dynk_setvalue(element_name, att_name, newValue)
     character(maxstrlen_dynk) element_name_stripped
     character(maxstrlen_dynk) att_name_stripped
 
+    !Original energies before energy update
+    real(kind=fPrec) e0fo, e0o
+
     ! For sanity check
     logical ldoubleElement
     ldoubleElement = .false.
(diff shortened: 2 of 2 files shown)


rdemaria:
The sign of sigma is a bit puzzling... when is beta0_new calculated? In principle, updating sigma can be done after or before the cavity kicks and the  renormalization of the other variables, although logically should be done at when updating delta, rvv and so on...

kyrsjo:
Yeah, it's a bit puzzling. As seen in the PR diff, I elected to do it when updating delta, rvv, etc.

Here is the actual old code for the thick part:
```
    40       e0o=e0
             e0fo=e0f
             call adia(n,e0f)

             do j=1,napx
               ejf0v(j)=ejfv(j)
               if(abs(dppoff).gt.pieni) then 
                 sigmv(j)=sigmv(j)-sigmoff(i)
               endif
               if(sigmv(j).lt.zero) sigmv(j)=((e0f*e0o)/(e0fo*e0))*sigmv(j) !hr01
               if(kz(ix).eq.12) then
                 ejv(j)=ejv(j)+(ed(ix)*sin_mb((hsyc(ix)*sigmv(j)+phas)+phasc(ix)))*nzz(j)
               else
                 ejv(j)=ejv(j)+(hsy(1)*sin_mb(hsy(3)*sigmv(j)+phas))*nzz(j)
               endif
               ejfv(j)=sqrt(ejv(j)**2-nucm(j)**2)                             !hr01
               rvv(j)=(ejv(j)*e0f)/(e0*ejfv(j))
               dpsv(j)=(ejfv(j)-e0f)/e0f
               oidpsv(j)=one/(one+dpsv(j))
               moidpsv(j)=mtc(j)/(one+dpsv(j))
               omoidpsv(j)=c1e3*((one-mtc(j))*oidpsv(j))
               dpsv1(j)=(dpsv(j)*c1e3)*oidpsv(j)                          !hr01
               if(sigmv(j).gt.zero) sigmv(j)=((e0f*e0o)/(e0fo*e0))*sigmv(j) !hr01

               yv(1,j)=(ejf0v(j)/ejfv(j))*yv(1,j)                         !hr01
               yv(2,j)=(ejf0v(j)/ejfv(j))*yv(2,j)                         !hr01
             end do

             call synuthck
```

rdemaria:
For me it is ok then. Do you have a simple example that we can use to benchmark the code with theory?  For instance implementing an acceleration on a simple linear lattice and measuring the bucket size with a bunch of particles that spans the longitudinal phase space?

PR #473, opened 2018-05-23
Sixtrack input module

vkbo:
Note: Hold off on merge. Need to run some more tests.

This PR adds the module sixtrack_input.
This module contains:
* Line-by-line parsing for Single Elements
* Line-by-line parsing for Block Definitions (including 5-space indent for line continuations)
* Line-by-line parsing for Structure Input (including brackets expansion)
* Line-by-line parsing for Displacement of Elements

The main fort.2 and fort.3 parsing loop in DATEN (goto labels 100 and 110) have been replaced by two select case regions. The first one will parse input block with the current logic, i.e. goto labels to regions with internal loops over lines, with a goto 110 at NEXT.

If the check falls through this select case, a new select case picks it up where line by line parsing is performed instead. A number of the newer modules like ZIPF, HION, ROOT, HDF5 and SCATTER have such parsing code, and have been moved here. So has the new implementation of SING, BLOC, STRU and DISP.

Currently all fast and medium tests that otherwise work, passes. A new test named geometry has been added to test the bracket expansion of structure input. This was not covered by any tests. The line continuation of block definitions is already covered, and the displacement block is covered as well by one test.

Dynamic allocation of block arrays (length of nblo) has been added, but currently defaults to the old value of 2500 as expansion breaks the logic of offsetting indices in structure input array. This will be updated soon by using negative indices for blocks instead of offset.

In string_tools a few new features have been added:
* A split character array function that will dynamically rescale to contain the data as opposed to the getfields method which is fixed to 160 chars by 15 elements.
* There are two routines str_toReal and chr_toReal that will handle strings and character arrays respectively and cast them to real(kind=fPrec) incorporating the 4 possible combinations of the FIO and CRLIBM flags.
* Two new routines str_toInt and chr_toInt that does a simple read() call. Just added for completeness, and is a bit simpler to use.
* A chr_expandBrackets routine that will expand notation like "foo 2(bar)" to "foo bar bar". This is used for structure inputs.


vkbo:
Had a discussion with @rdemaria, and we should probably hold off on this PR until RC2 is released.

vkbo:
Added sixtrack_input parsing for INIT and TRAC blocks as well.
This triggered the need to move several common block to common_modules.f90. Two new ones were made, mod_commond which contains commondl, and mod_commont which contains commont1, commont2, commontr and commonta. In addition commonm1 was moved t  mod_commonmn and  commonex to mod_common.

Also fixed a number of new bugs that affected DA, and made sure the conflict between variable names in commont1 and DA routines were resolved. Previously, commont2 was a copy of commont1 with different variable names as three of these were in conflict with other variables used in DA version, especially in FOX. These are now handled by a copy routine that makes sure the double set of variables are in sync. They were small enough that this is an option. Two 2x6 matrices and a len 2 vector.

rdemaria:
The original names of the common blocks are not particularly meaningful. By looking at the code, do you think you can assign new names that better reflects the content?

vkbo:
Sure, we could easily do that I think. If we can come up with actual meaningful names!

JamesMolson:
This stop collimation from building here

PR #475, opened 2018-05-24
End of UA

vkbo:
Removed thin6dua and thick6dua


amereghe:
Was the user manual updated on how to simulate a ramp, then?

PR #564, opened 2018-07-13
Dump output rounding and cleanup

vkbo:
This PR replaces all output writing in DUMP with calls to the roundctl library when CRLIBM is on.
In the same PR write(lout,*) has also been replaced.
In addition, the input parser has been updated to use the new string split routine.

There is also a minor bug fix in aperture in the middle of this PR. It's a change in the header and format of the aperture_losses.dat file.


source/dump.f90
```diff
@@ -429,9 +426,10 @@ subroutine dump_initialise
 
   implicit none
 
-  integer i,j
-  logical lopen
+  integer i,j,k,l
+  logical lOpen, rErr
   character(len=256) filename
+  character(len=16) tasbuf(6,6)
```
kyrsjo: Can you guarantee that len=16 is enough?

source/dump.f90
```diff
@@ -1479,60 +1668,19 @@ subroutine dump_beam_population(nturn, i, ix, unit, fmt, lhighprec, loc_clo, tas
 
   ! Unrecognized format fmt
   else
-    write (lout,*) "DUMP> Format",fmt, "not understood for unit", unit
+    write (lout,"(2(a,i0))") "DUMP> ERROR Format ",fmt," not understood for unit ",unit
```
kyrsjo: Here you may also want to say which filename, since the unit may no longer be known by the user.

source/dump.f90
```diff
@@ -1479,60 +1668,19 @@ subroutine dump_beam_population(nturn, i, ix, unit, fmt, lhighprec, loc_clo, tas
...
-1989 format (2(1x,I8),1X,F12.5,6(1X,1PE25.18))        !fmt 4 / hiprec
-1990 format (2(1x,I8),1X,F12.5,6(1X,1PE16.9))         !fmt 4 / not hiprec
-
-1991 format (2(1x,I8),1X,F12.5,27(1X,1PE25.18))       !fmt 5&6 / hiprec
-1992 format (2(1x,I8),1X,F12.5,27(1X,1PE16.9))        !fmt 5&6 / not hiprec
-
 end subroutine dump_beam_population
 
 ! ================================================================================================================================ !
 subroutine dump_comnul
```
kyrsjo: Have these array initializations gone to the array create/expand code?

source/string_tools.f90
```diff
@@ -398,24 +398,24 @@ function chr_strip(theString) result(retString)
 end function chr_strip
 
 ! ================================================================================================ !
-!  Pad String with Zeros or Spaces
+!  Pad String Spaces
 !  V.K. Berglyd Olsen, BE-ABP-HSS
 !  Last modified: 2018-05-18
 ! ================================================================================================ !
-function chr_padZero(theString, theSize) result(retString)
+function chr_lpad(theString, theSize) result(retString)
```
kyrsjo: Sure you don't want to import leftpad from somewhere in the cloud? :P

kyrsjo:
Looks good to me. I would just add the file name output in that one error message; then do whatever is currently equivalent to bumping the version number and merge!

source/dump.f90
vkbo: Yes. That is determined by the call to the casting routine.

source/dump.f90
vkbo: Yes.

source/dump.f90
vkbo: Done.

PR #580, opened 2018-07-19
Combined function

tpersson:
Only includes quadrupolar terms.


tpersson:
Don't merge yet. Tests needs to be fixed.

tpersson:
yes! 
Thanks!
________________________________________

vkbo:
Comments:
* There were several bugs in the LaTeX. Please check that the manual builds!
* Most places in SixTrack the comparison with pieni is >= not >, but I noticed it was previously done differently with these variables, so I reverted my initial changes. Note: the newver standards use >, <, == syntax like in other languages.
* Added some parentheses to the new calculations.
* Removed some additional spaces and end of lines, and added line endings at end of file. I think you can set your editor to do this automatically?

vkbo:
Could you also add this to the top of the changelog Under the RC4 header?

PR #581, opened 2018-07-19
Custom input read and output rounding

vkbo:
This PR removes the rest of the calls to old splitfld routine and also replaces the rest of dtostr calls with the new wrapper routine chr_fromReal (roundctl wrapper).

Reading of fort.33 is moved to a separate module do avoid duplication of code.
Also cleans out all the now unused routines related to rounding and splitfld.


vkbo:
@rdemaria Check is still pending on this. Bug?

PR #587, opened 2018-07-23
Build system updates.

JamesMolson:


JamesMolson:
hmm, this will pass the lxplus option as is to cmake, which doesn't work!

PR #593, opened 2018-08-06
regenerate input in build directory if source changed

rdemaria:


vkbo:
This PR is incomplete and should not have been merged yet.

rdemaria:
Sorry, my bad. Do you refer to  eliminate run_prepro?

vkbo:
No, to that it ignores changes to include files. Trying to fix it now ...

vkbo:
Since CMake does not resolve dependencies well here, I don't think there is any other way than to either:
1. Make all files depend on all includes, which both slows down the build process and results in a full rebuild anyway when an included file is change, or
2. Manually or with a script resolve these dependencies and add them to the relevant files.

rdemaria:
I would try with a  script to resolve dependencies, e.g. 

```
awk  '/#include/ {print $2}' track_thin.f90 | sort -u
```

Can we use awk and sort also in msys, right?

PR #595, opened 2018-08-06
More cleanup in CMake

vkbo:
This does some further cleanup of CMakeLists from #593 
The PR did not include the include files as dependencies, however, that seems to be a difficult thing to achieve as CMake does not resolve the dependencies itself, so it needs to be added in another way.
As it stands, if include files are changed, a full rebuild is required.
Still adding this PR as some redundant stuff has been cleaned out.


rdemaria:
James can you review this? or I can do myself...

PR #596, opened 2018-08-06
Add format 101

rdemaria:


source/dump.f90
```diff
@@ -1666,7 +1669,63 @@ subroutine dump_beam_population(nturn, i, ix, unit, fmt, lhighprec, loc_clo, tas
...
+      call h5_writeData(dump_hdf5DataSet(ix), 12, napx, ejfv(j))
+      call h5_writeData(dump_hdf5DataSet(ix), 13, napx, mdpsv(j))
+      call h5_writeData(dump_hdf5DataSet(ix), 14, napx, moidpsv(j))
+      call h5_writeData(dump_hdf5DataSet(ix), 15, napx, rvv(j))
+      call h5_writeData(dump_hdf5DataSet(ix), 16, napx, nucm(j))
+      call h5_writeData(dump_hdf5DataSet(ix), 17, napx, mtc(j))
+      call h5_writeData(dump_hdf5DataSet(ix), 18, napx, e0)
+      call h5_writeData(dump_hdf5DataSet(ix), 19, napx, e0f)
+      call h5_finaliseWrite(dump_hdf5DataSet(ix))
+    else
```
vkbo: This will definitely not work. You have to define the dataset format for the HDF5 output in `dump_initialise` and add all the column names there. The data set ID is stored in the `dump_hdf5DataSet` array in a similar way to filenames for regular file dumps.

source/dump.f90
```diff
@@ -1666,7 +1669,63 @@ subroutine dump_beam_population(nturn, i, ix, unit, fmt, lhighprec, loc_clo, tas
 #endif
     end if
 
+  ! ------------------------------------------------------------------ !
+  !  Format #101                                                       !
+  !  Same as fmt 3, but with additional variable                       !
+  ! ------------------------------------------------------------------ !
+  else if(fmt == 101) then
```
vkbo: Since all the existent formats are in numerical order, why not just use the next number in the line, 10, instead of jumping to 101?

doc/user_manual/chExtraOutput.tex
```diff
@@ -65,15 +93,15 @@ \section{Dumping of Beam Population} \label{sec:DUMP}
...
         \hline
-        Lines  & \texttt{turn structure\_element\_idx single\_element\_idx single\_element\_name s x1[m] x1'[rad] y1[m] y2'[rad] momentum[GeV/c] dE/E[GeV]} \\
+	    Lines  & \texttt{turn structure\_element\_idx single\_element\_idx single\_element\_name s x[m] xp[rad] y[m] yp[rad] P[GeV/c] (E-E0)[eV] t[s]}\\
         \hline
 
         \rowcolor{blue!15}
         \texttt{1} & Format for aperture check \\
         \hline
-        Header & \texttt{\# ID turn s[m] x[mm] xp[mrad] y[mm] yp[mrad] dE/E ktrack} \\
+	    Header & \texttt{\# ID turn s[m] x[mm] xp[mrad] y[mm] yp[mrad] dE/E ktrack} \\
```
kyrsjo: TAB character?

doc/user_manual/chExtraOutput.tex
```diff
@@ -37,6 +40,31 @@ \section{Dumping of Beam Population} \label{sec:DUMP}
     \texttt{FRONT}         & keyword & If present anywhere in the \texttt{DUMP} block, this keyword triggers the DUMPed particles to be dumped in front of the element, i.e.\ before the kick. This works for all elements, including \texttt{BLOC}s, when combined with the \texttt{ALL} as \texttt{element\ name}. Note that \texttt{FRONT} is not yet supported for thick tracking, and trying to use this combination will produce a run-time error.
 \end{longtabu}
 
+\paragraph{Conventions}~
+
+The following table show a summary of the quantities used in describing the output format. The units are declared explicitly for each dump. 
+
+\bigskip
+\begin{longtabu}{@{}llllp{0.65\linewidth}}
```
kyrsjo: Thanks, this is very useful!

doc/user_manual/chExtraOutput.tex
rdemaria: just spaces for indentation. I am now going re-indent and split long lines.  I believe long cell in the table should have they own lines and tables should be indented...

rdemaria: I would enforce 80 lines limit in latex source, since we concluded that for reading text 80 lines is more readable ;-)

vkbo: Please don't do that. I spent a lot of time reformatting the LaTeX source to conform to the standard @kyrsjo and others started to implement.
That is: a line break after each sentence.
Any decent editor has the option to wrap long lines, no? That solves the text readability.

rdemaria:
Does readDump3.f90 work also for format 7 and 8?

rdemaria:
wrapping line breaks indentation, for tables is a bit of a nightmare... but
we can discuss...

rdemaria:
Does somebody know whether the normalized coordinates normalize (sigma,psigma) or (z,delta)? The original documentation indicated (z,dp/p), knowing that z was often misuses for sigma, dp/p is not particularly sound and sigma, dp/p cannot be normalized using symplectic matrices because the are not conjugate pairs...

doc/user_manual/chExtraOutput.tex
vkbo: > wrapping line breaks indentation, for tables is a bit of a nightmare... but we can discuss...

Not really? My editor keeps indentation on wrapped lines, and so can vim as of about 4 years ago.

rdemaria: Thanks for the hint! Very useful indeed. This does not solve the issue of the columns that happen at any random place in the line, but at least is more readable.

kyrsjo:
We decided to update the other headers as well, right? I guess that's what is currently outstanding.

rdemaria:
@amereghe , @JamesMolson, is it  ok for you to change the headers of the dump files to match the definitions?

kyrsjo:
Alessio says yes!

kyrsjo:
By the way, the same problem may be present in FMA's NORM files and the corresponding documentation :/ I don't think these are really needed anymore (FMA used to have its own normalization routines which we could probably delete now @mfittere ).

kyrsjo:
There is also the same problem in the column names for HDF5 @vkbo

vkbo:
I know. I'll fix it in a separate PR as I need to make some changes to the HDF5 implementation.

kyrsjo:
Another problem: For formats 7/8/9, we pull (E-E0)/E0 through TAS to normalize. Manual calls it `ndp/p[1e-3 sqrt(m)]` or `psigma`. Is this really correct?

kyrsjo:
OK, I'm done. @rdemaria , please take a look. Question about what the normalization matrix is really doing should probably be answered (currently it is being applied to \delta E / E_0), however I think we can merge before fixing this.

kyrsjo:
Hmm, now that test is always passing? Must have been a strange fluke. The strange thing is that the real error message is missing from fort.93. Also the error message should probably contain the file name, not only the unit number.

PR #598, opened 2018-08-07
User manual build test

vkbo:
This PR adds a test for the user manual build.
I updated the makefile as well.
The test runs in the user manual folder directly, but I don't think that is a problem.
I assume the circleci test will fail as latex is not installed there.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,8 +5,8 @@ jobs:
       - image: debian:unstable
     steps:
       - run:
-         name: setup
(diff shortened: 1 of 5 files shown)


rdemaria:
Can you test a docker image starting from debian/unstable that contains the minimum set of tex packages?

vkbo:
It's not the latex bit that's failing, it's the SixTestWrapper.

vkbo:
I give up. I don't know why this is failing and I don't know how to test it. Everything works just fine locally.

PR #601, opened 2018-08-07
Collimat flag remove

kyrsjo:
Removing the COLLIMAT compilation flag.

NOTE: In progress, do not merge yet.


source/track_thin.f90
```diff
@@ -1159,7 +1152,7 @@ subroutine thin4d(nthinerr)
 
 630 continue
 
-#if defined(ROOT) && !defined(COLLIMAT)
```
kyrsjo: WTF? @JamesMolson

source/track_thin.f90
```diff
@@ -2154,11 +2143,11 @@ subroutine thin6d(nthinerr)
...
 650 continue !END loop over structure elements
 
-#ifdef COLLIMAT
-    call collimate_end_turn
-#endif
+    if (do_coll) then
+      call collimate_end_turn
+    endif
 
-#if defined(ROOT) && !defined(COLLIMAT)
```
kyrsjo: Huh? @JamesMolson

kyrsjo:
Done! Seems to work perfectly. Veronica also added 1 very relevant new test (PR #603).

kyrsjo:
Thanks! It would be great if @JamesMolson could take a look at the ROOT / COLLIMAT build flag stuff that I tagged; the logic of the flag combined with the logic inside the IFDEF looked strange.

JamesMolson:
@kyrsjo The logic is totally fine

test/CMakeLists.txt
```diff
@@ -174,14 +173,14 @@ endif()
 
 #Collimation tests (needs backtrk)
 #use extra checks on fort.999 and other collimation loss files
-if(COLLIMAT AND BACKTRK)
+if(BACKTRK)
```
JamesMolson: this flag does not exist

source/aperture.f90
```diff
@@ -1021,20 +1006,20 @@ subroutine lostpart(turn, i, ix, llost, nthinerr)
           lparID = .false.
           jjx=1
 
-#ifdef COLLIMAT
-          do jj=1,npart
```
JamesMolson: @amereghe This bit

amereghe: https://github.com/SixTrack/SixTrack/blame/master/source/aperture.f90

amereghe:
to avoid mistakes, let's block run if post-pr is there together with FLUKA / COLLIMAT / SCATTER blocks, to avoid inconsistencies in `fort.10`

source/mainarrays.f90
```diff
@@ -56,9 +54,7 @@ subroutine allocate_arrays
 #ifdef CR
   call cr_expand_arrays(npart)
 #endif
-#ifdef COLLIMAT
   call collimation_allocate_arrays
-#endif
```
vkbo: Hmm, this is not a good idea. Collimation allocates a lot of arrays. Should probably find a way around this.

kyrsjo:
Done! Please merge.

source/mainarrays.f90
vkbo: @kyrsjo There's still this ...

vkbo:
Not yet! See my comment above.

source/mainarrays.f90
kyrsjo: Could we call that from inside the COLL block input parsing maybe?

vkbo: yeah, we could probably call the alloc routine with sizes 1 (if unallocated causes stuff to explode) and then call expand after a coll block has been seen. I.e. in daten when the openBlock flag is true.

source/aperture.f90
kyrsjo: I'll let you sort this out in a different PR :)

PR #606, opened 2018-08-09
Multiple bug fixes

vkbo:
This PR started with an attempt to fix the underflow error in subroutine damult in module dabnew.
This was brought on by the fact that building with build type Debug caused **all** tests to fail.
Most of those failures are addressed here.

**Underflow**

This error is fixed by checking that the multiplication can be done by summing the logs of the numbers first.
This is done with a function added in core_tools.f90.
The fix solves the issue in dabnew, but in some of the tests the underflow happens in the call to exp_mb in mathlib_bouncer  instead.

**Non-existent elements in MULT block**

It turns out there were never a check in place for the existence of the single element specified in the MULT block causing the coefficients to be written to index 0 in the storage array, which does not exist. A non-existent element now causes an error.

**Out of bounds on particle array in track_thin**

An extra include statement outside the particle loop in one spot caused an ttempt to set values  in yv(napx+1,:), which does not exist. Fixed.

**BDEX**

In #601 the bdex call in tracking was commented back in. This needs to be looked into as it tries to access a non-existent array location. It is cmmented back out again.

Edit: This turned out to be an attempt to look up a BLOC index, which after nblo offset gives a negative `ix`. There is now an additional if-test for this. BDEX call is commented back in.

**Collimation array allocation**

Also changed in #601, but is implemented incorrectly. Changes reversed, and needs to be looked into again in a separate PR.

**Combined function updates with index out of bounds**

In #580. the dki array was written to using structure element indices, but this is a single element indexed array. This caused ~12 tests to fail with bounds checks enabled. The index has been changed from i to ix, and seems to work again.

**FMA/boncy bouncy castle**

Replaced the atan call in FMA with atan2, and changed bouncy_castle to return a quiet NaN in the case of atan2(0,0). This also required the canonicals for dump7_fma and dump8_fma to change because the angles shifted by one pi. In addition, there was a difference in the last digit in a few places.


vkbo:
The BDEX bit doesn't fail a test directly, but the code when present tries to access and index in one of the arrays that does not exist. This fails on gfortran Debug builds due to the `-fcheck=all` flag

vkbo:
BDEX is now fixed. It was missing a check on the `ix` variable if it was negative due to originating for a BLOC element.

The collimation array expansion has also been updated. The `collimation_allocate_arrays` just calls `collimation_expand_arrays` which now has a `if(.not. do_call) return` line after the array that always needs to be allocated. To make sure the coll arrays are expanded correctly and therefore not dependant on the order in the input file, an additional call to `collimation_expand_arrays` is added in the `collimate_postInput` routine which is run just after `ENDE`.

PR #607, opened 2018-08-10
Trombone fix

jodander:
This is a bugfix for the 'Trombone' (or 'Matrix') element in SixTrack. The actual fix is only the switching of two indices which were in the wrong order. In addition:

1. A test case (orbit6d-element-trombone-kick) was added to test that 'kicks' from the Trombone element are handled properly. This test case is flagged as not having a fort.10, being 'fast' (runs in < 1 s) and being run by default in test/CMakeLists.txt.
2. The orbit6d-element-trombone test case had to be updated.

NOTE: When I ran the full test suite over night, it failed the test cases prob1 and prob3. As neither one involves any Trombone element it should not be caused by my commits but it might be worth mentioning nevertheless.


vkbo:
Looks like the Sixin.zip file is missing for the test.

vkbo:
I don't think the canonicals for the prob1 and prob3 tests have been updated, so they're expected to fail. All fast and medium tests should pass.

tpersson:
I think there might be an issue with non-matching sixin.zip. Could you have a look?

jodander:
I think I forgot to update the sixin.zip in the orbit6d-element-trombone whilst updating fort.3, gonna double-check and fix if that is the case.

kyrsjo:
When were the prob1/prob3 tests broken? It would be good to track this...

rdemaria:
http://abp-cdash.web.cern.ch/abp-cdash/testSummary.php?project=15&name=prob1&date=2018-07-12
http://abp-cdash.web.cern.ch/abp-cdash/testSummary.php?project=15&name=prob1&date=2018-07-13

kyrsjo:
I will run the tests, all the test except elens, overnight. With and without CR.

PR #608, opened 2018-08-10
New DYNK tests

vkbo:
This PR adds three new DYNK tests. One tests all the single and double value operators, onechecks the random number generators, and one checks the polynomials etc.

Tests use SCATTER's scaling variable to test the result. Canonicals are dynksets.dat and scatter_log.dat.

Note: This PR also adds the option in the test/CMakeList.txt to have tests with no STF/fort.90 file. Like in these two tests, these are not needed. This saves space in the repository.

Also added here is a new function abs(x).

Still not tested: PELP; filters, ONOFF, PIPE.


rdemaria:
Ok to merge?

vkbo:
Yes, it's ok to merge.

PR #609, opened 2018-08-11
Dynk Updates - Part II

vkbo:
This PR finishes the rewrite of parseFun and some of the related routines.
This should be merged after #608 as that branch was merged into this due to the new tests added there.

~~There's still a Part III to come that finishes the rest, but that is mostly cleanup and error message formatting. This one completes all the important stuff.~~

Edit: The rest of the module has now been updated too, so this is the end of this.

PR #611, opened 2018-08-12
Remove getfields split and zero terminated strings

vkbo:
This PR removes the old split routine, and must therefore be merged after #609 as DYNK usage of this routine is removed there.

This also removes the last zero padded strings and therefore also stringzerotrim.
All calls to stringzerotrim or chr_trimZero have been converted to plain trim().
The chr_trimZero routine is left in the code in case it is needed for cleaning strings returned from c.


vkbo:
This does not change the logic of DYNK, that is in a different PR.

kyrsjo:
Sorry, my bad - I meant reading of data inside DYNK, which FIR/IIR is doing a lot of. That's PR #618  - once that is merged into master and master is merged into this, if it still works, this branch can be merged. AFAIK the input reading was basically the same in FIR and IIR.

PR #615, opened 2018-08-14
Particles Module

vkbo:
Added new module mod_particles and moved the reference energy update calculation there from DYNK.
Also considering moving some of the particle arrays here for this PR. Need to discuss.

~~**Don't merge yet!**~~

Done. Ready for merge.


kyrsjo:
Could probably call the energy update routine from the rf and crab cavities also!

PR #616, opened 2018-08-16
temporary fixes to E lenses

amereghe:
This is not the actual PR for restoring eLenses in master - hence please **do not merge**
I need it just to keep track of a couple of pieces of info - and to tell everyone that I've starting working again on this front.

```SixTrack/utils.s90``` will disappear. Its content is already somewhere in the main source code of SixTrack, ie:
   * ```invert_tas``` is in ```source/six_fox.f90``` as is;
   * ```lininterp``` is in ```source/dynk.f90``` -- but I will need to extract it and put it in a middle module, common to both eLenses and DYNK.

Checking other left-overs left behind...

### CODE REVIEW

**Comment by vkbo** [comment]
Hmm, one of the commits in this PR deletes lib/librachnive. That has to be reverted for this to be able to merge.

**Comment by vkbo** [comment]
It may just be the head of the submodule that changed, which can be fixed. I'm not at work today, but @kyrsjo may be able to fix it. We just did the same on master last week. The submodule HEAD needs to point to a specific branch in the libarchive submodule.

**Comment by kyrsjo** [comment]
Diff as seen on GitHub incomplete, I guess that's because it was merged and later reverted? We need to look at the diff locally then.

Also: In your comment on top you say that  
> SixTrack/utils.s90 will disappear. Its content is already somewhere in the main source code of SixTrack, ie:

however it appears that utils.s90 appeared, not dissappered?

**Comment by amereghe** [comment]
> Diff as seen on GitHub incomplete, I guess that's because it was merged and later reverted? We need to look at the diff locally then.

ok

> however it appears that utils.s90 appeared, not dissappered?

still working on that :)

**Comment by amereghe** [comment]
ok - I think I have everything I need for this first merge:
   * code ported to new internal structure (including DYNK) + computation re-worked a bit;
   * only uniform hollow eLenses;
   * no ions yet -- the code exits;
   * tests updated -- DYNK ones re-factored;
   * documentation updated;

The only missing points are:
   * report in the documentation the behavior with ions;
   * update the physics manual with the changes in the code structure/calculations;

**Comment by vkbo** [comment]
Great!

Could you also remove '-E elens' from the test .config.yml file in the .circleci folder? It should pass the elens tests now, but they are currently not run.
Also, there is not point updating fort.6 canonicals as they change all the time anyway.
There also seems to be some inconsistencies with the Sixin.zip files as the CheckTestInput_STF test fails.

**Comment by amereghe** [comment]
>  there is not point updating fort.6 canonicals as they change all the time anyway.

True - though I have activated the debug mode:
```
SETT--------------------------------------------------------------------
DEBUG
NEXT
```
To have the echo of the elens parsing - but this is not something fundamental.

> There also seems to be some inconsistencies with the Sixin.zip files as the CheckTestInput_STF test fails.

Strange, as I have re-generated all `Sixin.zip` files. I will double check.

amereghe:
Forgot to ask: what shall we do with the test `elensidealthin6d_DYNK_ZIPF` ? This is the `elensidealthin6d_DYNK` with simply the `ZIPF` block added, which zips some `DUMP` files. Honestly, I think that this is not strictly needed, although I fear that there are not so many tests for `ZIPF`...

rdemaria:
STF is failing, is it due to missing STF canonical?

vkbo:
Yeah, sorry, I meant STF, not Sixin.

vkbo:
Still doesn't pass the STF file test. Are all fort.90s and singletrackfiles up to date?

amereghe:
ok - I was not aware that for producing the `fort.90` I had to explicitly switch off `STF`. I've re-run the tests recompiling with `-STF`, regenerated the `fort.90` and committed them.

At the same time, I have removed `-E elens` from `.circleci/config.yml`

PR #617, opened 2018-08-17
adding comments to included files

rdemaria:
This adds comments to the include file to better locate them in the merged fortran files.
It could not be done for including in FOX because comments and empty line mess with dafor.

script used
```
for i in *.f90
do echo '!'"start $i" >$i.temp
cat $i >>$i.temp
echo '!'"end $i" >>$i.temp
mv $i.temp $i; done
```


vkbo:
This is a very good idea. Could you even make it say `! start include/$1` instead to make it clear it is an include file?

rdemaria:
Do you have a similar solution for FOX? Apparently empty lines are significant and comments around includes mess this up...

rdemaria:
yes for me it is ok

PR #618, opened 2018-08-17
Test IIR

kyrsjo:
Adding a test for the infinite inpulse responce filter, used as a simple integrator of phase. This example should probably be added to the DYNK manual as well...

To be merged before that other DYNK PR.


test/dynk_IIR/fort.3
```diff
@@ -0,0 +1,105 @@
...
+NEXT
+
+SYNC
+35640 .0003225 16 0 26658.864 938.2796 1
+ 1 1
+NEXT
+
+DYNK--------------------------------------------------------------------
+    DEBUG
+// Setting up the crabs (since we are starting tracking inside the crab bump)
```
vkbo: Comments can be indented if they use ! as comment character.

test/dynk_IIR/extra_checks.txt
```diff
@@ -0,0 +1 @@
+dynksets.dat          diff -q --strip-trailing-cr
```
vkbo: Second part is no longer used.

Issue #626, opened 2018-08-20
Binary FMA test fails with C/R


vkbo:
When run with checkpoint restarting, the singletrackfile is truncated.

kyrsjo:
It seems to affect most of the big FMA tests -- fma_binary, dump7_fma, ... it seems that all the dumps and the singletrackfile are truncated. fort.10 is also not matching.

It seems to run for a few restarts, but then crashing with `errout_status=-1`, the first error messages are:
```
******* Last     6 lines of file 'fort.93': *******
 SIXTRACR STOP/ABEND checking fort.10
 SIXTRACR STOP/ABEND writing a fort.10
 SIXTRACR STOP/ABEND writing fort.10, lines           0 /          30
 SIXTRACR STOP/ABEND copying fort.92 to fort.6
 SIXTRACR stop

******* Done writing tail of file 'fort.93' to stderr *******
******* Last     2 lines of file 'fort.6': *******
ABEND> ERROR Problems writing to fort.10. ierro = 5001
 SIXTRACR stop
******* Done writing tail of file 'fort.6' to stderr *******
Stopping, errout_status=   -1
STOP 1
SixTrack CR was killed: 256
```

Later it crashes like:
```
******* Last     4 lines of file 'fort.93': *******
 SIXTRACR STOP/ABEND checking fort.10
 SIXTRACR STOP/ABEND copying fort.92 to fort.6
 SIXTRACR stop

******* Done writing tail of file 'fort.93' to stderr *******
******* Last    40 lines of file 'fort.6': *******
  Number of reverse-revolutions:                            0
  Turns per coor.-printout:                             50000
  Flat bottom up to turn:                                   0
  Turns per print on dataset:                               1
  Acceleration up to turn:                                  0
  Turns per print on dataset:                               1
  Flat top number of turns:                              5000

  Turns per print on dataset:                               1
  Tracking start at element no.:                            1
  Initial amplitude-h in (mm):                              0.489704000
  Coupling  eps-y/eps-x:                                    0.071797000
  Number of C.O. iterations:                               50
  Precision of C.O. deviation:                              0.100000000E-11
  Precision of C.O. slope:                                  0.100000000E-14
  Number of q-adj. iterations:                             10
  Change in k-strength by:                                  0.100000000E-09
  Precision of q-adjustement:                               0.100000000E-09
  Number of chromat.-adj. iter.:                           10
  Change in sex.-strength by:                               0.100000000E-05
[...]

PR #629, opened 2018-08-21
FMA/DUMP/CR Fixes

vkbo:
Fixes issue #626 where binary FMA tests sometimes failed with C/R.
The issue was partially in prroe, abend and closeUnits where fort.93 was truncated.
The actual bug was in re-positioning of binary DUMP files changed in PR #596.


source/close.f90
```diff
@@ -28,6 +28,8 @@ subroutine closeUnits
 
   ! Then iterate through the first 1000 units
   do chkUnit=1, funit_minUnit-1
+    ! Do not close C/R files. handled by abend
+    if(chkUnit >= 91 .and. chkUnit <= 96) cycle
```
kyrsjo: Should probably also include `lout` and `6`. The comment should be "written to without preceding open in `abend`, closing handled in `abend`.

SixTrack version `v4.7.18` skips the following units in `closeUnits`: 5, 6, 36-58, 59-89 (STF), 91-97, 101-109, anything above 111 that's not 600 (if beam_expfile_open), not in DUMP, not dynksets.dat (aka 665). not a DYNKPIPE, not from SCATTER.

kyrsjo: Actually lout is written to in abend, but not closed. It probably should be closed.

vkbo: Most of these are either no loner in use, were never in use, or for the named files, have been replaced by dynamic units.

I'll add fort.6 to the list. Perhaps the track files too?

PR #639, opened 2018-08-24
buildLibraries rewrite

vkbo:
This PR splits up the buildLibraries script, and adds build from source for HDF5 with static linking to SixTrack.


buildLibraries.sh
```diff
@@ -1,141 +1,53 @@
...
 
-cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_BZip2=OFF -DENABLE_ZLIB=ON -DENABLE_CAT=OFF -DENABLE_CPIO=OFF -DENABLE_EXPAT=OFF -DENABLE_INSTALL=OFF -DENABLE_LIBXML2=OFF -DENABLE_LZMA=OFF -DENABLE_NETTLE=OFF -DENABLE_OPENSSL=OFF -DENABLE_TAR=OFF -DENABLE_CNG=OFF -DENABLE_ICONV=OFF -DENABLE_TEST=OFF -DZLIB_LIBRARY=$ZLIB_PATH -DZLIB_INCLUDE_DIR=$ZLIB_BASE/install/include -G "Unix Makefiles" ../libarchive -LH
-
-if [[ $(pwd) == /afs/* ]]; then
-    #AFS doesn't like parallel make
-    make
-else
-    # Machines with low memory doesn't like an automatic -j
-    make -j 4
+if [ $HDF5 = true ] || [ $ALL = true ]; then
```
kyrsjo: Zlib is also a dependency of HDF5 as well. I propose to make a separate script for building zlib, and calling this when HDF or libarchive is true, before building either of them.

vkbo: This has already been done in a different branch I have. There is no rush to merge this, so we can just merge both when the other is ready.

rdemaria: If HDF5 static is possible I would also include the feature by default

vkbo: That would trigger the need to ensure it works on all platforms and build types. I would not recommend that at this stage.

kyrsjo: Agreed. HDF5 is a bit fickle sometimes, and it is not possible to build to a true static binary as it explicitly links dlopen. AFAIK LibArchive has a similar issue.

PR #640, opened 2018-08-24
Testing framework for comparison between SixTrack and MAD-X

jodander:
This is the testing framework previously covered in today's SixTrack meeting. It provides a Python interface for defining tracking code independent (i.e. SixTrack and MAD-X) simulation runs, executing these runs, reading outputs as DataFrames and outputting the difference between them.

Right now a rudimentary `README.md` file is included which gives a broad idea of what the script does but I believe this should be decipherable from the commented code.

The `exampleTest.py` script contains a basic example that should elucidate how the procedure works. Also, in the case you would be interested in additional example cases I still have some on [here](https://github.com/jodander/SixTrack/blob/devjodan/source/pytools/compareSix2Madx/element_test.py) but I did not have enough time to properly refactor them so beware.

Lastly, thanks again for having me and if you have any questions I'll gladly answer them.


jodander:
Hey @Kyrre!

Did not add the tests before as they were not properly refactored and I wasn't sure if they were of interest. They are now added in a format that is easily modified to whatever purpose you may have in mind.

Also, @hyperboleburger is an old account of mine that apparently is associated with my Git setup on my old laptop, hence the misattribution of the last commit.

kyrsjo:
Good! Please add the new script to the readme + add a link to the presentation :)

Regarding the committer name, one way to fix it (before you commit anything else on the branch) would be to do a git `commit --ammend`, then `git push --force` to your branch. Or we can just leave it.

kyrsjo:
Thanks! From me it's ready, however @tpersson should really do the formal review. Then we can merge :)

tpersson:
Sure! I will try to do it tomorrow !

jodander:
An indentation error was just fixed. Right now @tpersson and I are making sure it runs properly on his setup, once that has been validated it should be ready for a merge.

kyrsjo:
Link to the actual PR:
https://github.com/jodander/SixTrack/pull/1

rdemaria:
Do you see something on the top right of the page?

PR #650, opened 2018-08-31
Bugfix for DYNK FIR/IIR

vkbo:
Incrementing of dynk_niData was missing in FIR/IIR parsing. Failed bounds check with debug.


kyrsjo:
Hmm, is that data actually ever used? It seems the same thing is also stored directly in the `dynks_fun` array, and the reference to the data in `dynk_iData` is not even saved? Or am I missing something?

vkbo:
I'm not sure what N is used for, but the value is definitely saved.
Can we at least fix the bug so it runs properly?

PR #653, opened 2018-08-31
Fix IIR the right way

kyrsjo:
Fixes IIR the right way by completely removing the broken and unused variable. This superseedes #650 and should be merged once #651 (which reverts #640 ) has been merged.


vkbo:
Does the manual need to be updated?

kyrsjo:
No manual update needed, it's a strictly internal memory management mess-up.

PR #656, opened 2018-09-03
Change sign of the rf-multipole 4th order skew

tpersson:
There were no tests including skew rf-multipole 2nd, 3rd and 4th order. I made them into a single test. They were also compared to MAD-X by Joel and I used the scripts to generate this test.


rdemaria:
I see the change of the sign of the transverse part  but not the longitudinal one. Was the energy kick wrong, then?

tpersson:
The longitudinal kick was correct.. The interesting thing which I discovered was that the FOX changed the energy in opposite. Actually, with the way it was defined in thin_track was not symplectic. This is fixed in the last commit. There was also another error there where one x and y seem to have been swapped.

rdemaria:
Are the other rf-multipoles ok?

tpersson:
For the tracking part I am confident but when it comes to the FOX part I don't dare to give any guarantees.

rdemaria:
You can always check that closed orbit is the closed orbit with tracking. You would need a lattice for which the rf-multipole effect is very strong. You could try the present version and the one after your fix to see whether the example is sensitive to the error you spotted.

tpersson:
I know. This is what I have done for the tests of the other elements. It is somewhat time consuming and in this case of the 4th order skew Rf-multi it is still not perfect (it became better). I think it might be due to the precision in the FOX part together with the multiplying with very small numbers and dividing by the speed of light..

rdemaria:
If there are doubts on the rf-multipoles we need to check them. The other option is to rewrite them using the general algorithm as in madx, that should be done anyway together with the tilt support.

tpersson:
I can guarantee that they have never been as correct as they are now and I don't have any more doubts on the Rf-multipoles than I have for other elements.

vkbo:
Does "for the time being" mean it is ready to merge or not?

PR #657, opened 2018-09-03
Minor bugfixes

vkbo:
This fixes:

* Undeclared variables in beamgas. This was due to a leftover COLLIMAT flag that prevented them from being decalred.
* NAFF c interface call in FMA returns double precision floats, which when SixTrack is built with single or quad must be cast to double.


vkbo:
Tagging @skostogl for reference for the NAFF/FMA fix.

kyrsjo:
I checked the test with quad, and the tunes are the same down to 10^-7 (in the worst case) or better.

vkbo:
Thanks for checking.
There are already significant differences to FMA with single/quad in that the NaNs are zero (probably due to CRLIBM being off). Consistency between float precision is an issue that should be addressed. But it doesn't affect production use.

PR #659, opened 2018-09-04
Bouncy NaNs

vkbo:
mathlib_bouncer did not return the same value in the atan2(0,0) case with and without CRLIBM.

The functionality of the internal atan2 function is correct with regards to the IEEE standard. The atan2_rn/rd/ru/rz routines now calls the internal atan2 in the cases where x=y=0 and y=0, x>0

Don't merge quite yet. Updating FMA canonicals and checking tests for all compilers first.

Also, this PR includes the 3 commits from #658 as that is required to build single/quad with ifort for testing.


vkbo:
atan2 is also used for collimation, so please confirm that this change is ok.

source/bouncy_castle.f90
```diff
@@ -5,14 +5,22 @@
...
   implicit none
 
   !These are the actual "bouncy functions" users are supposed to call
   public :: sin_mb, asin_mb, sinh_mb, cos_mb, acos_mb, cosh_mb, tan_mb, atan_mb, atan2_mb, exp_mb, log_mb, log10_mb
 
-  real(kind=real64), parameter :: mb_pi  = 3.1415926535897932d0
-  real(kind=real64), parameter :: mb_pi2 = 1.5707963267948966d0
+! real(kind=fPrec), parameter :: mb_pi   = 3.1415926535897932d0
+! real(kind=fPrec), parameter :: mb_pi2  = 1.5707963267948966d0
+  real(kind=fPrec), parameter :: mb_pi   = 3.141592653589793238462643383279502884197169399375105820974_fPrec
```
kyrsjo: Is the _fPrec accepted by ifort here?

source/bouncy_castle.f90
```diff
@@ -756,37 +761,33 @@ real(kind=real64) function asin_rd(x)
...
-          else
-             atan2_rd=mb_pi
-          endif
-       else
-          atan2_rd=atan_rd(y/x)
-          if (x.lt.0d0) then
-             atan2_rd=sign(mb_pi,y)+atan2_rd
-          endif
-       endif
+      if (y.eq.0d0) then
```
kyrsjo: Could this if/else logic be moved to the calling `_mb` function, instead of being duplicated?

kyrsjo:
@rdemaria :
> We could rely more on the intrinsic for corner cases, but we leave it like this for the time being

Yes, but apparently the intrinsics (i.e. NAGFOR) are not really consistent.

source/bouncy_castle.f90
vkbo: Yes. Should it not?

source/bouncy_castle.f90
vkbo: Yes, there are several things that can be improved in the flow of this module. I don't have the time for re-structuring the code now.

vkbo:
@rdemaria is referring to the sign(pi,y) case as well, which could be handled by the intrinsic atan2(), but then we also depend on its definition of pi, which may or may not be the same as we provide?

PR #662, opened 2018-09-10
Fix update of energy variables 

tpersson:
Do NOT merge, tests are not updated (Yes, I know that they should be and they will be) but before we are sure that this is the final version I will not update the tests.
This PR fixes the following issues:

-  Updates dpsv1 in beam-beam and beamgas. This was probably never done. These variables are used for the bends and hence if there was a bend the effect was not included before the particle passed through another element like the cavity. 
-  Updates the omoidpsv everywhere. This was missing on many places and again this is used for bends but only has an impact when mtc=!1 (relative mass to charge ratio)
-  Correct definition of dpsv1 everywhere. dpsv1 was wrongly defined in most places but this only had impact for ions. 
- In the collimation part I changed how the energy was updated. Please check if this is correct, maybe there is a special reason why it was done as it was.  

Please review the changes carefully it is very easy to miss something here... 

Also I found that dpsq(j), dpd(j) are updated in the mod_energy.f90 but they are never used by the elements in track_thin.f90. However, they are used on other places. Is the idea to have it that general?


rdemaria:
dpsq(j), dpd(j) are use in the thick tracking and normally updated by 

`if(ithick.eq.1) call envarsv(dpsv,moidpsv,rvv,ekv)`

mod_particles should have this as well I believe

vkbo:
The `part_updateEnergy` routine has this, yes.

rdemaria:
@vkbo I cannot see the line in the source code...

rdemaria:
@tpersson can you update the user manual with the HION block and possibly DISP if it was in Pascal's thesis?

vkbo:
No, you're right. It only has:
`if(ithick == 1 .and. e0 /= e0o) call synuthck`

vkbo:
But synuthck updates those variables.

tpersson:
Yeah my question was simply if the idea was to have part_updateEnergy subroutine that general that it would be used for both thin and thick. The performance loss is rather small...

vkbo:
The current routine can be called after changing either reference energy, particle energies or particle momenta. For changes to the delta variables, it may be better to create a new routine that updates the arrays the other way around. It will probably be faster than adding even more if statements to the current logic, although as long as they are outside of the particle loop. the cost is minimal.

There is already an if-statement in the loop. It should probably be moved outside as well. So far the routine has not been used in any performance critical places.

amereghe:
Collimation part looks fine.
It was done that way since k2 (the original scattering routine in collimation-sixtrack) is not able to treat ions, and indeed all the ion-cleaning simulations are run with the Fluka-SixTrack coupling

tpersson:
Ok! I see! I thought it could be something like that but it was unclear to me since some of the ion variables were still updated.

vkbo:
> `if(ithick == 1 .and. e0 /= e0o)` does not look right. dpsq(j), dpd(j) should be updated also if `e0==e0o`...

Hang on ...

The loop in the routine already updates those two arrays:
```
  do j=1, napx
    ejfv(j)     = sqrt(ejv(j)**2 - nucm(j)**2)       ! Momentum [MeV/c]
    dpsv(j)     = (ejfv(j)*(nucm0/nucm(j))-e0f)/e0f  ! Delta_p/p0 = delta
    dpsv1(j)    = (dpsv(j)*c1e3)/(one + dpsv(j))
    dpd(j)      = one + dpsv(j)
    dpsq(j)     = sqrt(dpd(j))
    oidpsv(j)   = one/(one + dpsv(j))
    moidpsv(j)  = mtc(j)/(one + dpsv(j))             ! Relative rigidity offset (mod_hions) [MV/c^2]
    omoidpsv(j) = c1e3*((one-mtc(j))*oidpsv(j))
    rvv(j)      = (ejv(j)*e0f)/(e0*ejfv(j))          ! Beta_0 / beta(j)
  end do
```

PR #663, opened 2018-09-21
Dynk for multipoles and Beam-Beam 

tpersson:
- Implements dynk for beam-beam (only expert). No new closed orbit calculation is re-triggered. An example is added when it gives identical results to apply dynk at turn 1 compared to the normal way. 
-Dynk for multipoles. Including a test with 100s of multipoles. 
- removed BBIV , AAIV and the index over the machines. 
-Removed a lot of multiplication and division by 1 in the initialization of the multipoles. 
-Updated the documentation with the new features.


source/dynk.f90
```diff
@@ -2503,7 +2556,36 @@ real(kind=fPrec) function dynk_getvalue(element_name, att_name)
           goto 100 ! ERROR
         end if
 
-        ! Not yet supported : Multipoles (11)
+
+      case(11)
+      im=irm(ii)
```
vkbo: Please follow the indentation of the other cases. The content of a case should be indented one level.

source/dynk.f90
```diff
@@ -2503,7 +2556,36 @@ real(kind=fPrec) function dynk_getvalue(element_name, att_name)
...
+      if(att_name=="scaleall") then
+
+      else if(att_name(1:1) == "a" .or. att_name(1:1) == "b") then
+        if(LEN_TRIM(att_name) .eq. 5) then
+           range = 3
+          read(att_name(2:2), *) orderMult
```
vkbo: Don't use read() for casting numerical values. These should be read with the chr_cast interface in module string_tools. In general: for float values ensures proper rounding, and for integers and floats it allows for error checking and capturing.

source/dynk.f90
```diff
@@ -2519,6 +2601,16 @@ real(kind=fPrec) function dynk_getvalue(element_name, att_name)
       ! Not yet supported : AC dipole (16)
 
       ! Not yet supported : beam-beam separation (20)
+      case(20)
+       if (att_name == "h-sep") then ! [mm]
```
vkbo: Wrong indentation in this section. Should be two space indentation in all new code in SixTrack.

source/include/beam6d.f90
```diff
@@ -22,13 +22,12 @@
...
   yv(1,j)=((track6d(2,j)*c1e3+clobeam(4,imbb(i)))-beamoff(4,imbb(i)))*moidpsv(j)
   yv(2,j)=((track6d(4,j)*c1e3+clobeam(5,imbb(i)))-beamoff(5,imbb(i)))*moidpsv(j)
-
   !TODO check for ions?
   
   ejfv(j)=(dpsv(j)*e0f+e0f)*(nucm(j)/nucm0) 
   ejv(j)=sqrt(ejfv(j)**2+nucm(j)**2)
   rvv(j)=(ejv(j)*e0f)/(e0*ejfv(j))
-  
+  dpsv1(j)=(dpsv(j)*c1e3)*oidpsv(j)
```
rdemaria: Why? I thought this was fixed in the other pull request on energy variable

source/track_thin.f90
```diff
@@ -105,47 +105,16 @@ subroutine trauthin(nthinerr)
       ktrack(i)=31
       goto 290
     endif
-#include "include/beams21.f90"
```
rdemaria: Where did all the includes go?

tpersson: This is now in initalize_elemnet in sixtrack.f90. 

I didn't  like the way that was done there with all the includes because it meant that there had to be in the order of 10 loops over number of particles but the information about the particles were never used. I think it is much cleaner now.

source/include/beam6d.f90
tpersson: I think this is some fail linked to git and my merge of master into the branch because the line is there in master. I had some conflicts and maybe this caused some confusion for git.

tpersson:
It is strange but something most have gone wrong in the merge.. I will investigate..

source/include/beam6d.f90
rdemaria: it is a bit fishy though, I would like to understand more...

BTW with a grep I have just found
```
dpsv1(j)=(dpsv(j)*c1e3)*moidpsv(j)
```
in track_thick that should be wrong to my understanding. Can you confirm?

source/track_thin.f90
vkbo: Killing includes is a good thing! There should be an award for that.

source/include/beam6d.f90
tpersson: Yeah I can confirm. Must have missed it.

vkbo:
My approval is for the changes I requested. Final approval of the implementation should come from @rdemaria .

source/track_thin.f90
rdemaria: Ok for me!

PR #666, opened 2018-09-27
Ranecu Updates

vkbo:
Minor update to the random number generator.
The code has been re-indented and cleaned up a bit (commit 1), and a new routine `recuinit` for initialising the random generator with either one or two seeds has been added (commit 2). This init routine should replace calls to `recuin` when setting the seeds provided in `fort.3` during input parsing. `recuinit` ensures the seeds are within valid range, and will generate a reasonable seed2 if only one seed is provided.

### DIFF
--- a/source/ranecu.f90
+++ b/source/ranecu.f90
@@ -4,15 +4,15 @@
 !
 ! References:
 !
-! F. James, A review of pseudorandom number generators,
-!  Computer Physics Communications,
-!  Volume 60, Issue 3, 1990, Pages 329-344, ISSN 0010-4655,
-!  https://doi.org/10.1016/0010-4655(90)90032-V.
-!  http://www.sciencedirect.com/science/article/pii/001046559090032V
+! [1] F. James, A review of pseudorandom number generators,
+!     Computer Physics Communications,
+!     Volume 60, Issue 3, 1990, Pages 329-344, ISSN 0010-4655,
+!     https://doi.org/10.1016/0010-4655(90)90032-V.
+!     http://www.sciencedirect.com/science/article/pii/001046559090032V
 !
-! P. L'Ecuyer. 1988. Efficient and portable combined random number generators.
-!  Commun. ACM 31, 6 (June 1988), 742-751. DOI=10.1145/62959.62969
-!  http://doi.acm.org/10.1145/62959.62969
+! [2] P. L'Ecuyer. 1988. Efficient and portable combined random number generators.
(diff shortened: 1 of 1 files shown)


source/ranecu.f90
```diff
@@ -73,6 +73,38 @@ subroutine ranecu(rvec,len,mcut)
 
 end subroutine ranecu
 
+! Init the random generator, that is, set seeds with proper value checks as described in [2]
```
kyrsjo: Maybe mention in the comment here that if not given, then the 2nd seed is calculated automatically.

source/ranecu.f90
```diff
@@ -73,6 +73,38 @@ subroutine ranecu(rvec,len,mcut)
...
+subroutine recuinit(is1,is2)
+
+  use crcoall
+
+  integer,           intent(in) :: is1
+  integer, optional, intent(in) :: is2
+
+  if(is1 < 1 .or. is1 > 2147483562) then
```
kyrsjo: Make these magic numbers into parameters (constants) for the module, then reference them by name?

source/ranecu.f90
```diff
@@ -73,6 +73,38 @@ subroutine ranecu(rvec,len,mcut)
...
+    iseed2 = iseed1 + 19770404
+    if(iseed2 > 2147483398) then
+      ! Subtract max of seed 1 to ensure lower than max of seed 2
+      iseed2 = iseed2 - 2147483562
+    end if
+  end if
+
+end subroutine recuinit
+
 ! Set the seeds
```
kyrsjo: Add to the comment `(to be used for continuing a random number stream with seeds extracted using 'recuut()')`

source/ranecu.f90
vkbo: I did it this way for readability and consistency since every other constant in this module is done the same way. The same values are also used in the generator.

This way, at least, you can do a direct comparison to the values described in the paper if you want to check the code.

source/ranecu.f90
vkbo: It is generally bad practice to make comments explaining the obvious, but I suppose. I think it is more important to put it into a developer manual (or a usage description at the top of the module like I've done for `hdf5_output` and `strings`.

source/ranecu.f90
vkbo: Yes, I'll add that. Good idea.

source/ranecu.f90
vkbo: Added a usage comment at the top.

PR #668, opened 2018-09-28
Beamgas build fix

vkbo:
BEAMGAS flag now builds again. Was just missing a `use mod_hions`.
Also, in the same PR, the hard coded file units have been replaced by units assigned by the `file_units` module.


vkbo:
Yes, the mod_hions was a fix for the build, and I noticed when I fixed this that we forgot to removed fixed unit numbers in this module. This should have been done a long time ago. :)

PR #672, opened 2018-10-03
Space indentation in CMake

vkbo:
Converted the CMake file to space indentation like the rest of the source as people keep mixing tabs and spaces.


kyrsjo:
This is also the case for `test/bin/SixTestWrapper.cpp`

PR #673, opened 2018-10-04
Less fort.6 output

vkbo:
This PR adds a flag to the `SETTINGS` block that controls writing the particle summary at the end of the simulation. The amount of output is also controlled by the `QUIET` flag as before.

The new flag is `PARTICLESUMMARY` and when present is taken as `.true.`. An explicit parameter can be added, as the default behaviour otherwise depends on the number of particles tracked. If it is 64 or less, it is on by default, and otherwise it is off.

A modification to the `chr_cast` routine: it now also accepts `ON` and `OFF` as logical values (should work wherever the casting routine is used for parsing.) In relation to that, I also added two functions to the `string_tools` module for converting a string to upper or lower case. `chr_toUpper` and `chr_toLower`.

This PR also stops `track_thin` from calling the collimation module if it sees a collimator aperture marker.


rdemaria:
Can give more context for the last line? Why was it introduced, why you removed it now?

vkbo:
Which last line are you referring to?

rdemaria:
"This PR also stops track_thin from calling the collimation module if it sees a collimator aperture marker."

vkbo:
Ah. This was never introduced. It shouldn't treat an aperture marker as a collimator. It's not a collimator. It doesn't make a difference to te result anyway, as the collimation module would reject it. But filtering it out earlier removes all the warnings.

PR #675, opened 2018-10-09
fixing parsing of aperture input infos

amereghe:
In particular, tilt angle set as last column, as implemented in MADX, instead of being last but two (as in the coupling version of SixTrack).

All users of the Fluka-SixTrack coupling will have to update their aperture model


vkbo:
Can I merge this? Or will there be more added?

amereghe:
Hi Veronica,
as already mentioned in a discussion, I would wait for:
- documentation (I am working on that);
- having a tag with all other changes, and then merge this pull request with a new tag. This is simply because the pull request makes the input of the LIMI block compliant with MADX-generated aperture markers and will lose the compatibility with the pre-processing tools to the Fluka-SixTrack coupling, still quite widely used.

vkbo:
Ok. then we will hold off until after 5.0.3.

vkbo:
To be merged for SixTrack 5.1

PR #670, opened 2018-10-02
Pythia8 Support for Scatter

vkbo:
This PR adds a PYTHIA module and a PYTHIA compiler flag to SixTrack.
The PYTHIA flag builds the interface and links to libpythia.so. To build the library, run .`/buildLibraries.sh pythia`

The PYTHIA configuration is handled with the PYTHIA block in `fort.3`, and when defined, gives access to a new generator in the `SCATTER` block called `PYTHIASIMPLE`.

The PR also includes changes to the HDF5 module, as well as adds a few HDF5 output options elsewhere in SixTrack. Specifically a dump of the collimation DB and the aperture model. Interface routines for writing arrays of metadata has also been added to the `HDF5` module, and such metadata added to the `DUMP` module.

The PR is not quite finished, the following remains:
* Adding PYTHIA to manual
* Updating manual for SCATTER and HDF5
* Updating canonicals for scatter_log (mainly change in header format)

PR #677, opened 2018-10-10
fix hdf5 and add parameters

rdemaria:
add fields to 101 and fixes hdf5
(not merge yet)


source/dump.f90
```diff
@@ -847,29 +847,33 @@ subroutine dump_initialise
...
+          setFields(10) = h5_dataField(name="(E-E0)/E0",   type=h5_typeReal)
+          setFields(11) = h5_dataField(name="E",      type=h5_typeReal)
+          setFields(12) = h5_dataField(name="PC",     type=h5_typeReal)
+          setFields(13) = h5_dataField(name="BETA0/BETA",     type=h5_typeReal)
+          setFields(14) = h5_dataField(name="(P-P0)/P0",   type=h5_typeReal)
+          setFields(14) = h5_dataField(name="1000*(P-P0)/P0",   type=h5_typeReal)
+          setFields(15) = h5_dataField(name="P0/P",   type=h5_typeReal)
+          setFields(16) = h5_dataField(name="MASS",   type=h5_typeReal)
+          setFields(17) = h5_dataField(name="CHARGE", type=h5_typeReal)
+          setFields(18) = h5_dataField(name="(CHARGE/CHARGE0)*(MASS0/MASS)", type=h5_typeReal)
```
vkbo: I think this is too much for a column header label. They need to be long enough to describe what's in them, but since these labels are used when accessing the data itself, they should be compact and more user friendly.

In my PR #670 I'm adding support for dataset metadata, where such information can be added to string arrays `colNames` and `colUnits`.

vkbo:
Closing this. PR #937 fixes the bugs in format 101.

PR #679, opened 2018-10-19
SixTrack Performance Fixes

vkbo:
Started fixing some of the performance issues in SixTrack.

In this PR:
* Replacing `myisnan` with the IEEE intrinsic routine. There was a large performance loss when a large number of calls were made to `myisnan` and inlining failed. Ensuring inlining happened solved the issue, but a small performance boost was added by using the intrinsic function as well.

Remaining Issues:
* Subroutine `lostpart` is still slow.
* Closed orbit search is significantly slower than before. Especially for certain tests like `dump9`.


vkbo:
Do you have any reason to believe optimisation should change this? That is exactly what CRLIBM is for. Optimisation is known to have an effect on 80-bit floating points, but we're not using that.

In any case, if it passes the tests (it does) it produces the expected results.

vkbo:
Just an added note: -O3 enables function inlining, which is why I changed it. Alternatively, we can enable inlining separately and stay with -O2.

rdemaria:
The -O2 was used to match the output of different compilers. Probably is not needed any more, but we should check if test passes at least with intel and if not put again -O2 for intel only.

rdemaria:
Hello, -O2 was needed in the past for numerical reproducibility. I agree if test passes is ok, I would only check with the intel compiler if possible. @vkbo can you do a test? If not possible let's merge and do it later...

vkbo:
> The -O2 was used to match the output of different compilers. Probably is not needed any more, but we should check if test passes at least with intel and if not put again -O2 for intel only.

The change from -O2 to -O3 is only for gfortran. ifort is already running with -O3.

rdemaria:
Ok, let's merge. Which gfortran are you using? I can test on 7.3 if you did not.

vkbo:
8.3 at the moment.

PR #680, opened 2018-10-23
Remove IEEE_ARITHMETIC

vkbo:
For some reason there is a heavy performance cost to using ieee_arithmetic on gfortran only. ifort and nagfor is fine. This is used for NaN checking.

Not ready to be merged yet because ifort hangs when linking. Not sure why yet.


vkbo:
Closing this PR. Will make a new from a different branch.

PR #684, opened 2018-10-30
DIST block updates

vkbo:
This PR cleans up the DIST block a bit and brings it up to date with current parsing scheme, file units assignment, and CRLIBM wrapping.

In preparation for the full DIST block to come, it also disables the `idfor = 3` way of enabling the DIST block. If a DIST block is present, it takes precedence over other formats.


source/main_cr.f90
```diff
@@ -1313,8 +1295,8 @@ end subroutine envarsv
```
amereghe: ok the if statement for identifying a particle which should have longitudinal variables initialised to 0,1, ... should be modified into something like:
`
if ( abs((ejfv(j)/nucm(j))/(e0f/nucm0)-one).lt.c1m15 .or. abs((ejv(j)/nucm(j))/(e0/nucm0)-one).lt.c1m15 ) then
`

source/main_cr.f90
```diff
@@ -1313,8 +1295,8 @@ end subroutine envarsv
 
 !            ejfv(j)   = e0f
 !            ejv(j)    = e0
-            ejfv(j)   = e0f*(nucm(j)/nucm0)          ! P. HERMES for hiSix
-            ejv(j)    = sqrt(ejfv(j)**2+nucm(j)**2)  ! P. HERMES for hiSix
+            ejfv(j)   = e0f*(nucm(j)/nucm0)
+            ejv(j)    = sqrt(ejfv(j)**2+nucm(j)**2)
             dpsv(j)   = zero
             oidpsv(j) = one
```
amereghe: Here we should add:
```
moidpsv(j)=mtc(j)
omoidpsv(j) = c1e3*(one-mtc(j))
```

source/main_cr.f90
```diff
@@ -1290,13 +1277,8 @@ end subroutine envarsv
...
-!         old proton only terms:
-!          ejv   (j) = sqrt(ejfv(j)**2+pma**2)
-!          dpsv  (j) = (ejfv(j)-e0f)/e0f
-!          oidpsv(j) = one/(one+dpsv(j))
-
-          ejv   (j)   = sqrt(ejfv(j)**2+nucm(j)**2)              ! hiSix
-          dpsv  (j)   = (ejfv(j)*(nucm0/nucm(j))-e0f)/e0f         ! hiSix
+          ejv   (j)   = sqrt(ejfv(j)**2+nucm(j)**2)
+          dpsv  (j)   = (ejfv(j)*(nucm0/nucm(j))-e0f)/e0f
           oidpsv(j)   = one/(one+dpsv(j))
```
amereghe: I think that here we should put something like (maybe right after `mtc(j) assignment`):
```
if ( abs(nucm(j)/nucm0-one).lt.c1m15) then
   nucm(j)=nucm0
   if (nzz(j).eq.zz0 .or. naa(j).eq.aa0) then
     naa(j)=aa0
     nzz(j)=zz0
     mtc(j)=one
   else
     write(lout"(a)") "DIST> ERROR mass and/or charge mismatch wrt sync particle"
     call prror(-1)
   end if
end if
```

source/mod_dist.f90
```diff
@@ -14,28 +14,14 @@
 !    m:          rest mass [GeV/c2]
 !    pc:         particle momentum [GeV/c]
 !    dt:         time delay with respect to the reference particle [s]
-!  - aa,zz and m are now taken into account for hisix
 !
 !  NOTA BENE:
 !  - id, gen and weight are assigned by the fluka_mod_init subroutine;
```
amereghe: I think that, for the time being, we can keep this comment and its implementation as it is.
But at a certain point, we have to merge these vars properly...

source/mod_dist.f90
```diff
@@ -14,28 +14,14 @@
...
 !  - z and zp are actually useless (but we never know);
-!  - aa, zz and m are not stored at the moment (safer decision from the code
-!    point of view, until a decision about ion tracking is taken);
-!    the subroutine fluka_send is then responsible for using the corresponding
+!  - the subroutine fluka_send is then responsible for using the corresponding
 !    values for protons through the interface, whereas the subroutine fluka_receive
 !    simply ignores the values passed through the FlukaIO interface;
```
amereghe: This comment is no longer true. It can be removed

source/main_cr.f90
```diff
@@ -1290,13 +1277,8 @@ end subroutine envarsv
...
           mtc     (j) = (nzz(j)*nucm0)/(zz0*nucm(j))
           moidpsv (j) = mtc(j)*oidpsv(j)
```
amereghe: I guess that here we should also initialise `dpd(j)` and `dpsq(j)`, as done in the energy update function, though I don't know exactly what these variables are for... and set suitable values for the on-momentum particles...

source/sixtrack.f90
```diff
@@ -702,7 +702,7 @@ subroutine daten
 
   case("DIST") ! Beam Distribution
     if(openBlock) then
-      continue
+      dist_enable = .true.
```
amereghe: (picky) shouldn't this flag be set to `.true.` only if there is a file to read?

rdemaria:
I propose that this PR converges towards the new dist block. I added the wiki page:

https://github.com/SixTrack/SixTrack/wiki/new-DIST-and-Simul-block

to discuss the new features. This is a based on the discussions we had before, plus few additional ideas that came afterwards. I propose to update the proposal in the wiki and rediscuss the features to implement here.

The main points for me is:
1) whether we want to keep backward compatibility with the existing (undocumented) DIST block. If yes I would keep DIST as it is and find a new keyword (e.g. PARTICLES) for the new DIST block.
2) whether we want to commit to a default file format. I prefer a mandatory explicit format in which columns are explicitly named and no default.

rdemaria:
After discussion with Alessio and Veronica, we agreed to implement SIMUL block, mulitple DIST block for otional SUBSAMPLE keyword and, only temporarily, accept the old DIST block. The feature will be still experimental until we implement a column base FORMAT keyword.

vkbo:
Note that this PR is only for bringing the existing DIST block up to speed in terms of CRLIBM and code structure (moving stuff out of maincr). It is not implementing any of the new functionality. That will be in other PRs.

source/sixtrack.f90
vkbo: The idea is that the presence of the DIST block should trigger the call to dist_readDist, and the parsing will fail there if the file is non-existent (which includes the case of a missing READ flag).

PR #686, opened 2018-10-31
Moving parsing of fort.13 out of maincr

vkbo:
This PR moves the parsing of fort.13 into a separate module for "legacy" input files. It also uses the automatic casting routines from string_tools.

This is a branch off of #684, so should be reviewed and merged after that.

### DIFF
--- a/source/main_cr.f90
+++ b/source/main_cr.f90
@@ -1266,8 +1266,8 @@ end interface
     if(dist_echo) call dist_echoDist
   end if
 
-      do 340 ia=1,napx,2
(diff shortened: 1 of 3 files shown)


vkbo:
Yes and no.
I think it is cleaner to keep mod_dist only for the DIST block. The read_input module was written specifically for the legacy support for old input formats to get them "out of the way".

amereghe:
ok - if the long term plan is to remove them, then I fully support the choice :)

vkbo:
A'ight!

PR #687, opened 2018-10-31
Collimation dist format 0

vkbo:
This PR adds do_thisdist = 0 as an option to collimation for bypassing the internal distribution generator.


source/collimation.f90
```diff
@@ -1799,18 +1801,16 @@ subroutine collimate_start_sample(nsample)
...
-    sigmv(i) = mys(i+(j-1)*napx00)
-    ejv(i)   = myp(i+(j-1)*napx00)
+    if(do_thisdis /= 0) then
+      xv(1,i)  = c1e3 *  myx(i+(j-1)*napx00) + torbx(1)
+      yv(1,i)  = c1e3 * myxp(i+(j-1)*napx00) + torbxp(1)
+      xv(2,i)  = c1e3 *  myy(i+(j-1)*napx00) + torby(1)
+      yv(2,i)  = c1e3 * myyp(i+(j-1)*napx00) + torbyp(1)
+      sigmv(i) = mys(i+(j-1)*napx00)
+      ejv(i)   = myp(i+(j-1)*napx00)
+    end if
```
amereghe: I would maybe swap the order of the loop/if-statement:
```
if(do_thisdis /= 0) then
   do i=1,napx00
      [...]
      ejv(j) = [...]
   end do
end if

do i=1,napx00
```

source/collimation.f90
```diff
@@ -1250,21 +1250,23 @@ subroutine collimate_init()
...
         call readdis(filename_dis, mynp, myx, myxp, myy, myyp, myp, mys)
-      else if(do_thisdis.eq.5) then
+      else if(do_thisdis == 5) then
         call makedis_ga(mynp, myalphax, myalphay, mybetax, mybetay, myemitx0_dist, myemity0_dist, &
      &           myenom, mynex, mdex, myney, mdey, myx, myxp, myy, myyp, myp, mys, enerror, bunchlength )
-      else if(do_thisdis.eq.6) then
+      else if(do_thisdis == 6) then
         call readdis_norm(filename_dis, mynp, myalphax, myalphay, mybetax, mybetay, &
      &           myemitx0_dist, myemity0_dist, myenom, myx, myxp, myy, myyp, myp, mys, enerror, bunchlength)
       else
```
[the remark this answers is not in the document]
vkbo: Sure. I just wanted to avoid changing anything else. Updated now/

source/collimation.f90
```diff
@@ -1309,14 +1311,14 @@ subroutine collimate_init()
...
-    call funit_requestUnit('dist0.dat', dist0_unit)
-    open(unit=dist0_unit,file='dist0.dat') !was 52
-    if(dowrite_dist) then
+    if(dowrite_dist .and. do_thisdis /= 0) then
+      call funit_requestUnit('dist0.dat', dist0_unit)
+      open(unit=dist0_unit,file='dist0.dat') !was 52
       do j = 1, mynp
         write(dist0_unit,'(6(1X,E23.15))') myx(j), myxp(j), myy(j), myyp(j), mys(j), myp(j)
       end do
+      close(dist0_unit)
```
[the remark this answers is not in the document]
vkbo: Done.

source/collimation.f90
vkbo: This is exactly what triggered the gfortran compiler bug in the closed orbit addition. This is therefore risky since we have no proper tests for this.

vkbo: I manually inlined the calculation, but I have mo way of checking it.

PR #692, opened 2018-11-02
Switch to NAFFlib.

kparasch:
Added NAFFlib submodule,
removed old naff source,
modified CMakeLists.txt to compile NAFFlib,
modified cmake_six to update NAFFlib submodule,
replaced test/fma_naff/fma_sixtrack.canonical where the estimated tunes differed sometimes but only on the 6th+ decimal.


kparasch:
I have not pinpointed the exact reason that this happens. There are some small changes between the codes like for example the fact that I have fixed the order that all numerical operations are applied ( e.g. a*b/c  to (a*b)/c ). What puzzles is me is that in most cases all of the digits are the same for the tunes of the two different codes and deviations are there only for a small set.

PR #693, opened 2018-11-05
More particle vector things ...

vkbo:
Fikses compilation (at least) for FLUKA after the change to 1D particle vectors in PR #691.


vkbo:
CR was broken too. Fixed now. Everything else seems to build fine.

I will merge this now since current master is broken for FLUKA and CR. Even if FLUKA is still not correct, it is at least less broken.

amereghe:
fine with me (even though late...)
One question: it seems we loose `source/include/crabkick.f90` and `source/include/acdipkick.f90`. I guess this is intentional but I don't see a replacement...

vkbo:
These files were replaced by Riccardo, but he didn't delete the old ones. We agreed to do that.

PR #695, opened 2018-11-08
Remove Multiple Machines (m, mmac, nmac)

vkbo:
Remove Multiple Machines (m, mmac, nmac)

### DIFF
--- a/source/main_cr.f90
+++ b/source/main_cr.f90
@@ -125,7 +125,7 @@ end interface
 
   ! "Old" variables
   integer i,itiono,i2,i3,ia,ia2,iar,iation,ib,ib0,ib1,ib2,ib3,id,ie,ig,ii,im,iposc,ix,izu,j,j2,jj,  &
-    k,kpz,kzz,l,ll,m,ncorruo,ncrr,nd,nd2,ndafi2,nerror,nlino,nlinoo,nmz,nthinerr
+    k,kpz,kzz,l,ll,ncorruo,ncrr,nd,nd2,ndafi2,nerror,nlino,nlinoo,nmz,nthinerr
   real(kind=fPrec) alf0s1,alf0s2,alf0s3,alf0x2,alf0x3,alf0z2,alf0z3,amp00,bet0s1,bet0s2,bet0s3,     &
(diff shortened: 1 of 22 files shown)


vkbo:
Just tagged everyone involved mostly as a FYI. You don't have to all approve it. It is mostly uncontroversial as in tracking the machine index was mostly hard coded to 1. This just reduces the dimensionality of the three arrays using this index, and the loop in maincr.

source/mod_fluc.f90
```diff
@@ -86,12 +84,6 @@ subroutine fluc_parseInputLine(inLine, iLine, iErr)
   fluc_iSeed1 = izu0
   fluc_iSeed2 = 0
 
-  if(mmac > nmac) then
-    write(lout,"(a,i0)") "FLUC> ERROR Maximum number of seeds for vectorisation is ",nmac
-    iErr = .true.
-    return
-  end if
-
```
amereghe: shouldn't we anyway leave the error message if someone is crazy enough to ask more than `mmac=1`? At least, until the `SIMUL` block will be the driving one for simulation parameters...

source/main_cr.f90
```diff
@@ -1161,7 +1135,7 @@ end subroutine envarsv
             oidpsv(ib2)=one/(one+dp1)
 ! Heavy ion variable
             moidpsv(ib2)=mtc(ib2)/(one+dp1)
-            nms(ib2)=m
+            nms(ib2)=1
```
amereghe: Shouldn't we get rid also of `nms(ib2)`? I haven't checked what this is for...

source/mod_fluc.f90
vkbo: The value isn't even read. It's skipped. Also, the variable mmac is deleted.

Or did you mean in case someone thought it still existed and tried to use it? It is described in the manual.

source/main_cr.f90
vkbo: I left it because it is used in postprocessing. I don't want to potentially start messing with the file formats.

source/mod_fluc.f90
vkbo: Added the error message back in. Made it for `mmac > 1` only, so it will accept 0 as well. Basically, if someone tries to make it deliberately more than 1, it will error.

PR #696, opened 2018-11-08
Remove clo6v/clop6v

vkbo:
Remove clo6v/clop6v

Note that this PR includes #695


rdemaria:
The general comment is that the closed orbit needs to depend on delta in case of 5D tracking. I understand that this feature was broken before, but now the new code prevent a proper implementation of this feature.

We should come back to this in the new post processing module, I have opened an issue for this #698.

vkbo:
It looks to me as if it was recalculated in the 4D/5D case before. It's the outer if statement in the imc loop. In the 6D case, closed orbit was only calculated for ib=1, but not in the 4D case.

I will have a look at this. clorb6v arrays should perhaps be kept, but scale with imc instead, not napx*imc.

rdemaria:
We decided to close and not merge, because the 5D functionality can still be used. We could review all this with the new post-processing module.

PR #701, opened 2018-11-15
refactorise code of online aperture checking

amereghe:
This PR aims at re-factoring the code of the online aperture checking, at least the part common to all kinds of studies (i.e. DA and collimation). The idea is to have an `#include` file, where:
   * we inline for all tracking routines the general aperture check;
   * the check against specific aperture types, reporting of losses (either in STDOUT or in ROOT/aperture files) and array compaction are done with specific calls, if at least a particle has been lost.


amereghe:
Overview of total test time:
sum

amereghe:
Overview of total CPU time for all `fast` tests for all the involved commits:
sum

PR #703, opened 2018-11-16
Meta and Time Modules

vkbo:
This PR adds two new modules:

**META**

The module `mod_meta` provides a way to write name/value pairs to file `sim_meta.dat` through the `meta_write` subroutine. The Subroutine has interfaces for char, real32/64/128, int16/32/64 and logicals. It has a default format for numbers, but this can be overwritten with an optional format string as a third argument.

The file format is a fixed column format with a `:` separator. This makes it easy to parse the output.

Example
```
# SixTrack Simulation Meta Data
################################################################################
SixTrackVersion                  : 5.0.2
ReleaseDate                      : 23.08.2018
GitHash                          : 923f81b282ed8fa010c8a4d7f10c7efebe9120cb
Compiler                         : gfortran
Features                         : TILT FAST STF CRLIBM CR
StartTime                        : 2018-11-16 10:56:52.568
NumParticles                     :           2
NumTurns                         :      162000
NumSingleElements                :         159
NumBlockElements                 :          59
NumStructureElements             :       12067
NumParticleTurns                 :      324010
AvgParticlesPerTurn              :           2.000
NumCheckPointRestarts            :           5
# END
```

**TIME**

The module `mod_time` keeps track of timing checkpoints (timestamps) through the simulation, and writes a report to file `sim_time.dat`. I've added a number of reasonable stamp points to the module, and inserted them into `main_cr` (and to an extent in `main_da`).

In addition, the module provides a final report of average time needed during tracking for the following: tracking time per particle (averaged over turns), per turn, per structure element, per particle-turn, and per particle-turn-element.

There is also a stop watch feature added that currently records how much tracking time is spent in the core routines of DUMP and COLLIMATION.

The file format is similar to the META data format, but also includes time units (sec, msec, usec ... etc) and in case there is a count related to the value, this is printed. This may make parsing a bit trickier, but the fixed column width should help.

Example:
```
# SixTrack Simulation Time Data
################################################################################
Internal_ZeroTime                :       0.000771 s
Stamp_AfterFileUnits             :       0.002634 s
Stamp_AfterDaten                 :       0.568586 s
Stamp_AfterClosedOrbit           :       3.347125 s
Stamp_AfterBeamDist              :       3.348741 s
Stamp_AfterInitialisation        :       3.348848 s
Stamp_AfterPreTrack              :       3.374568 s
Stamp_AfterTracking              :       6.098786 s
Stamp_AfterPostProcessing        :       6.098817 s
Stamp_BeforeExit                 :       6.098862 s
Sum_Tracking                     :       2.724218 s
Avg_PerParticle                  :      51.562486 ms
Avg_PerTurn                      :     227.018167 ms
Avg_PerElement                   :       0.094218 ms
Avg_PerParticleTurn              :    4296.873817 us
Avg_PerParticleTurnElement       :     148.608765 ns
Cost_DumpModule                  :       0.001382 s   [N = 2]
Cost_CollimationModule           :       1.493465 s   [N = 864]
Cost_ScatterModule               :       0.000000 s   [N = 0]
# END
```


vkbo:
Note: The meta module works with CR (it counts the number of restarts, and stores particle-turns between them), but the time module does not keep correct time when CRed for now.

The time module provides values useful for estimating times for BOINC. More can be added, and more derived values can be calculated from information from both these modules.

PR #704, opened 2018-11-19
Bug fixes and build changes

vkbo:
Removed some left over code in aperture that also broke nagfor build.

Re-added the NAFF build option flag as NAFF is now an external library. It needs to be initialised with `git submodule update --init lib/NAFFlib`. and if missing, the CMake will error on this.

The question remains whether the NAFF option should be ON or OFF by default. Please discuss.


vkbo:
You and @kyrsjo can fight over that. :P

vkbo:
As it is currently set up, cmake will tell you what command to run, but not run it for you.

rdemaria:
It looks like git submodule in cmake is recognized practice

https://cliutils.gitlab.io/modern-cmake/chapters/projects/submodule.html

vkbo:
So, @kyrsjo, we are leaning towards going back to the previous implementation of having cmake_six pull the necessary submodules for default build.

I don't like the idea of cmake itself modifying the source directory. We've specifically opted for an out of source build approach, and cmake itself should not do anything more than suggest how to pull the missing dependencies.

I know the cmake_six script is intended as a wrapper script only, but it is after all a helper script still, and this is where the "help" should occur, if we out it anywhere at all.

My personal preference is what this PR does. Neither cmake_six nor cmake itself should do this for you. CMake tells you precisely what to do in either case.

Note that if you try to build from a release tar.gz or zip, neither option actually works. Which is a problem.

vkbo:
Actually, there may be a compromise. We have the buildLibraries script. We can add everything there, and ensure that it also works when we try to build from tar.gz source.

the default behaviour of buildLibraries.sh is to pull everything. We could make it to by default just pull the libraries needed for standard builds, and then have cmake_six check that this has been done.

PR #712, opened 2018-11-28
Dump of particle final state

vkbo:
This PR adds a small subroutine to mod_particles that is called just before postpr in main_cr.

It opens a binary file stream and writes:
A header consisting of 8 bytes with napx and npart (2 x int32)
A record for each particle of 72 bytes consisting of llostp, nlostp (2 x int32), x,y,xp,yp,sigmav,dpsv,ejfv,ejv (8 x real64)

The file is intended as a replacement/alternative for singletrackfile.dat/fort.90 for checking the end result of a simulation.


vkbo:
I will split this into two separate PRs. Closing this.

PR #720, opened 2018-12-03
E lenses radial profile

amereghe:
E-lens module can handle a radial profile read from ASCII file. At the same time:
   * e-lens kick fully chromatic, i.e. taking into account `rvv`;
   * `lininterp` function moved out of `DYNK`;
   * added e-lens current and kinetic energy among the `DYNK`-able parameters;
   * added radial profile to tests of e-lens + brand new test during ramp, for checking `DYNK` as ramp, variation of e-lens kicks with beam energy and modulation of kicks during ramp;
   * updating manuals accordingly (user and physics) + correcting e-lens hyphenation;


amereghe:
moved to https://github.com/SixTrack/SixTrack/pull/727 as big canonicals have been removed to save space

PR #723, opened 2018-12-03
Dump final state

vkbo:
This PR adds a flag in the SETTINGS block for dumping the final state of the particle arrays (1:npart).

The options are:
`FINALSTATE binary`
`FINALSTATE text`

Interestingly, the dumped text file shows that particle ID and lost flag are not preserved for particles after they have been moved to index > napx (lost particles).


source/mainarrays.f90
```diff
@@ -162,3 +162,148 @@ subroutine expand_thickarrays(nele_request, npart_request, nblz_request, nblo_re
...
+
+      tnapx = tnapx - 1
+    end do
+  end if
+
+#ifdef FLUKA
+  if(fluka_enable) then
+    tnapx = napx
+    do j=napx,1,-1
+      if(llostp(j)) call fluka_lostpart(tnapx, j) ! Inform fluka
```
amereghe: we have to apply the same logics in `fluka_lostpart`

source/include/lostpart.f90
```diff
@@ -32,7 +32,7 @@
   if(llost) then
     ! report losses to user
     call aperture_reportLoss(n,i,ix)
-    call compactArrays
+    if(.not.apflag) call compactArrays
```
amereghe: this same `if` statement should be added also in `collimation.f90`, at `call compactArrays`

source/mainarrays.f90
vkbo: I looked at fluka_lostpart, and it looked to me that this was equivalent, but it doesn't retain the old information. I assumed that was not necessary because it is anyway overwritten.

However, I'm happy to rewrite the routine to do the same thing.

source/include/lostpart.f90
vkbo: Hmm, that causes a circular dependency in collimation. Is the apflag variable not internal to aperture? If not, I do have to move it to one of the common modules.

source/mainarrays.f90
vkbo: Also spotted a bug in that loop. `tnapx = tnapx - 1` should be inside the if statement.

rdemaria:
For boinc validator, I agree, although I would use the hash to avoid sending too much data around. I would tend to use still fort.10 if postprocessing is available because we might miss an hardware error during the post-processing.

vkbo:
The original idea was to use the HASH module to hash the final state file, but separating the two tools means any file can be hashed and the has file put into the BOINC return zip file. It can in principle be used to compare a number of files in one single small text file, including fort.10.

source/include/lostpart.f90
amereghe: ok - no worries. Actually, I've just realised that that check is necessary for losses in the aperture, not for losses in collimators

amereghe:
yes - the comment about the `fort.10` is more for saying that we should validate the final state of particles, regardless if the user wants to evaluate DA or losses.

PR #724, opened 2018-12-03
Enforce NEXT after PRINT

vkbo:
Not to be merged until we do a minor version bump. This breaks technically invalid fort.3 files that have worked anyway.

This PR removes the bypass of the NEXT check after the PRINT block statement in fort.3
If it detects that the next line is not a NEXT, it will error and ask if the NEXT was forgotten.

When a PRINT block is used properly, it still gives a warning that this will be removed at some point, and that the PRINT flag now lives in the SETTINGS block. An example block is provided.

In addition, when an error occurs in fort.3, the last five lines read from the file are printed back to make it easier to spot the error.

Example:
```
INPUT> Printout of input parameters ENABLED
INPUT> WARNING The PRINT block is deprectaed and will be removed in a future release. Please use:

SETTINGS
  PRINT
NEXT

INPUT> ERROR PRINT block does not take any parameters. Did you forget to close it with a NEXT?
INPUT> ERROR in fort.3
INPUT> Line 3: 'COMMENT'
INPUT> Previous Lines:

    1 | GEOMETRYFILE Long FODO-cell for SixTrack tests
    2 | PRINT INPUT
    3 | COMMENT

    +++++++++++++++++++++++++++++
    +      ERROR DETECTED!      +
    + RUN TERMINATED ABNORMALLY +
    +++++++++++++++++++++++++++++

STOP 1

```


vkbo:
Yeah, they were mostly the orbit_* tests, which I assume are copies of each other.

rdemaria:
For me it is important that we keep the compatibility with the DA studies because we a have a bilion of fort.3 of them

vkbo:
To be merged for SixTrack 5.1

PR #726, opened 2018-12-04
Mod HASH

vkbo:
Depends on PR #723. The commits for that PR are also included here before that is merged.

This PR includes the old md5 c 1990s code from MIT, updated with the RFC1321 hex values for constants, and modified to work in 64bit builds as well as 32bit.

A new block is available called `HASH`, where a list of `MD5SUM filename` lines can be added. At the end of `main_cr`, but before `ZIPF`, a routine is called to compute the md5sum of these files and write them into a file `hash.md5` which is a valid md5sum file.

The hashed files can also be verified with the command: `md5sum -c hash.md5`

The HASH module performs a self test during initialisation to ensure that it produces the correct reference hashes as specified in RFC1321. If this does not pass, an error message is written to the hash.md5 file instead of hash values. SixTrack will not exit because of this (but any diff with the hash file will fail if used with a test.)

The hash.md5 file can be zipped for BOINC, as well as added as canonical to be used with tests.

The hash module is enabled via the HASHLIB compiler flag. I think after some testing, and if we start using it in tests, we can set it to ON by default.

I need to update the manual before this is merged.


vkbo:
A few further changes:
* The scatter_aperture test has been modified to use the hash.md5 file for verifying all output.
* The HASHLIB flag is ON by default.

I think maybe we can remove this flag entirely though, as the modified test above passed with all supported compilers, but we can probably wait till after the nightly build has been run to verify all build options work.

kyrsjo:
Looked through the PR as @vkbo asked, looks good to me.

vkbo:
Left a roundctl dependency in the build though, so it doesn't build with single and quad. Fixing now.

PR #731, opened 2018-12-06
Rename nlostp

vkbo:
This PR renames `nlostp` to `partID`.
It also adds a global array `parentID` for later use. Is intended for use with scatter and Fluka, at least.


vkbo:
Indeed. As we discussed, it may be easier to have a new global one, and replace the internal ones when possible as we move forwards. With this global one, the scatter module will be a test case on whether it works as expected before we start re-implementing/merging this elsewhere.

vkbo:
Since this is a code structure change, I think I will hold off on the merge until after release of 5.0.4.

PR #732, opened 2018-12-06
CR Kill Switch

vkbo:
Adds a setting `CRKILLSWITCH` to the `SETTINGS` block which takes a list of turn numbers where SixTrack will stop automatically when running with CR. The kill turn is tracked in a temporary file so that it doesn't kill on the same turn twice.

The intention is to force a kill in tracking also for short tests that may not actually have the C/R feature properly tested in the current timed kills by SixTestWrapper.

The SixTestWrapper currently does not know whether SixTrack stopped via the kill switch, or exited for other reasons, so a file named `crresratme.tmp` is also written. SixTestWrapper should detect this, delete that file, restart the job.

This needs to be added. @kyrsjo ?


source/checkpoint_restart.f90
```diff
@@ -109,6 +110,75 @@ subroutine cr_expand_arrays(npart_new)
...
+
+  inquire(file="crkillswitch.tmp",exist=fExist)
+  if(fExist .eqv. .false.) then
+    open(crksunit,file="crkillswitch.tmp",form="unformatted",access="stream",status="replace")
+    write(crksunit) 0,0
+    flush(crksunit)
+    close(crksunit)
+  end if
+
+  open(crksunit,file="crkillswitch.tmp",form="unformatted",access="stream",status="old")
```
kyrsjo: Open for 'r','w','rw'?

source/checkpoint_restart.f90
```diff
@@ -109,6 +110,75 @@ subroutine cr_expand_arrays(npart_new)
...
+
+  do i=1,size(st_killturns,1)
+    if(iTurn == st_killturns(i) .and. iTurn > pTurn) then
+      killIt = .true.
+      exit
+    end if
+  end do
+
+  if(killIt) then
+    nKills = nKilld + 1
```
amereghe: typo? `nKills` instead of `nKilld`

source/checkpoint_restart.f90
```diff
@@ -109,6 +110,75 @@ subroutine cr_expand_arrays(npart_new)
 
 end subroutine cr_expand_arrays
 
+! ================================================================================================ !
+!  CR KILL SWITCH
+!  V.K. Berglyd Olsen, BE-ABP-HSS
+!  Last modified: 2018-11-16
+!  This routine will kill SixTrack if the current turn number matches a number in crkillturns
+! ================================================================================================ !
+subroutine cr_killSwitch(iTurn)
```
amereghe: If I correctly understand, the call to this routine is at each turn... maybe we can already target the call to the correct turn...

source/checkpoint_restart.f90
```diff
@@ -109,6 +110,75 @@ subroutine cr_expand_arrays(npart_new)
...
+
+  killIt = .false.
+
+  if(crksunit == -1) then
+    call funit_requestUnit("crkillswitch.tmp",crksunit)
+  end if
+
+  inquire(file="crkillswitch.tmp",exist=fExist)
+  if(fExist .eqv. .false.) then
+    open(crksunit,file="crkillswitch.tmp",form="unformatted",access="stream",status="replace")
```
amereghe: why don't you use `units_openUnit` directly?

source/checkpoint_restart.f90
kyrsjo: I fixed it by checking if we are on a kill turn before opening files etc.. Anyway this function is not called if the CRKILLSWITCH is `.true.`

source/checkpoint_restart.f90
kyrsjo: Done.

source/checkpoint_restart.f90
kyrsjo: fixed.

source/checkpoint_restart.f90
vkbo: Because it's supposed to be used for global files for flushing and closing at end of simulation, and to properly interface with BOINC. Here the tmp file is opened and closed internally in the routine every time, and the files are not needed for anything other than temporary storage.

kyrsjo:
Random request: If the CircleCI test fails (visible on ctest exit code?), could we dump the `Testing/LastTest.log` file?

vkbo:
@kyrsjo and I updated the SixTestWrapper. This now works as expected on Unix. Still need to verify on Windows before merge.
There are now two C/R counters in sim_meta.dat.
```
CR_RestartCount                  :           7
CR_KillSwitchCount               :           3
```

PR #733, opened 2018-12-07
make aperture module BOINC-compatible

amereghe:
More in detail, this PR includes:
   * parenthetisation;
   * splitting of current initialisation of aperture module on different calls (necessary to make CR effective);
   * CR aperture losses (file position from `fort.93`) and `llostp` array;
   * management of file units via `units_openUnit` (e.g. aperture losses, aperture echo, aperture x-secs (not active yet...));

Additional fixes, to make test pass:
   * CR fix in `dump`, in `dump_crpoint`:
      * `fileunit` is actually used by the `write` statement - before it was always writing to unit `95`;
      * flushing unit being written;
   * CR fix in `scatter`: seeds were correctly saved and read, but not copied from CR vars to regular vars;


source/dump.f90
```diff
@@ -1927,7 +1929,9 @@ subroutine dump_crpoint(fileunit,lerror,ierro)
   integer, intent(inout) :: ierro
   integer j
 
-  write(95,err=100,iostat=ierro) (dumpfilepos(j),j=-1,nele)
+  write(fileunit,err=100,iostat=ierro) (dumpfilepos(j),j=-1,nele)
+  endfile (fileunit,iostat=ierro)
+  backspace (fileunit,iostat=ierro)
```
vkbo: Nice catch of this bug. We've had some unidentified issues caused by this.

PR #738, opened 2018-12-08
ELENS Test Fixes

vkbo:
The elens tests failed on BOINC due to missing extra_inputs.txt files in all the tests.
This is non-critical for the release. After fix, all the tests passed just fine.

In the process of figuring out what was wrong, I also cleaned up a few things in C/R log output.

I also noticed a lot of fort.1000+ files in the run folder that shouldn't be there. I added a few flush commands here and there, but this needs looking into.


vkbo:
Ignore the back and forth merge spam. They are all empty. I didn't notice that the my command to make a branch failed, and commited to my master.

vkbo:
Yeah, I added the extra checpoints during debugging because most of them never got a restart. I just left them in anyway. It's probably a good idea in general for short tests.

PR #740, opened 2018-12-10
Cleanup uninitialised variables

vkbo:
There were a number of uninitialised variables. Most of them were non-issues in that they cases where they were uninitialised were anyway conditions that were otherwise excluded, but one of the cases produced a file full of NaN values on windows for the `wire_thin6d_ibeco0` test. This was due to the beam--beam code using uninitialised variables to compute offset when `ibeco == 0`.


source/collimation.f90
```diff
@@ -3192,7 +3192,7 @@ subroutine collimate_end_collimator()
   real(kind=fPrec) hdfx,hdfxp,hdfy,hdfyp,hdfdee,hdfs
 #endif
 
-  real(kind=fPrec) stracki
+  ! real(kind=fPrec) stracki ! stracki makes no sense here
```
amereghe: not sure about this... I think it is important to keep it, and maybe it should have been in the subroutine interface.
Is there any chance that the compilers were actually taking the value from the parent subroutine?

source/string_tools.f90
```diff
@@ -293,15 +293,16 @@ function chr_expandBrackets(theString) result(theResult)
   theBuffer = " "//theString//" "
   bPos(:,:) = 0
   lSP  = 0
+  tSP  = 0
```
amereghe: forgot `! -Wmaybe-uninitialized`

source/collimation.f90
amereghe: as discussed, this subroutine does not even need `stracki`, and anyway a non-zero long collimator as from `fort.2` is caught earlier in the code and the code will immediately exit

source/string_tools.f90
vkbo: I added those flags in places where I zeroed values without being 100% sure it was correct. This is my code, so I know what it's supposed to do.

source/collimation.f90
vkbo: Also, the `stracki` variable is never set in this routine, so the behaviour is unpredictable.

PR #743, opened 2018-12-12
SixTest support for demanding that at number of restarts have happened.

kyrsjo:


kyrsjo:
Hmm, the DA checks fail. I guess we need to modify the call to SixTestWrapper for them too...

vkbo:
Yeah, it's easy to forget the DA checks. Could you also update to most recent master?

PR #745, opened 2018-12-12
CMake updates BOINC etc

vkbo:
A few tweaks here and there, in CMake and main_cr mainly, while trying to sort out the segfault issues on debian/ubuntu with gcc/gfortran.

This PR also bumps the BOINC API to latest client version, 7.14.2.

Needs testing on Windows, both for changes here, and those from PRs #743 and #732.

### DIFF
--- a/source/mod_units.f90
+++ b/source/mod_units.f90
@@ -27,13 +27,13 @@ subroutine units_openUnit(unit,fileName,formatted,mode,err,status,recl)
 
   implicit none
 
-  integer,                    intent(in)    :: unit
-  character(len=*),           intent(in)    :: fileName
-  logical,                    intent(in)    :: formatted
-  character(len=*),           intent(in)    :: mode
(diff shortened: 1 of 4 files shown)


source/mod_units.f90
```diff
@@ -122,37 +122,40 @@ subroutine units_openUnit(unit,fileName,formatted,mode,err,status,recl)
...
     err = .true.
     write(lout,"(a,i0)") "UNITS> File '"//trim(fFileName)//"' reported iostat = ",ioStat
-  else
-    err = .false.
   end if
   return
 
+10 continue
+  err = .true.
+  write(lout,"(a)") "UNITS> File '"//trim(fFileName)//"' reported an error"
```
kyrsjo: Why not `prror` if the err flag is not given? Also why not write to stderr instead/in addition to lout?

vkbo: Yes, but not for this PR.

PR #749, opened 2018-12-17
remove allocation of dum in sbc in beam6d

rdemaria:
sbc is called each lens, each turn and each particles, however a temporary array has been declared allocatable and therefore allocated and deallocated many times. Alessio figured it out, during tests on boinc, since the file mem_alloc.log grew to 200 MB until being killed by boinc.


vkbo:
While it is unnecessary to use alloc for a 13 element array like this (especially as it has an overhead), we could also disable the mod `mem_alloc.log` in general as it too has an overhead. During the development and rewrite into dynamically allocated npart, nele, etc, James used it for debugging. It used to write to fort.6, and I diverted it to a separate file.

We could just comment out all the code related to it, or wrap it in `#ifdef DEBUG`.

vkbo:
`mod_alloc` still keeps track of the maximum amount of memory allocated by the module, which is now written to the `sim_meta.dat` as well as the file James created for this. That should be enough to keep track of what goes on in the module.

I can make a separate PR on this. It will take me 5 minutes.

amereghe:
Fine with me to make the other PR to have `mem_alloc.log` produced under `#ifdef` pre-processing flag.
...or maybe it can be triggered by the `SETTINGS` block...
Thanks for the PR and merging!

vkbo:
It can't be triggered by the debug flag as the common_modules depend on mod_alloc, creating a circular dependency. But that is fixable by slitting up common_modules. I prefer to keep that file as it is because it is cleaner to keep them all in one place.

PR #751, opened 2018-12-19
Boinc Sixin.zip fix

vkbo:
Tests with BOINC build failed due to extraction of Sixin.zip not triggering. Issue resolved.

In the process, I also clarified an error message in daten, moved the opening of fort.2 and fort.3 there, and came across an issue with restarting aperture_losses file when running CR.

Note: file unit variables should never be initialised as 0, because 0 is the error unit. They should always be initialised as -1, which will cause a critical error if uninitialised when writing occurs. I.e. easier to trap.

### DIFF
--- a/source/sixtrack.f90
+++ b/source/sixtrack.f90
@@ -236,7 +236,7 @@ subroutine daten
 
   call f_open(unit=3,file="fort.3",formatted=.true.,mode="r",err=fErr)
   if(fErr) then
-    write(lout,"(a)") "INPUT> ERROR Could not open fort.2"
+    write(lout,"(a)") "INPUT> ERROR Could not open fort.3"
     call prror
   end if
 
--- a/source/sixtrack.f90
+++ b/source/sixtrack.f90
@@ -234,7 +234,7 @@ subroutine daten
 ! ================================================================================================ !
 
 90 continue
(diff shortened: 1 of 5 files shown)


source/sixtrack.f90
```diff
@@ -233,11 +234,17 @@ subroutine daten
 !  READ FORT.3 HEADER
 ! ================================================================================================ !
 
+  call f_open(unit=3,file="fort.3",formatted=.true.,mode="r",err=fErr)
+  if(fErr) then
+    write(lout,"(a)") "INPUT> ERROR Could not open fort.2"
```
amereghe: it should be `write(lout,"(a)") "INPUT> ERROR Could not open fort.3"`

vkbo: Good catch. That's a typo. They put 2 and 3 too close to each other on the keyboard!

vkbo:
@amereghe 
Yeah, the aperture_losses error occurred when the CR routine tried to open the file as unit 0 when restrating because the f-requestUnit call was made after the repositioning of the file. With the new version of mod_units, it's safe to call f_requestUnit before any f_open call, even if a unit has already been assigned.

Also, fixed the typo. Could you approve it?

PR #754, opened 2018-12-20
Fixing parsing file of aperture

amereghe:
This PR fixes issues with parsing aperture file from `LOAD` statement, so that:
   * any keyword in the `LIMI` block is no longer ignored, as it was the case so far;
   * the code is cleaner.

### DIFF
--- a/source/aperture.f90
+++ b/source/aperture.f90
@@ -2518,7 +2518,7 @@ subroutine aper_parseLoadFile(load_file, iLine, iErr)
 
   implicit none
 
-  character(len=16),intent(in)    :: load_file
+  character(len=64),intent(in)    :: load_file
   integer,          intent(in)    :: iLine
   logical,          intent(inout) :: iErr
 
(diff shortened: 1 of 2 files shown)


source/aperture.f90
```diff
@@ -2582,6 +2580,7 @@ subroutine aper_parseInputLine(inLine, iLine, iErr)
   logical,          intent(inout) :: iErr
 
   character(len=:), allocatable   :: lnSplit(:)
+  character(len=16)               :: load_file
```
rdemaria: Can we have a large or len=* for the load_file variable?

vkbo: mod_units accepts file names up to 64 characters. Maybe a good choice. 16 is a bit small.

amereghe: will do that. 16 was the initial choice, but 64 is fine as well.
Shall we then extend mod_units to accept `character(len=:)`?

amereghe: ok done to 64 chars
Shall we change mod_units such that we go to 255, as for BOINC?

vkbo: No, the limitations in mod_units have to stay as a fixed number. We can extend it beyond 64 characters, but I don't think it's a good idea. We should not use len=: either. The storage array is a type (struct), and having them fixed size makes it easier to handle in memory I would think.

There is a maximum length of file paths in Windows, which is why Boinc is fixed to the value it is. That is for the entire file path, so I found it reasonable to allow the filename to be a quarter of that. 64 characters should be more than enough.

PR #756, opened 2018-12-31
Generalized RF-Multipoles

tpersson:
This implements the generalized RF-Multipoles. 

The documentation is updated and a test is added. It is also implemented so it should work with DYNK. It has not been tested with misalignment but should in principle work.

In the cross check with the old implementation 3 errors were found in the FOX part of the implementation of the RF-multipoles (no errors in the tracking part)


tpersson:
I forgot to update the tests. Will fix that now.

vkbo:
Seems you also broke the DA version?

source/sixtrack_input.f90
```diff
@@ -1836,6 +1839,101 @@ subroutine sixin_parseInputLineMULT(inLine, iLine, iErr)
 
 end subroutine sixin_parseInputLineMULT
 
+
+subroutine sixin_parseInputLineRF_MULT(inLine, iLine, iErr)
```
vkbo: Please call the routine `sixin_parseInputLineRFMU` so it matches the block name. This is how the others are implemented.

test/CMakeLists.txt
```diff
@@ -53,6 +53,8 @@ set(TESTS_NOFORT10
   dump4
   dump5
   dump6
+  fileunits_DYNK_DUMP
```
vkbo: This test has been deleted. Please don't add this line back in.

source/include/dainicom.f90
```diff
@@ -14,4 +14,4 @@
 !FOX  D V DA EXT DELTAS NORD NVAR ;
 !FOX  D V DA EXT KCRABDA NORD NVAR ;
 !FOX  D V DA EXT MOIDA NORD NVAR ;
-!FOX  D V RE INT MTCDA ;
+!FOX  D V RE INT MTCDA ;
```
vkbo: Please keep the \n at the end of the file.

source/sixtrack_input.f90
```diff
@@ -1836,6 +1839,101 @@ subroutine sixin_parseInputLineMULT(inLine, iLine, iErr)
...
+  real(kind=fPrec) namp0,nphase0,samp0,sphase0, freq0
+  integer          nSplit,i,nmul,iil
+  logical          spErr
+
+  save nmul,iil
+   
+
+  call chr_split(inLine, lnSplit, nSplit, spErr)
+  if(spErr) then
+    write(lout,"(a)") "MULT> ERROR Failed to parse input line."
```
vkbo: These lines should probably start with "RFMU>" so the user knows which block it is.

vkbo:
Also, when adding new tests, please consider using the finalstate.dat file instead of DUMP files. If the files are large, also use the HASH module and commit the HASH file as a canonical instead of the large DUMP files.

The usage is described in the manual.

In this case, the files (that I've checked) are small, so it's not necessary to use the HASH module.

PR #758, opened 2019-01-09
Rfmult bugfixes

vkbo:
PR #756 did not build on intel due to the proprocessor handling includes slightly differently than the GNU compiler. This PR fixes this.

In addition a bit of cleanup was necessary in the code. Write messages had inconsistent tags, wrong indentation, and extra spaces at end of lines.

~~The test added in #756 also produces a different result with nagfor than the other compilers, and therefore fails the test on singletrackfile.dat. Still needs to be fixed.~~

This was due to sin and cos being called instead of sin_mb and cos_mb. Fixed.


vkbo:
Also, I noticed that envarsv is called several places inside particle loops. This subroutine already loops over all particles, and should be called outside of the particle loop. I will fix this in a separate PR together with some other performance issues I want to address.

PR #759, opened 2019-01-10
Energy update changes

vkbo:
This changes how particle energy is handled in crabs and RF. That is, we now call the part_updatePartEnergy routine, which now also handles update of particle angles when momentum changes. This is now an optional flag to the routine.

This PR also moves the call to `envarsv` out of particle loops (it already loops over particles internally).

This PR also fixes a potential bug with a remaining hard coded file unit in the elens module.

There are also some minor improvements to the logging in mod_hash and mod_units.

This PR changes a number of tests because the part_updatePartEnergy computes one of the particle arrays slightly differently than in tracking. This was always implemented in different ways in the code, so a set of tests would have changed either way. The differences are in a few decimals.

While updating the tests, all the ones that tracked large singletrackfile.dat canonicals were replaced with final_state.dat files in text format to avoid bloating the repository further.


rdemaria:
Is there any case in which you call part_updatePartEnergy(1,.false.)?

vkbo:
Yes, both when updating reference energy and during initialisation (reading particle distributions).

rdemaria:
Ok, but then you never need the if statement and the flag can absorbed in
case 1, right?

source/mod_particles.f90
```diff
@@ -146,6 +159,11 @@ subroutine part_updatePartEnergy(refArray)
   omoidpsv(1:napx) = ((one-mtc(1:napx))*oidpsv(1:napx))*c1e3
   rvv(1:napx)      = (ejv(1:napx)*e0f)/(e0*ejfv(1:napx))     ! Beta_0 / beta(j)
 
+  if(doUpdateAngle) then ! Update particle angle, but only if energy change during tracking
+    yv1(1:napx)    = (ejf0v(1:napx)/ejfv(1:napx))*yv1(1:napx)
+    yv2(1:napx)    = (ejf0v(1:napx)/ejfv(1:napx))*yv2(1:napx)
+  end if
+
```
amereghe: comment in this subroutine are a bit misleading - you update in the subroutine particle angles if explicitly requested, but the request in general is done only if the energy changes during tracking, i.e. your flag here does not remind that it is true only when energy is changed - and strictly speaking there is nothing checking that this is true

source/mod_hash.f90
```diff
@@ -116,10 +110,8 @@ subroutine hash_initialise
     tStatus = "Failed"
     hash_selfTestOK = .false.
   end if
-  if(st_debug) then
-    write(lout,"(a)") "HASH>   #3 : "//md5Digest//" < 'abcdefghijklmnopqrstuvwxyz'"
-    write(lout,"(a)") "HASH>    3 : "//md5Valid//" > "//tStatus
-  end if
+  ! write(lout,"(a)") "HASH>   #3 : "//md5Digest//" < 'abcdefghijklmnopqrstuvwxyz'"
+  ! write(lout,"(a)") "HASH>    3 : "//md5Valid//" > "//tStatus
```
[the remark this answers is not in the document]
vkbo: They're development debugging info. The result of the fail test is reported in the final file, so it is not necessary for runtime.

I decided also to remove it from the global debug flag if statement because this is printed during initialisation, which causes a few issues as this is called very early in maincr.

source/mod_particles.f90
vkbo: The comment is just outdated. There used to be a second element in that if statement. Will update.

PR #762, opened 2019-01-11
Minor bug

vkbo:
aper_parseInputLine is now recursive and needs to be labelled as such.


amereghe:
why?

vkbo:
Because you changed the routines, and they now call each other. Therefore the compiler sees this as potentially recursive, even though it isn't in practice.

amereghe:
oh - good catch! but why do we need to declare only one as recursive?

vkbo:
Only when running Debug build with gfortran. It is a lot more picky and exits on this.

vkbo:
Also, not so much a "good catch" as a red "failed" on the test build system :P

amereghe:
ok, great, thanks

PR #763, opened 2019-01-14
Missaligned rf-multipoles

tpersson:
This PR implements the misalignment of the RF-multipoles. 

The old test had also to be updated due to a numerical difference. The change that caused this was:
yv2(j) = yv2(j) + (((-REAL(Cp0))*tilts(i)) + ((AIMAG(Cp0)))*tiltc(i))*(moidpsv(j)*c1e3) and is due to the change in the order of operations. 

The export from MAD-X has also been implemented and I will update when this has been included in the master branch there (should be today).


source/include/rfmulti.f90
```diff
@@ -30,9 +30,10 @@
 
   Sp1 = Sp1 * (x_t+imag*y_t);
 
-  yv1(j) = yv1(j) - ((real(Cp0)*c1e3)*moidpsv(j))
-  yv2(j) = yv2(j) + ((aimag(Cp0)*c1e3)*moidpsv(j))
-  ejv(j) = ejv(j) - ((real(Sp1)*(c1e3*(e0f*(crabfreq*(two*pi))))))/clight
+  yv1(j) = yv1(j) - (((REAL(Cp0))*tiltc(i))  + ((AIMAG(Cp0)))*tilts(i))*(moidpsv(j)*c1e3)  
+  yv2(j) = yv2(j) + (((-REAL(Cp0))*tilts(i)) + ((AIMAG(Cp0)))*tiltc(i))*(moidpsv(j)*c1e3)
+  ejv(j)   = ejv(j) - ((real(Sp1)*(c1e3*(e0f*(crabfreq*(two*pi))))))/clight
+  ejf0v(j) = ejfv(j)
```
vkbo: The line `ejf0v(j) = ejfv(j)` is redundant. This is already handled in `part_updatePartEnergy` when using `updateAngles = .true.`

test/generalized-rf-multipoles-missalignments/fort.3
```diff
@@ -0,0 +1,74 @@
+GEOMETRYFILE Long FODO-cell for SixTrack tests
+PRINT INPUT
```
vkbo: Please use the `SETTINGS` block in new tests, so that global debug can be enabled. All new tests should have:

```
SETTINGS
    PRINT
    DEBUG
NEXT
```

at the top instead of the `PRINT` block.

PR #766, opened 2019-01-15
Updates for fort.12 and fort.13

vkbo:
This PR:

* Changes fort.13 to prioritise energy over delta (delta is recomputed from energy). This changes a few tests that uses fort.13 as input file.
* Adds a routine that writes fort.12 using the roundctl library. The old code wrote all particle pairs up to napxo, so including lost particles. That is still the case, but I'm not sure if that is the best approach?


vkbo:
To be merged for 5.2 due to changes in input and output files.

rdemaria:
We need to fix also BB 6D because it updates momentum (and I am not sure it updates all the variables)

vkbo:
Yeah, I'm aware, but we should do that in a separate PR.

vkbo:
After discussing this with @rdemaria, @kyrsjo and @amereghe, I've now disabled writing of fot.12. The write routine is still there for the time being, and this can be re-enabled by adding `call writeFort12` to `maincr`.

It also turned out that the particle pair dump after tracking in fort.6 contained wrong information. This has now been fixed.

Due to the particle shuffle routine we added earlier, the xvl, yvl, dpsvl, ejfvl and ejvl routines are redundant. They've been deleted. In addition the ixv array was unused, so I deleted that too.

vkbo:
Turns out some people are using fort.12 after all, although it has been writing the wrong particle IDs for a while for lost particles. In any case, it will now default to write fort.12 if there are up to 64 particles, and default to not write the file if there are more. This can be overwritten with the WRITEFORT12 flag in the SETTINGS block.

PR #769, opened 2019-01-16
Simulation block

vkbo:
This is the beginning of the SIMU block in fort.3.
Making the PR now so we can discuss the implementation, keywords, etc.

Implemented so far:
* The relevant parameters from the TRAC block.


kyrsjo:
Why was it closed?

PR #775, opened 2019-01-29
Tests Cleanup

vkbo:
This PR cleans up a bunch of tests with very large canonicals. This reduces the test folder size by about 800 MB!

The canonicals in the 5 tests affected are now using the hash module instead. Some minor changes to the hash module has been made too.


source/sixtrack.f90
```diff
@@ -5635,7 +5630,7 @@ subroutine corrorb
       ivflag=0
       icflag=0
 
-      write(lout,*)
+      write(lout,"(a)") ""
       write(lout,10000)
```
kyrsjo: What is FORMAT 10000 ? Please add a comment or make it explicit.

kyrsjo: There are a couple of more FORMATs in this routine, but cleaning them up should probably be done in another PR.

source/track_thick.f90
```diff
@@ -656,7 +657,7 @@ subroutine thck4d(nthinerr)
         end if
         if(fluka_inside) then
           if(fluka_debug) then
-            write(lout,*) '[Fluka] Skipping lattice element at ',i
+            write(lout,"(a,i0)") "FLUKA> Skipping lattice element at ",i
```
kyrsjo: Make sure to give a heads-up to @amereghe about this change, in case of scripts that are `grep`ing.

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@
 ###################################################################################################
 
 # Set the cmake version required
-cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.3 FATAL_ERROR)
```
kyrsjo: Mention this and why in the changelog

test/CMakeLists.txt
```diff
@@ -255,7 +255,6 @@ if(CR)
     dynk_random
     scatter_aperture
     scatter_bbelastic
-    dump4
```
kyrsjo: ?

source/sixtrack_input.f90
```diff
@@ -751,8 +750,8 @@ subroutine sixin_parseInputLineSTRU(inLine, iLine, iErr)
     return
   end if
 
-  if(nSplit > 40) then
-    write(lout,"(a)") "GEOMETRY> ERROR Structure input line cannot have more then 40 elements."
+  if(nSplit < 1 .or. nSplit > 40) then
```
kyrsjo: No empty lines allowed here? Comment-only lines may be pretty useful in the GEOM block...

PR #777, opened 2019-01-31
Final State w/ions

vkbo:
This PR adds the `ions` flag to the `INITIALSTATE` and `FINALSTATE` flags in the `SETTINGS` block. This means the additional ion columns are dumped to these files too.

The tests elensidealthin6d_ions, thin4d_ions and thin6d_ions have been updated to test and use these files instead of singletrackfile.dat.


rdemaria:
Why not always adding ions variables?

vkbo:
Because they take up more space in the files. Since these files are primarily for verifying tests, I though it was best to keep size as small as possible -- when possible.

rdemaria:
I think the files can be used for other purposes as well. I agree it save space, but to me it seems a bit a complication for the user. An option would be to check if all extra variables are ones and we don't save  them. But it is also ok if you leave it like this.

rdemaria:
I think the files can be used for other purposes as well. I agree it save space, but to me it seems a bit a complication for the user. An option would be to check if all extra variables are ones and we don't save  them. But it is also ok if you leave it like this.

vkbo:
I think having things switch on and off automatically is a lot more confusing than a simple flag. Sometimes you want to dump those to check that they are what you expect.

In any case, these files are not really there for the production use, but for the test suite and dev debug. This is why they're in the SETTINGS block with the other debug options. If we want to improve dumping of the beams we should do that in DUMP.

rdemaria:
Ok

doc/user_manual/chGeneralInput.tex
```diff
@@ -54,8 +54,8 @@ \section{Settings} \label{STSett}
...
     \texttt{WRITEFORT12} & Enable or disable the writing of \texttt{fort.12} after tracking. The flag takes an optional parameter to explicitly state whether it is ON or OFF. If omitted, it is assumed the user requests it to be ON. If the flag is omitted entirely, the default behaviour is determined by the particle count. If SixTrack is running with 64 particles or less, it is ON by default, otherwise OFF.\index{fort.12}\\
-    \texttt{INITIALSTATE} & Followed by either ``binary'' or ``text''. This will write a file before tracking containing the initial state of all particles to either a binary or a text file.\index{initial state}\\
-    \texttt{FINALSTATE} & Followed by either ``binary'' or ``text''. This will write a file after tracking containing the final state of all particles, including those lost during tracking, to either a binary or a text file.\index{final state}\\
+    \texttt{INITIALSTATE} & Followed by either ``binary'' or ``text''. This will write a file before tracking containing the initial state of all particles to either a binary or a text file.\index{initial state} Adding ``ions'' as a second keyword will also dump the additional ion columns.\\
```
amereghe: maybe just remind which are the ion columns

doc/user_manual/chGeneralInput.tex
```diff
@@ -54,8 +54,8 @@ \section{Settings} \label{STSett}
...
+    \texttt{FINALSTATE} & Followed by either ``binary'' or ``text''. This will write a file after tracking containing the final state of all particles, including those lost during tracking, to either a binary or a text file.\index{final state} Adding ``ions'' as a second keyword will also dump the additional ion columns.\\
```
amereghe: same comment here as before

source/sixtrack_input.f90
```diff
@@ -420,15 +434,29 @@ subroutine sixin_parseInputLineSETT(inLine, iLine, iErr)
...
       write(lout,"(a,i0)") "INPUT> Particle final state will be dumped as a binary file"
-    else
+    case(2)
       write(lout,"(a,i0)") "INPUT> Particle final state will be dumped as a text file"
-    end if
+    case(3)
+      write(lout,"(a,i0)") "INPUT> Particle final state will be dumped as a binary file with ion data included"
+    case(4)
+      write(lout,"(a,i0)") "INPUT> Particle final state will be dumped as a text file with ion data included"
+    end select
```
amereghe: I am not a big fan of variables encoding multiple states of different logical flags... I would have had two boolean vars instead. But this is I guess rater a problem of taste...

doc/user_manual/chGeneralInput.tex
vkbo: I think this is better done, if needed, by linking to the IONS block. Otherwise there are multiple places to maintain this information.

Also, the columns in the (text) file do have headers.

amereghe: ok let's go with the link to the `IONS` block

vkbo: Done

PR #783, opened 2019-02-19
Longer Filenames

vkbo:
Some minor modifications to mod_units to allow for arbitrary length file names.

This is mostly related to PR #776 which needs longer names. But the current limitation may be an issue for any input file specifying long paths outside of the simulation folder.

I've also renamed the constant mFNameLen to mFileName, and added the constant mPathName. These are set to 64 ad 255 respectively.


source/string_tools.f90
```diff
@@ -256,6 +257,34 @@ subroutine chr_scanString(theString, theMask, maxLen, nValues, nIndent, hasErr)
...
+  iE = len_trim(pathName)
+  do iC=iE,1,-1
+    if(pathName(iC:iC) == "/" .or. pathName(iC:iC) == "\") then
+      iS = iC + 1
+      exit
+    end if
+  end do
+
+  if(iE-iS > mFileName) then
+    fileName = pathName(iS:iS+mFileName)
```
amereghe: Here you have a filename truncation - don't you want to raise an error message and stop processing?

vkbo: No, actually, I don't want to cause an error. The function is only used for the log file.

There is a point that someone else may want to use it to extract the file name though, in which case a truncated file name can be dangerous. Maybe we should return an error value at least so this can be parsed.

amereghe: that's exactly the case I was thinking about... there is no explicit warning in the function which limits its use to the log file (even if grepping through the code will pop up immediately the only concerned line), hence a user may want to re-use that function - since it is also located in a utility module :)

vkbo:
Ok, changed the logic a bit. The only purpose of that function was to make sure the log wasn't too cluttered. Found a different solution.

Also found a bug in how DUMP checks the length of element names.

PR #785, opened 2019-03-13
Pi to numerical_constants

vkbo:
Moved the four variables based on pi to numerical_constants and changed them to parameters.

Not sure why this wasn't done when the many definitions of pi were cleaned up.


rdemaria:
just a question, why was rad replaced by radn?

vkbo:
> just a question, why was rad replaced by radn?

Name conflict with lielib common variables when moving `rad` to `numerical_constants`

rdemaria:
Thanks!

PR #788, opened 2019-03-20
Some CMake updates and a bit of cleanup

vkbo:
Changes:

* Uses the CMake built in feature to parse the version string instead of the manual regex+if approach
* Extracts the numerical version from `version.f90` instead of assembling it from the version number
* Sets a generator if none is set. This makes it a bit simpler to build on MinGW/Windows as you don't have to specify `-G "Unix Makefiles"`
* Prints a bit more build info
* Also fixed a bug where if we end up with default fortran compiler (f90) the messages printed went to stderr, which stopped the cmake_six script (non-empty stderr file)

The PR was prompted by an attempt to fix a warning with cmake 3.13 when trying to use tab completion with cmake. The warning is related to source directory not being specified, but it seems to be an issue with the tab completion implementation rather than our cmake script.

Will check this again when I have cmake 3.14 installed when the warning will become a fatal error.


CMakeLists.txt
```diff
@@ -1316,19 +1310,19 @@ IF(BUILD_TESTING)
...
-SET(CPACK_PACKAGE_VERSION_MINOR ${SixTrack_VERSION_MINOR})
-SET(CPACK_PACKAGE_VERSION_PATCH ${SixTrack_VERSION_PATCH})
-INCLUDE(CPack)
+#set(CPACK_GENERATOR RPM DEB STGZ TGZ TBZ2 TXZ ZIP 7Z)
+set(CPACK_GENERATOR RPM DEB TGZ ZIP) #Take only the most common targets
+#set(CPACK_PACKAGE_CONTACT "sixtrack.support@cern.ch") # non-existent email address
+set(CPACK_PACKAGE_VERSION ${SixTrack_VERSION})
+set(CPACK_PACKAGE_VENDOR "CERN")
+set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "SixTrack symplectic particle tracking code for particle accelerators.")
+set(CPACK_PACKAGE_VERSION_MAJOR ${${PROJECT_NAME}_VERSION_MAJOR})
```
rdemaria: why ${PROJECT_NAME} is  used here only?

vkbo: Should be used for line 1319 too I see, but this bit of code doesn't work anyway.

JamesMolson: Works for me (mostly).

JamesMolson: well one line needs uncommenting for the .deb generation to work.

vkbo: Yeah, I couldn't get the deb one to work properly, but I checked that it sets package version correctly anyway.

rdemaria: Which line is upsetting deb creation?

JamesMolson: CPackDeb: Debian package requires a maintainer for a package, set
  CPACK_PACKAGE_CONTACT or CPACK_DEBIAN_PACKAGE_MAINTAINER

rdemaria: We can use  sixtrack.contact@cern.ch or sixtrack.support@cern.ch. The emails are redirected to my account.

PR #793, opened 2019-03-21
Nagfor bug fixes for 5.2

vkbo:
Fixed bug in aperture with write format 'l' instead of 'l1'


vkbo:
It took 10 minutes, no problem :)

PR #794, opened 2019-03-22
Fixes for 5.2

vkbo:
Fixes `DEBUG` build (compiler flag)
Fixes an uninitialised error flag in `scatter_crpoint`

### DIFF
--- a/source/mod_alloc.f90
+++ b/source/mod_alloc.f90
@@ -13,7 +13,7 @@ module mod_alloc
   implicit none
 
 #ifdef DEBUG
-  integer, public, save :: alloc_log ! Logfile Unit
+  integer, public, save :: alloc_log = 1000 ! Logfile Unit
 #endif
 
   ! The total quantity of memory allocated (bits)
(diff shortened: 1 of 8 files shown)


source/mod_alloc.f90
```diff
@@ -189,20 +187,6 @@ subroutine print_alloc(ename, type, requested_bits)
...
-  call f_requestUnit("maximum_memory_allocation_mbytes.txt",memunit)
-  call f_open(unit=memunit,file="maximum_memory_allocation_mbytes.txt",formatted=.true.,mode="w")
-  write(memunit,"(f10.3)") real(maximum_bits,real64)/real(mbyte,real64)
-  flush(memunit)
-  call f_close(memunit)
-#else
-  return
-#endif
-end subroutine alloc_exit
-
```
amereghe: Is the max memory usage saved somewhere else?

vkbo: Yes, in the `sim_meta.dat` file.

PR #798, opened 2019-03-26
Collimation Dist Module (Step 3)

vkbo:
This PR moves all the do_thisdis dependant distribution generation subroutines to a new module, and cleans up the code. There was A LOT of code duplication in there!

Since the code is no longer duplicated, the order of which random numbers are used in the heneration of distributions has changed and the distributions are therefore different, but should be statistically the same. Before I moved the subroutines I added tests for all dist types. Still need to plot the before and after just to make sure they are the same.

This PR includes the Step 1 and Step 2 pull requests, so until those are merged, everything shows up in the diff here.


vkbo:
Closing because I forked the branch to revert some changes. Will make a new PR.

PR #803, opened 2019-04-02
Bugfixes

vkbo:
Fixes these bugs:
* Extended indices for `elpos` array from `1:nblz` to `0:nblz+1` to match `dcum`.
* Fixes `orglat` so that also structure names are reordered when the lattice has a `GO` keyword. This needed some change in the routine calculating s-ccordinates as well.
* Fixed a bug in `plato_seq.f90` where the comment states in `tuneffti` that it will not accept frequency indices of 0 or 1, but does anyway and returns a value NaN. This causes a segfault on Debug type builds. The routine now returns a one instead in these cases. A similar fix was applied to `tunebt2`.
* Removed two `close(6)` calls in `abend` that caused an abort with BOINC when building with nagfor.


source/mod_geometry.f90
```diff
@@ -733,6 +734,14 @@ subroutine geom_calcDcum
     dcum(i) = tmpDcum
   end do
 
+  ! Correct the elpos array after a GO reshuffle
+  if(kanf > 1 .and. strumcol) then
+    sGo  = elpos(1)
```
amereghe: don't you want to use `elpos(0)`?

PR #804, opened 2019-04-03
Remove imc (variation of momentum offset)

vkbo:
This removes the variation of momentum offset feature (the flag `imc` in `fort.3`).

As a bonus, a lot of arrays with one dimension being `npart` could now be reduced in rank. In total, this frees up 85*8=680 bytes of memory per tracked particle.


kyrsjo:
I looked through, and it all looks good.

vkbo:
Ok. Will merge after 5.2.1.

PR #810, opened 2019-04-08
Added MEMUSAGE flag and code to mod_meta

vkbo:
This PR adds a `MEMUSAGE` flag to CMake that, when ON, and when building with gfortran and on Unix, will extract the process id, and peak memory and high water mark values from /proc/PID/status.

The values are written to `sim_meta.dat`.
The information will be extracted by the nightly build script.

Example:
```
Exec_PID                         :       25834
Exec_VmHWM[MiB]                  :          83.590
Exec_VmPeak[MiB]                 :         112.324
```


JamesMolson:
good thing /proc is a "unix" standard...

kyrsjo:
BSD doesn't count, it's GNU :P

And Hurd I don't care...

kyrsjo:
Apparently even BSD has a procfs. According to Veronica it should fail gracefully if the file is missing (i.e. on OSX).

But if you could test (very soon, should be merged before the nightly tests kick off)?

vkbo:
Doesn't matter if it's a standard as it's not a production feature. It will just return a -1 if it cannot fint the file or parse it.

PR #819, opened 2019-04-10
Beam dist: bugfix and script

vkbo:
This PR adds a plotting script for initial_state and fila_state files in devtools.

While testing it, I also found a bug in collimation where the updating of energy-dependant routines were called with the `j` index rather than the `i` index used in that particular loop. The index `j` is here used for sample number, which is always 1. Consequently, only particle 1 was re-computed correctly when the internal collimation dist generator was used. This is of course recomputed as soon as an element which changes particle energy is encountered.

Additionally, the PR moves the collimation init routine to main_cr, just before tracking. The begin sample routine is called from the init routine, and the end sample routine is called at the top of the collimation exit routine. Since the sample feature is disabled, this is equivalent, but has the added bonus that the initial state file contains the distribution generated by the collimation module.


source/track_thin.f90
```diff
@@ -456,14 +456,7 @@ subroutine trauthin(nthinerr)
...
       write(lout,"(a)") "TRACKING> Calling thin6d subroutine"
       write(lout,"(a)") ""
-      if (do_coll) then
-        call collimate_init()
-        call collimate_start_sample(1) ! Changed to only do 1 sample
-      endif
       call thin6d(nthinerr)
-      if (do_coll) then
-        call collimate_end_sample(1) ! Changed to only do 1 sample
-      endif
```
amereghe: I would have left these calls here, actually... I think it's cleaner

source/main_cr.f90
```diff
@@ -1321,6 +1322,10 @@ end subroutine envarsv
 
   ! Initialise Modules
   call dump_initialise
+  if(iclo6 > 0 .and. ithick == 0 .and. do_coll) then
```
amereghe: not sure this is the best check in this position...
I guess we should have a general check after `daten` that exits the program if this condition is not satisfied

source/collimation.f90
```diff
@@ -1743,34 +1747,21 @@ subroutine collimate_start_sample(nsample)
...
-    rvv(j)=(ejv(j)*e0f)/(e0*ejfv(j))
-    dpsv(j)=(ejfv(j)*(nucm0/nucm(j))-e0f)/e0f
-    oidpsv(j)=one/(one+dpsv(j))
-    moidpsv(j)=mtc(j)/(one+dpsv(j))
-    omoidpsv(j)=c1e3*((one-mtc(j))*oidpsv(j))
-    dpsv1(j)=(dpsv(j)*c1e3)*oidpsv(j)
-
-    partID(i)=i
-    parentID(i)=i
+  call part_updatePartEnergy(1)
```
amereghe: I think this should go inside the above check

source/main_cr.f90
vkbo: Added the check to the exit call as well, as we discussed.

source/track_thin.f90
vkbo: As we discussed, the reason this was moved to main_cr is so it's called before pre-track. This is necessary because the collimation module generates beam distribution, which must be before the initial_state file is written.

source/collimation.f90
```diff
@@ -1743,34 +1747,21 @@ subroutine collimate_start_sample(nsample)
...
 
-    do ieff =1, numeff
+  do i=1,napx00
```
kyrsjo: Unneeded variable, but apparently removed in a later PR.

vkbo: Huh? It's not unneeded, but there is no difference between `napx` and `napx00`, so `napx00` can be replaced.

PR #821, opened 2019-04-10
Minor cleanup

vkbo:
Remove a few global arrays that were only used one place, and removed the allocate thickarrays routines from the common modules as the expand thickarrays routines do the same thing.

### DIFF
--- a/source/sixve.f90
+++ b/source/sixve.f90
@@ -148,8 +148,8 @@ subroutine blocksv
 
 #ifdef FLUKA
   ! Entirely re-initialise to 0.0 hv(...) and bl1v(...) arrays
-  hv(:,:,:)   = zero
-  bl1v(:,:,:) = zero
+  hv(:,:,:)     = zero
+  bl1v(:,:,:,:) = zero
 #endif
   do ia=1,napx
     do k=1,mblo
--- a/source/sixve.f90
+++ b/source/sixve.f90
@@ -143,20 +143,9 @@ subroutine blocksv
   implicit none
 
   integer ia, ikk, j, jm, k, lkk, mkk
-  real(kind=fPrec) dpoff
-!     integer i,itiono,i1,i2,i3,ia,ia2,iar,iation,ib,ib0,ib1,ib2,ib3,id,&
-!    &idate,ie,ig,ii,ikk,im,imonth,iposc,irecuin,itime,ix,izu,j,j2,jj,  &
-!    &jm,k,kpz,kzz,l,lkk,ll,m,mkk,napxto,ncorruo,ncrr,nd,nd2,ndafi2,    &
-!    &nerror,nlino,nlinoo,nmz,nthinerr
-!     double precision alf0s1,alf0s2,alf0s3,alf0x2,alf0x3,alf0z2,alf0z3,&
-!    &amp00,bet0s1,bet0s2,bet0s3,bet0x2,bet0x3,bet0z2,bet0z3,chi,coc,   &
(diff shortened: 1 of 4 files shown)


source/sixve.f90
```diff
@@ -131,92 +131,63 @@ subroutine sumpos
...
-!    &nerror,nlino,nlinoo,nmz,nthinerr
-!     double precision alf0s1,alf0s2,alf0s3,alf0x2,alf0x3,alf0z2,alf0z3,&
-!    &amp00,bet0s1,bet0s2,bet0s3,bet0x2,bet0x3,bet0z2,bet0z3,chi,coc,   &
-!    &dam1,dchi,ddp1,dp0,dp00,dp10,dpoff,dpsic,dps0,dsign,gam0s1,gam0s2,&
-!    &gam0s3,gam0x1,gam0x2,gam0x3,gam0z1,gam0z2,gam0z3,phag,r0,r0a,rat0,&
-!    &rdev,rmean,rsqsum,rsum,sic,tasia56,tasiar16,tasiar26,tasiar36,    &
-!    &tasiar46,tasiar56,tasiar61,tasiar62,tasiar63,tasiar64,tasiar65,   &
-!    &taus,x11,x13
-
-save
```
kyrsjo: Are you sure that you should remove the `save`?

source/sixve.f90
```diff
@@ -131,92 +131,63 @@ subroutine sumpos
...
-          hv(2,lkk,ia,j)=hv(2,lkk,ia,j-1)*al(1,lkk,ia,ikk)+ hv(4,lkk,ia,j-1)*al(2,lkk,ia,ikk)
-          hv(3,lkk,ia,j)=hv(1,lkk,ia,j-1)*al(3,lkk,ia,ikk)+ hv(3,lkk,ia,j-1)*al(4,lkk,ia,ikk)
-          hv(4,lkk,ia,j)=hv(2,lkk,ia,j-1)*al(3,lkk,ia,ikk)+ hv(4,lkk,ia,j-1)*al(4,lkk,ia,ikk)
-!hr05         hv(5,lkk,ia,j)=hv(5,lkk,ia,j-1)*al(1,lkk,ia,ikk)+ hv(6,   &
-!hr05&lkk,ia,j-1)*al(2,lkk,ia,ikk)+al(5,lkk,ia,ikk)/dpoff
-          hv(5,lkk,ia,j)=(hv(5,lkk,ia,j-1)*al(1,lkk,ia,ikk)+ hv(6,lkk,ia,j-1)*al(2,lkk,ia,ikk))+al(5,lkk,ia,ikk)/dpoff
-!hr05         hv(6,lkk,ia,j)=hv(5,lkk,ia,j-1)*al(3,lkk,ia,ikk)+ hv(6,   &
-!hr05&lkk,ia,j-1)*al(4,lkk,ia,ikk)+al(6,lkk,ia,ikk)/dpoff
-          hv(6,lkk,ia,j)=(hv(5,lkk,ia,j-1)*al(3,lkk,ia,ikk)+ hv(6,lkk,ia,j-1)*al(4,lkk,ia,ikk))+al(6,lkk,ia,ikk)/dpoff
+      if(jm > 1) then
```
kyrsjo: Logic used to be jm != 1. Are you sure it (eg. `mel`) cannot be 0 or negative?

source/sixve.f90
vkbo: Yes.

source/sixve.f90
vkbo: The loop runs from 2 to jm, so even if jm is 0, there is nothing to loop over.

PR #822, opened 2019-04-10
Thick Arrays

vkbo:
There were 30 arrays in `mod_common_main` that were only used in `synuthck`, where they were initialised inside particle loops before every time they were used. That is, at no point were the values accessed elsewhere than within the loop and on the same particle. In other words, they do not need to be arrays.

In fact, this is how it is done in the routine `envars`. In `envarsv` the arrays are allocated locally, but also here the values are never accessed outside of loops, so also in this routine the arrays are redundant.

Each commit in the PR addresses one array, or a set of related arrays. It may be easier to read the diff on a commit by commit basis.


source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
+  integer ih1,ih2,j,kz1,l,l1,l2
 
-  real(kind=fPrec) fokm
+  ! Local version of variables normally found in mod_common_main
+  real(kind=fPrec) dpd(npart),dpsq(npart),dpsv(npart),fokqv(npart),oidpsv(npart),rvv(npart)
+  real(kind=fPrec) aek,afok,as3,as4,as6,co,fi,fok,fok1,g,gl,hc,hi,hi1,hm,hp,hs,rho,rhoc,rhoi,&
+    si,siq,sm1,sm12,sm2,sm23,sm3,wf,wfa,wfhi,fokm
 
-!-----------------------------------------------------------------------
   save
```
kyrsjo: Do we need to `save` this, thus keeping the arrays always-in-memory in case the function gets called again?

Also, isn't `npart` dynamic now? How does that work when you declare a static e.g. `dpd(npart)`?

source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
-    if (kz1.eq.4) goto 60
-    if (kz1.eq.5) goto 40
-    if (kz1.eq.6) goto 60
-    if (kz1.eq.7) goto 100
-    if (kz1.eq.8) goto 100
-    if (kz1.eq.9) goto 140
+    if(abs(el(l)) <= pieni) goto 160
+
+    kz1 = kz(l)+1
+    if(kz1 == 1) goto 20
```
kyrsjo: Select case? Elseif?

source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
+      as3  = (-one*rvv(j))*(((dpsv(j)*rho)/(two*dpsq(j)))*sm23-(rho*dpsq(j))*(one-sm1))
+      as4  = ((-one*rvv(j))*sm23)/c2e3
+      as6  = ((-one*rvv(j))*(el(l)+sm1*sm2))/c4e3
+      as(1,ih1,j,l) = (el(l)*(one-rvv(j))-rvv(j)*((dpsv(j)**2/(four*dpd(j)))*sm12+dpsv(j)*(el(l)-sm2)))*c1e3
+      as(2,ih1,j,l) = (-one*rvv(j))*((dpsv(j)/((two*rho)*dpsq(j)))*sm12-(sm2*dpsq(j))/rho)+fok1*as3
+      as(3,ih1,j,l) = as3
+      as(4,ih1,j,l) = as4+(two*as6)*fok1
+      as(5,ih1,j,l) = (as6*fok1**2-(rvv(j)*sm12)/(c4e3*rho**2))+fok1*as4
+      as(6,ih1,j,l) = as6
 !--VERTIKAL
```
kyrsjo: Degermanizing required :)

source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
+      al(4,1,j,l) = one
+      al(4,2,j,l) = one
+      as(6,1,j,l) = ((-one*rvv(j))*el(l))/c2e3
+      as(6,2,j,l) = as(6,1,j,l)
+      as(1,1,j,l) = (el(l)*(one-rvv(j)))*c1e3
+    end do
     goto 160
 !-----------------------------------------------------------------------
 !  RECTANGULAR MAGNET
 !  HORIZONTAL
```
kyrsjo: Vertical or horizontal? Comments are strange here...

source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
+      al(4,ih2,j,l) = al(1,ih2,j,l)
+      as6 = ((-one*rvv(j))*al(2,ih2,j,l))/c2e3
+      as(4,ih2,j,l) = ((-one*two)*as6)*fok1
+      as(5,ih2,j,l) = as6*fok1**2
+      as(6,ih2,j,l) = as6
+    end do
     goto 160
 !-----------------------------------------------------------------------
 !  SEKTORMAGNET
 !  HORIZONTAL
```
kyrsjo: There are also vertical ones here...

source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
+
+      sm12 = el(l)-al(1,ih1,j,l)*al(2,ih1,j,l)
+      sm23 = al(2,ih1,j,l)*al(3,ih1,j,l)
+      as(1,ih1,j,l) = (el(l)*(one-rvv(j))-rvv(j)*((dpsv(j)**2/(four*dpd(j)))*sm12+dpsv(j)*(el(l)-al(2,ih1,j,l))))*c1e3
+      as(2,ih1,j,l) = (-one*rvv(j))*((dpsv(j)/((two*rho)*dpsq(j)))*sm12-dpd(j)*siq)
+      as(3,ih1,j,l) = (-one*rvv(j))*(((dpsv(j)*rho)/(two*dpsq(j)))*sm23-dpd(j)*rhoc)
+      as(4,ih1,j,l) = ((-one*rvv(j))*sm23)/c2e3
+      as(5,ih1,j,l) = ((-one*rvv(j))*sm12)/(c4e3*rho**2)
+      as(6,ih1,j,l) = ((-one*rvv(j))*(el(l)+al(1,ih1,j,l)*al(2,ih1,j,l)))/c4e3
 !--VERTIKAL
```
kyrsjo: More german...

source/sixve.f90
```diff
@@ -211,397 +211,365 @@ subroutine envarsv(dpsv,oidpsv,rvv,ekv)
...
+      al(1,ih2,j,l) = one
+      al(2,ih2,j,l) = el(l)
+      al(3,ih2,j,l) = zero
+      al(4,ih2,j,l) = one
+      as(6,ih2,j,l) = ((-one*rvv(j))*al(2,ih2,j,l))/c2e3
+    end do
     goto 160
 !-----------------------------------------------------------------------
 !  QUADRUPOLE
 !  FOCUSSING
```
kyrsjo: Splelling

kyrsjo: Also, defocusing is further down in the same block...

source/track_thick.f90
```diff
@@ -1886,317 +1886,324 @@ subroutine synuthck
...
-!Eric
 !-----------------------------------------------------------------------
 !  DRIFTLENGTH
 !-----------------------------------------------------------------------
-    if (kz1.eq.1) then
+    if(kz1 == 1) then
       goto 20
 !-----------------------------------------------------------------------
 !  RECTANGULAR MAGNET
 !  HORIZONTAL
```
kyrsjo: See comments about comments in the previous file. I won't repeat this further...

source/track_thick.f90
```diff
@@ -1886,317 +1886,324 @@ subroutine synuthck
...
+140   do j=1,napx
+        rhoi = ed(l)/dpsq(j)
+        fok  = rhoi*tan_mb((el(l)*rhoi)*half)                  !hr01
+        al(3,1,j,l) = fok
+        al(3,2,j,l) = -fok
+      end do
       goto 160
     else
 !Eric
 ! Is really an error but old code went to 160
```
kyrsjo: If it is an error, then print a warning at least? Or the comment is wrong and kz can be not in the range of the else-ifs...

source/sixve.f90
vkbo: Can't because the goto labels are in use also aside from selecting on kz

source/sixve.f90
vkbo: I de-germanised and norwegianised it. True story.

source/sixve.f90
vkbo: No idea.

source/sixve.f90
vkbo: I assumed the save was needed for the remaining arrays, which is only a few now, so it's not much of an issue. Although I could test if removing save blows up.

source/track_thick.f90
vkbo: Feel free to investigate whether Eric is correct, and add this. The thick tracking code is not well maintained, and not all that well covered in tests. Although most of this routine is covered.

vkbo:
Ok, so, after your comments @kyrsjo , I removed the temp arrays in `envarsv` and loaded the global ones instead. See my last few commits that do them one by one. I also removed the save flag, which is now trivial anyway since there are no arrays left.

The dpd and dpsq arrays are also now the global ones, and should be fine. The fokqv one is computed in the exact same way as in `thick_arrays`, which is `+/-ekv`.

kyrsjo:
I'll heard you like PRs, so I'm making a PR to your PR!

vkbo:
> I'll heard you like PRs, so I'm making a PR to your PR!

Careful, or I'll PR your PR to my PR.

kyrsjo:
Wait, I need to PR my PR so you can PR your PR with my PR!

kyrsjo:
Plz mrg. https://github.com/vkbo/SixTrack/pull/2

PR #823, opened 2019-04-11
Print flag bug

vkbo:
A bit of code was skipped if the PRINT flag was off, causing a segfault due to negative index as it computes `kz(k) = abs(kz(k))` for `kz(k) == 12`.

### DIFF
--- a/source/sixtrack.f90
+++ b/source/sixtrack.f90
@@ -963,7 +963,7 @@ subroutine daten
   if(sixin_ncy2 == 0) il1 = il-1
   do k=1,il1
     if(abs(kz(k)) == 12) then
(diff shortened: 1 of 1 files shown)


source/sixtrack.f90
```diff
@@ -929,6 +929,15 @@ subroutine daten
     end do
   end if
 
+  il1 = il
+  if(sixin_ncy2 == 0) il1 = il-1
```
kyrsjo: Consider adding a comment that this is (as far as I remember) due to skipping the last, automatically inserted, CAV "element".

vkbo: If you want me to add a comment, please provide the comment you want added.

vkbo:
It makes no sense to move the computing of the phase into the middle of the skipped print section. All it does is convert degrees to radians, so I will just de-convert it for the print.

PR #824, opened 2019-04-11
Improve coverage

vkbo:
Two things to improve coverage:
* Removes an unused debugging routine I wrote a while back.
* Adds two tests to make sure the full set of particle state options are tested.


source/sixtrack.f90
```diff
@@ -1106,10 +1113,9 @@ subroutine daten
     write(lout,"(a)") repeat("-",200)
     do j=1,il
       if(kz(j) == 20 .and. parbe(j,17) == 1) then
-        write(lout,"(t10,a16,5x,i4,7x,1pe10.3,2x,1pe10.3,2x,1pe10.3,2x,1pe10.3,2x,1pe10.3,2x,1pe10.3&
```
kyrsjo: Thanks.

source/sixtrack.f90
```diff
@@ -1120,7 +1126,7 @@ subroutine daten
     write(lout,"(a)") str_divLine
     do j=1,il
       if (kz(j) == 20 .and. parbe(j,17) == 0) then
-        write(lout,"(t10,a16,5x,i4,7x,1pe10.3,2x,1pe10.3,2x,1pe10.3,2x,1pe10.3)") &
```
kyrsjo: Ditto.

vkbo:
This PR includes the print flag bug PR #823, as the tests I add here segfaulted without it. So the other one should be merged first. Anyone care to approve it?

PR #825, opened 2019-04-11
Testing that SixTrack fails properly

vkbo:
I was thinking we need to test that SixTrack also exits properly on invalid input.
This PR does not need to be merged yet, it's just a proposed way for checking this.

How it works:
* Adds a special sett of tests starting with `error_` that also is given the label `error`
* These tests call a `errorTests.py` instead of the SixTrack exec.
* The script takes a template fort.3 file and inserts a set of lines from a dictionary of arrays. The dictionary key is the name of the test, and the array has the entry per line for the block.
* A loop generates and executes these in a loop and records prints to a file the file name and True/False for exit == 0
* It also appends the returned `stderr` outputs to a file that can be compared.
* The two resulting files are diffed with their canonicals, and the test pass or fail accordingly.

In is short, fast and easily extendable.


vkbo:
After a discussion with @kyrsjo, we changed the logic a bit.
* There is now a function that does all the checking, and it now writes both a summary file and a tail of the fort.6 at the error point.
* We also diff the standard output files directly, thus not going via the SixTestWrapper at all.

kyrsjo:
Hi, this all looks good to me, however I think e.g. @rdemaria and/or @tpersson should do the review, since I'm a bit close to the development on this one...

vkbo:
Now that errors are printed to stderr, it is easier to collect them into a single file for comparison here. Scripts have been updated accordingly.

Ready to merge, and for more tests to be added.

PR #828, opened 2019-04-12
Cavities changes

vkbo:
After finding the bug with cavities, @kyrsjo and I went through some of the code and wanted to make some changes.

This PR:
* Adds two tests with tracking below transition energy, splitting an RF cavity into two. One using the `kz = -12, phase = 0 degrees` convention, and one using the `kz = +12, phase = 180 degrees` convention. I added one test for thin 6D and one for thick 6D.
* Changes the input parsing to keep the `kz = +/-12` value for cavities, while before the sign was stored in an array of size `nele`, and the `kz(ix)` set to the absolute value. The four places this sign is used now just uses the standard Fortran `sign(1,kz(ix))` call instead.
* All of the initialisation of type 12 elements is now handled by `initialise_elements` instead of being distributed three places.


source/common_modules.f90
```diff
@@ -197,12 +197,16 @@ integer,          save :: curveff    = 0    ! Enable the curvature effect in a c
...
   integer,          save :: idp        = 0    ! Synchrotron motion
-  integer,          save :: ncy        = 0    ! Number of cavity locations
   integer,          save :: ixcav      = 0    ! Stores ix, presumably for cavity
   integer,          save :: icode      = 0
   integer,          save :: idam       = 0
   integer,          save :: its6d      = 0
 
+  ! RF Cavities
+  integer,          save :: icy        = 0    ! Accelerating cavity: Number of "CAV" locations in STRUCT
+  integer,          save :: ncy        = 0    ! Accelerating cavity: Number of cavity locations in SING
```
kyrsjo: Cavity location in SING? Should this be in STRUCT?

source/sixda.f90
```diff
@@ -788,7 +788,7 @@ subroutine runda
           write(7,*) '5'
```
kyrsjo: `Write(7,*)` ? Not using the allocation, using format * ?

source/sixtrack.f90
```diff
@@ -1680,473 +1671,375 @@ subroutine wzsub(x,y,u,v)
...
 end subroutine wzsub
 
-subroutine initialize_element(ix,lfirst)
-!-----------------------------------------------------------------------
-!     K.Sjobak & A.Santamaria, BE-ABP/HSS
-!     last modified: 23-12-2016
-!     Initialize a lattice element with index elIdx,
-!     such as done when reading fort.2 (GEOM) and in DYNK.
+! ================================================================================================ !
+! K. Sjobak, A. Santamaria, BE-ABP-HSS
```
kyrsjo: Add your name

source/sixtrack.f90
```diff
@@ -1680,473 +1671,375 @@ subroutine wzsub(x,y,u,v)
...
-                  beamoff(5,imbb(i))=track6d(4,1)*c1e3
-                  beamoff(6,imbb(i))=track6d(6,1)
-                  napx=napx0
+  ! Cavities (ktrack = 2 for thin)
+  elseif(abs(kz(ix)) == 12) then
+    dynk_elemData(ix,3) = el(ix)
+    phasc(ix) = el(ix)*rad
+    el(ix) = zero
+    if(lfirst) then
+      if(abs(ed(ix)) > pieni .and. abs(ek(ix)) > pieni) then
```
kyrsjo: Hmm. This logic would interfer with DYNK of RF cavities, but I'm not entirely sure how to solve it.

Basically how it is now, I think it will disable the cavity completely (switching to 4D tracking if all cavities are like this) if neither the voltage or frequency is set already here just after parsing the geometry; which is *probably* OK, since if the frequency is unset, then the cavity is really not properly initialized... This should at least be detected in dynk_pretrack or sanitycheck tough, before trying to use the RFCAV.

I believe there is a "if voltage is zero AND not used by DYNK then ktrack=31" at some later point which would make it the main disabling?

There is some history to these kind of fixes; there may be some useful notes in this PR, which fixed a bunch of bugs in the RF cavity parsing: https://github.com/SixTrack/SixTrack/commit/334d0a3e4ae15c9f0403e12e425d1aab47b42bcf

As a side, I wonder if we really need the `kp` array at all? At least for RF cavities... AFAIK they are (where?) also used by aperture.

source/common_modules.f90
vkbo: The value is incremented in SING parsing.

source/sixda.f90
vkbo: Got nothing to do with this PR.

source/sixtrack.f90
vkbo: Unless I missed something, this is exactly the same logic as is currently in SixTrack.

source/common_modules.f90
vkbo: `ncy` is `icy*mper`. Updated comment.

Issue #826, opened 2019-04-12
Segfault on missing elens block when elens in lattice


vkbo:
If there are elements of type 29 in the single element list (at least with thick 6d), but no elens block in fort.3, SixTrack segfaults (with debug build). This should probably be caught before tracking.

Error: Fortran runtime error: Index '0' of dimension 1 of array 'elens_offset_x' below lower bound of 1

Can be reproduced by commenting out the elens block in test elensidealthck6d.

amereghe:
thanks for signalling it.
yes there is no explicit check like this, as for the chebyshev maps - I am wondering what about other element types...

vkbo:
There's a similar bug between SCATTER and DYNK as well. I'll get to it soon.
I think as a principle, SixTrack shouldn't segfault on erroneous user input, but it isn't always easy to catch these.

amereghe:
agreed - I simply think that such a check does not exist at all.
I guess that the best thing to do is to have a subroutine called after daten (or anyway once input parsing is done) that performs the check.
I can draft it in the context of the issue on elens

vkbo:
Agree. A few modules have `_postInput` routines that are either called when the block is closed, or after `ENDE`. I think that is a relatively clean approach.

amereghe:
that's what I thought first - but I think it is better to have the search in a place only, with all concerned elements

vkbo:
Hmm, I'm not sure I agree here. I think it's cleaner if each block solves its own dependencies within the corresponding module. We've been trying to keep block-specific code in their own modules, and not spread the code around. Since the dependency here is a bit more detailed than just the existence of other blocks, I think this is a better approach.

@kyrsjo ?

amereghe:
I think that indeed what you want to spot is something missing in your blocks - so I think we should have an explicit loop on all SINGLE ELEMENTS, calling check functions specific to the element type.

It is fine with me to go with the `_postInput` subroutines - it is just less evident to new comers that those also check missing info from the user...

vkbo:
Ah, I see. You want a loop that calls specific checking routines in feature modules that check dependencies on an element by element basis. That is very specific to this issue though, and in case of the DYNK-SCATTER case, becomes very convoluted.

amereghe:
ok - for elenses and chebyshev, the check is already there, but in the `_parseInputDone` subroutines. I will move it to the `_postInput` subroutines

vkbo:
Fixed in #833

PR #833, opened 2019-04-15
moving check of lattice against un-declared chebyshev/electron lenses flagged in fort.2 but not declared inf fort.3

amereghe:
This PR fixes issue https://github.com/SixTrack/SixTrack/issues/826

Electron lenses / chebyshev lenses flagged in `fort.2` but not declared as such in `fort.3` were caught only in case the respective blocks were active, since the check was in `_parseInputDone` instead of `_postInput`. Hence, in case the blocks were not there, the flagged elements caused sixtrack to segfault.

At the same time:
   * added a check that the number of chebyshev lenses / electron lenses found in the SINGLE ELEMENT array matches that found in the `fort.3`.


amereghe:
not for merging yet!

amereghe:
ok ready to merge

PR #839, opened 2019-04-17
foxifying e-lenses

amereghe:
This PR aims at foxifying e-lenses. The most difficult part was to get into the logics of fox and to foxify elenses with an electron radial profile from measurements - there, I moved from a linear interpolation to a polinomial interpolation around the point of interest of order that can be chosen by the user.

I took inspiration for the foxification of elenses from `wireda`, hence with a subroutine defined in the elens module called in `six_fox.f90`. A variable states if the elens should be considered or not in fox, defaulting to `.true.`; anyway, the user is free to explicitly tell SixTrack if to consider or not each elens for FOX in the `fort.3` file. Anyway, the elens kick is taken into account in the FOX calculation only if the closed orbit goes through the electron distribution, i.e. the hollow part does not contribute to FOX.

The polinomial interpolation is essentially general, implemented in the `utils` module, and it can be used anywhere needed. I have added also an integration subroutine, trying to be more consistent with the polinomial interpolation when computing the integral of the interpolated function. Polinomial interpolation of radial elens profiles is presently set to linear, but it can be changed in the future; the user can change this setting via the `fort.3`.

Major changes involve:
   * `elens` module;
   * `utils` module;
   * addition of a couple of tests about the FOX part:
      1. a test verifying that the kicks considered by the FOX part are consistent with the kicks used during regular tracking;
      1. a test verifying that for a uniform, non hollow lens, the tune shift due to the elens is similar to that of a quadrupole, with the same sign on both planes.

Other minor changes:
   * introduced `elens_kz`, storing the type in `fort.2` (i.e. `29`), and `elens_ktrack` (i.e. `63`) and used throughout the code - hence few modules have been affected (eg `dynk`);
   * kick from elenses moved from the include file to a subroutine coded in the `elens` module;
   * added a constant in `numerical_constants` (needed for polinomial interpolation);
   * elens tests updated;
   * increased acceptable repo size to 660k kB;


source/utils.f90
```diff
@@ -70,4 +72,346 @@ real(kind=fPrec) function lininterp(x,xvals,yvals,datalen)
...
+
+    if(datalen <= 0) then
+      checkArray=.false.
+      if (llprint) write(lout,"(a)") "UTILS> ERROR checkArray: datalen was 0!"
+    end if
+    do ii=1, datalen-1
+      if(xvals(ii) >= xvals(ii+1)) then
+        checkArray=.false.
+        if (llprint) write(lout,"(a)") "UTILS> ERROR checkArray: xvals should be in increasing order"
+      end if
```
vkbo: These should now be printed to `lerr`, not `lout`. Also, do these errors cause an exit somewhere based on the checkArray flag?

source/utils.f90
```diff
@@ -70,4 +72,346 @@ real(kind=fPrec) function lininterp(x,xvals,yvals,datalen)
...
+    do mm=1,nn-1      ! For each column of the tableau,
+      do ii=1,nn-mm   ! we loop over the current c’s and d’s and update them.
+        ho=xa(ii)-x
+        hp=xa(ii+mm)-x
+        ww=Cs(ii+1)-Ds(ii)
+        den=ho-hp
+        if (den.eq.zero) then
+          ! This error can occur only if two input xa’s are (to within roundoff) identical.
+          write(lout,"(A)") "UTILS> ERROR nevilleMethod den.eq.zero"
+          call prror
```
vkbo: Another `lout` that should be `lerr`.

source/utils.f90
```diff
@@ -70,4 +72,346 @@ real(kind=fPrec) function lininterp(x,xvals,yvals,datalen)
...
+       case(1)
+          tmpInt=fmax-fmin
+       case(2)
+          tmpInt=two*(pi*(fmax-fmin))
+       case(3)
+          tmpInt=four*(pi*(fmax-fmin))
+       case default
+          write(lout,"(a,i0)") "UTILS> ERROR polintegrate: order not recognised:",order
+          write(lout,"(a)")    "UTILS>       recognised values: 1D (1), 2D (2), 3D (3)"
+          call prror
```
vkbo: And two more lines of `lerr`.

source/elens.f90
```diff
@@ -660,6 +679,226 @@ subroutine normaliseRadialProfile(ifile)
...
+!FOX       LINLIN=LINLIN+(TMPTMP*TMPCOF)/RKK ;
+!FOX       TMPTMP=TMPTMP*RR ;
+        end do
+!FOX    FRR=(LINLIN-ELEFR1)/ELENOR ;
+        call dealloc(cof,'cof')
+
+      case default
+        write(lout,"(a,i0,a)") "ELENS> ERROR in elens_kick_fox: elens_type=",elens_type(ielens(ix))," not recognized. "
+        write(lout,"(a)")      "ELENS>       Possible values for type are: 1, 2 and 3"
+        call prror
```
vkbo: These should be `lerr`, not `lout`.

amereghe:
Hi Veronica,
thanks for the feedback. I have also this line in the CR part of elenses:
```
write(lout,"(a,i0)") "READERR in elens_crcheck; fileUnit = ",fileUnit
write(93,  "(a,i0)") "READERR in elens_crcheck; fileUnit = ",fileUnit
```
shall I replace `lout` by `lerr`?
Thanks,

source/utils.f90
amereghe: done.
For the moment, `checkArray` is not used, but I will plug it in into the main code soon

vkbo:
Yes. The error messages from CR should also go to `lerr`, and start with `SIXTRACR> ERROR`. This is not done consistently throughout SixTrack, but it's good to clean up as we go.

amereghe:
ok the PR is still not complete but most of the code should be there.
The FOX part of the elens now should be more consistent, following the considerations in [2019-11-04_slide4RDeMaria.pdf](https://github.com/SixTrack/SixTrack/files/3813392/2019-11-04_slide4RDeMaria.pdf).

I am stuck on the following aspects:
* how to properly set the first derivative of the radial kick to `0.0` or to a specific value? I guess I should use a line like `call dapek(FRR,hh,<myVal>)`, but how should `hh` look like? `hh(:)=1`?
* I have debug messages showing me that the constant part of the kick is correctly computed by the FOX part - I have a test, not committed for the time being not to use too much repo sapce. How to check that the elens kick computed in the context of FOX has an effect on the CO search? Typically, the kick is in the order of few urad. The optics computed by `linopt` does not reflect the FOX search. On the other hand, the FOX iteration ends up with the same 6D closed orbit no matter if elenses are considered or not - which rings me a bell, but I don't understand where/what to search, since `YY` is modified by the FOX part of elenses...

@rdemaria @tpersson can you have a quick look at the code and give me some feedback?
Thanks a lot in advance,

amereghe:
ok, issue about elens kick not taken into account in FOX solved - cfr. [fix](https://github.com/SixTrack/SixTrack/pull/839/commits/0487dc81e2d95dbb2e6128589bf47c82b0f9b513).

The point on the derivative is still there...

amereghe:
ok I think I am done with the FOX part.
@tpersson @rdemaria could you please review the code?
If it is correct, then I will update the tests once forever and proceed with the other PRs.
Thanks!
Cheers,
A.

amereghe:
For the record - without the check that `RRA` is `zero`, you get:
```
                                      
ELENS> ELENS_KICK_FOX computing at XA= 0.0000000000000000E+00 - YA= 0.0000000000000000E+00
DAFUN> ERROR SQRT does not exist for vector 474 const term =  0.00000E+00
DABNEW> ERROR DAFUN 

    +++++++++++++++++++++++++++++
    +      ERROR DETECTED!      +
    + RUN TERMINATED ABNORMALLY +
    +++++++++++++++++++++++++++++
```

amereghe:
ok issue solved skipping the `SQRT` and setting `RR` directly to `ZERO` via FOX

amereghe:
Ready for merge! @rdemaria @tpersson

amereghe:
@rdemaria @tpersson could you please review this PR? 
It is there since a couple of weeks ago, and it should be merged in before the other one on e-lenses...
Thanks!
Cheers,

rdemaria:
I am off this week, I can approve, but I have not looked in detail

amereghe:
Hi @rdemaria ,

ok. I think that the implementation is sound, as there is a specific test that verifies that for a (full) electron lens I get a detuning similar to that of a quadrupole of the same gradient, with the same sign on both Qx and Qy (contrary to the case of the quad).

@tpersson @rdemaria , shall I wait for a detailed review (it would be good, though maybe not immediate) or shall I merge straight away?

amereghe:
Hi @rdemaria ,
please find attached a new [.pdf](https://github.com/SixTrack/SixTrack/files/4224785/2020-02-19_slides4RDeMaria.pdf) for discussion, with the correct math (the previous .pdf had some errors, sorry about it), and trying to be as short as possible. You will find a slide with the case of the regular e-lens, followed by one for the case of full e-lens and (radially) thin e-lens.

amereghe:
As agreed with @rdemaria , at r=R1 and r=R2 kick and derivatives of elens will be dictated by the corona part.
For ease of life, we merge this PR (and the following one), and the final FOX implementation will arrive soon with a clean, new PR.

PR #840, opened 2019-04-17
Hexadecimal version of pi and pi-related constants

vkbo:
Just another small change on my todo list.

I've defined all the constants that are pi-related (and two more) in terms of a correctly truncated binary representation. They are all based on the gfortran casting of the quad precision, and tuned to produce the exact same binary to the size of the mantissa for double and single precision. 

**Changes**
* The double precision one was corrected in the last bit for the `eulergamma` variable (which is the same as stated on wikipedia, but the variable isn't used anywhere in SixTrack currently anyway). 
* All of the single precision variables were corrected on the last bit. That is, setting the values with the full list of decimals tended to round the last bit up for single precision, while the corresponding double precision bit was lower.

**Testing**
* Debian: Passes all fast tests with gfortran, ifort and nagfor.
* Windows10: Passes all fast tests with gfortran for both 64 and 32 bit.

The code used to generate the hex values is this one:
```
program main

  use, intrinsic :: iso_fortran_env

  implicit none

  real(kind=real32),  parameter :: pi32   = 3.1415925_real32
  real(kind=real64),  parameter :: pi64   = 3.141592653589793238462643383279502884197169399375105820974_real64
  real(kind=real128), parameter :: pi128  = 3.141592653589793238462643383279502884197169399375105820974_real128

  real(kind=real32),  parameter :: pi32x  = transfer(z'40490fda',1.0_real32)
  real(kind=real64),  parameter :: pi64x  = transfer(z'400921fb54442d18',1.0_real64)
  real(kind=real128), parameter :: pi128x = transfer(z'4000921fb54442d18469898cc51701b8',1.0_real128)

  write(*,"(a,f10.8,28x,z8.8,33x,b32.32)")    "REAL32  : ",pi32, pi32, pi32
  write(*,"(a,f19.17,19x,z16.16,22x,b64.64)") "REAL64  : ",pi64, pi64, pi64
  write(*,"(a,f36.34,2x,z32.32,2x,b128.128)") "REAL128 : ",pi128,pi128,pi128

  write(*,"(a,f10.8,28x,z8.8,33x,b32.32)")    "REAL32  : ",pi32x, pi32x, pi32x
  write(*,"(a,f19.17,19x,z16.16,22x,b64.64)") "REAL64  : ",pi64x, pi64x, pi64x
  write(*,"(a,f36.34,2x,z32.32,2x,b128.128)") "REAL128 : ",pi128x,pi128x,pi128x

end program main
```

For pi, it outputs this:
```
REAL32  : 3.14159250                            40490FDA                                 01000000010010010000111111011010
REAL64  : 3.14159265358979312                   400921FB54442D18                      0100000000001001001000011111101101010100010001000010110100011000
REAL128 : 3.1415926535897932384626433832795028  4000921FB54442D18469898CC51701B8  01000000000000001001001000011111101101010100010001000010110100011000010001101001100010011000110011000101000101110000000110111000
REAL32  : 3.14159250                            40490FDA                                 01000000010010010000111111011010
REAL64  : 3.14159265358979312                   400921FB54442D18                      0100000000001001001000011111101101010100010001000010110100011000
REAL128 : 3.1415926535897932384626433832795028  4000921FB54442D18469898CC51701B8  01000000000000001001001000011111101101010100010001000010110100011000010001101001100010011000110011000101000101110000000110111000
```

### DIFF
--- a/source/constants.f90
+++ b/source/constants.f90
@@ -59,10 +59,9 @@ module numerical_constants
 
 ! real(kind=fPrec), parameter :: pi      = 3.141592653589793238462643383279502884197169399375105820974_fPrec
 ! real(kind=fPrec), parameter :: inv_ln2 = 1.442695040888963407359924681001892137426645954152985934135_fPrec
-! real(kind=fPrec), parameter :: pi2     = 0.5_fPrec*pi
-! real(kind=fPrec), parameter :: twopi   = 2.0_fPrec*pi
-! real(kind=fPrec), parameter :: pisqrt  = sqrt(pi)
-! real(kind=fPrec), parameter :: rad     = pi/180.0_fPrec
+
+  ! We set the parameters of irrational numbers explicitly to their nearest binary representation
+  ! without depending on rounding from decimal representation. These use decimal round near.
 
 #ifdef SINGLE_MATH
   real(kind=fPrec), parameter :: pi      = transfer(z'40490fdb',1.0_fPrec)
@@ -71,6 +70,8 @@ module numerical_constants
   real(kind=fPrec), parameter :: pisqrt  = transfer(z'3fe2dfc5',1.0_fPrec) ! sqrt(pi)
   real(kind=fPrec), parameter :: inv_ln2 = transfer(z'3fb8aa3b',1.0_fPrec) ! 1/log(2)
   real(kind=fPrec), parameter :: rad     = transfer(z'3c8efa35',1.0_fPrec) ! pi/180.0_fPrec
+! real(kind=fPrec), parameter :: pieni   = transfer(z'00800000',1.0_fPrec) ! Smallest normal number
+! real(kind=fPrec), parameter :: pieni   = transfer(z'006CE3EE',1.0_fPrec) ! 1e-38 is subnormal
 #endif
 #ifdef DOUBLE_MATH
   real(kind=fPrec), parameter :: pi      = transfer(z'400921fb54442d18',1.0_fPrec)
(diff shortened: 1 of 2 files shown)


vkbo:
Also passes on Windows10, both 64bit and 32bit.

kyrsjo:
Cool! Also tagging @ericmcintosh , I assume he's interested :)

A quick comment: It would be great to document this somewhere properly - something like a folder inside the doc folder with (1) the example code and (2) a MarkDown document describing the hows and whys (e.g. why truncate ...110101 to ...11010 and not round to ...11011, as seen for PI in single precision) etc., so that things get done consistently throughout SixTrack.

vkbo:
Changed the single precision pi to round up on last bit instead. The discussion I read suggested truncating the binary, while wikipedia and other sources uses roun nearest not round zero since the next bits are `101`.

ericmcintosh:
Sorry I am not up to speed on this. PI and Wijipedia......

I seem to remember that Frank? calculated Pi ot Pi/2 and then used that value 
for other related constants.
Frank will tell us.

I can promise to provide a doc section on coding comnventions in particlualr
with respect to reproducibility.

Eric.

vkbo:
Hi Eric,

The point was just to explicitly set pi, and other related constants, directly to a binary value so that conversion from decimal numbers isn't necessary. Especially since this is known to not be entirely reproducible.

Wikipedia is only relevant because the articles about single and double precision uses pi as an example, so I cross-checked my binary values with theirs.

vkbo:
The code to set this for double precision is:
```
  real(kind=fPrec), parameter :: pi      = transfer(z'400921fb54442d18',1.0_fPrec)
  real(kind=fPrec), parameter :: pi2     = transfer(z'3ff921fb54442d18',1.0_fPrec)
  real(kind=fPrec), parameter :: twopi   = transfer(z'401921fb54442d18',1.0_fPrec)
  real(kind=fPrec), parameter :: pisqrt  = transfer(z'3ffc5bf891b4ef6a',1.0_fPrec)
  real(kind=fPrec), parameter :: inv_ln2 = transfer(z'3ff71547652b82fe',1.0_fPrec)
  real(kind=fPrec), parameter :: rad     = transfer(z'3f91df46a2529d39',1.0_fPrec)
```

ericmcintosh:
Many thanks Veronica. Very good.  Eric.
(Not sure hex constants were accepted everywhere a long time ago! :-)

ericmcintosh:
Hello again Veronica. Very good. We are compiling with round nearest
so I am pretty sure you are correct in rounding Pi in that way.
(Of course if we were compiling with other round options, it
should maybe be different.)  I am not sure of anything these days!
Old age. Thanks again.  Eric.

kyrsjo:
Probably a good idea to take 20 mins to do the doc thing before merging?

vkbo:
I added a few comments in the code. The markdown documentation can be done later.
The small fortran program to compute the binary and hex values is available here:
https://gist.github.com/vkbo/0e5d0c4b7b1ecb533a0486c79a30f741

I also did some digging into the `pieni` variable. Since it is set to `1e-38`, I assume it has something to do with the smallest normal number representable in single precision, however, `1e-38` is actually a subnormal number in 32 bit. The smallest is about `1.175e-38`. Changing it to this, and setting the double and quad to the same binary value, seems to alter the value of two numbers in `fort.10` related to phase advance. I left the hex code commented out in the source, and did not change the definition if `pieni`.

PR #844, opened 2019-04-23
AVX512

vkbo:
Played a bit around with the AVX512 flags in CMake.

No matter what I do, I can't get a GNU exec to produce the same binary results, so most tests fail. For Intel it's not an issue. Although, I couldn't get the SixTestWrapper to build with icpc on my system.

In any case, this PR sets the generic separates `AVX512` from `AVX512_SKYLAKE` for Intel compilers, and for GNU it sets the flags that my `lscpu` prints for Skylake as a starting point.


CMakeLists.txt
```diff
@@ -225,10 +226,10 @@ if(AVX2)
   set(SIXTRACK_FEAT_CR "${SIXTRACK_FEAT_CR}_avx2")
   set(SIXTRACK_FEAT_DA "${SIXTRACK_FEAT_DA}_avx2")
 endif(AVX2)
-if(AVX-512)
+if(AVX512 OR AVX512_SKYLAKE)
```
kyrsjo: Are these flags compatible? Is it OK to activate both?

vkbo: They are just meta flags, so SKYLAKE includes the other one. I can double-check both compilers, but in principle all you get is duplicate flags in the list.

vkbo: gfortran has no complaint, but ifort will override the skylake one with the common one with a warning. I'll add checks to make sure the input makes sense.

PR #849, opened 2019-04-25
Foxifying chebyshev polynomials maps

amereghe:
This PR aims at foxifying chebyshev lenses. I took inspiration for the FOXification from `wireda` and the FOXification of elenses (PRs https://github.com/SixTrack/SixTrack/pull/839 and https://github.com/SixTrack/SixTrack/pull/1056), hence with a subroutine defined in the `cheby` module called in `six_fox.f90`. A variable states if the chebyshev lens should be considered or not in FOX, defaulting to `.true.` (hence Chebyshev lenses are considered in FOX calculations). This setting can be modified by the user. User and physics manual updated.

Other changes:
   * chebyshev maps are actually used on a squared domain;
   * the check against R1 and R2 takes into account rounding issues;
   * max order of cheby polynomials is a parameter in the module header;
   * derivatives of Chebyshev polynomials, used to compute the kick, no longer go through a dangerous division by `1-u^2`;
   * update chebyshev tests:
      * let R1 be 0;
      * calculations checked;
      * add tests, checking:
         * kick by fox routine;
         * tune shift by a Chebyshev map implementing a simple quad;
   * tabular method for inputting/outputting coefficients;
   * minor house-keeping:
      * module fully private, apart from components explicitly labelled as public;
      * module-wide imported modules shrunk down to only `floatPrecision` and `numerical_constants`;
      * improving CPU time when checking if particle/point falls inside domain of chebyshev map;
      * introduced the `cheby_resetI` subroutine;
      * updating some comments.


amereghe:
@rdemaria , @tpersson , have you had time to have a look at the PR?

rdemaria:
I did not see problems, good to go for me.

amereghe:
Hi @rdemaria ,
many thanks. Just let me update the branch against master, and then we can merge.
@tpersson ?

amereghe:
@JamesMolson can I ask you to check the PR numbers of your recent entries in the changelog? I approved and merged the PR forgetting that we report there also the PRs

JamesMolson:
sure

JamesMolson:
```
* Do not perform the pair mapping when geant4 collimation is enabled. #1059
* Enable single sided collimators with geant4 collimation. #1060
* Fix building with geant4 collimation with geant4 releases >= 10.06. #1060
* Fix a mass miss-match with geant4 when entering non-ground state ions into geant4. #1062
* Fix a crash with miss-matched format strings when writing the aperture losses file with geant4 enabled (and not FLUKA). #1062
```

```
* Collimator material names are now case insensitive in geant4. #1062
* In the HION block the PDG ID can now be set as the 5th value. #1062
```

```
* Update FLUKAIO reference. #1057
* Add Si as a possible collimator material for G4. #1059
* Add particle ID and parent ID tracking with geant4. #1062
* Allow setting particle statistical weights in geant4 #1062
* Always enable the EMD physics process in geant4. #1062
* Use global id/parent/weight variables in the FLUKA coupling. #1062
* Start to enable the ability to use collimation with thick lens lattices. #1062
```

PR #851, opened 2019-04-26
Error Test Wrapper Fix

vkbo:
But of course all compilers print a different message when calling 'stop 1'

This PR does not touch the SixTrack source code.


vkbo:
I put a comment in the file explaining why it's commented out. Fixing that one requires modification of the source code, and that can wait.

PR #853, opened 2019-04-26
Library, BOINC and CR fixes/updates

vkbo:
A few things triggered by the latest BOINC release build and testing.

* The building of libArchive linked to the system libz, and our libz header. SixTrack linked to our libz for both. On Windows: BOOM!
* When building with BOINC + API, no checkpoint files are written when the test suite runs since SixTrack asks BOINC for permission to write them, which is doesn't get. That fails all the tests that requires at least one checkpoint/restart to pass. This PR changes the logic slightly so that it always checkpoints on turn 1 and ignores the BOINC API, but obeys it on all other turns.
* Also cleaned up the output to fort.93 from writebin as it was a bit too spammy. It now writes 2 lines instead of 5, but with the same info.
* Also copied the latest release note back into the changelog as I forgot to mention one PR.


vkbo:
Made another couple of changes after talking to @kyrsjo. This is now ready.

I will make a new PR on C/R addressing a few other issues. Now that there is a lot less C/R code (we removed a lot of particle arrays lately), we can simplify the module a bit.

PR #854, opened 2019-04-30
Checkpoint/Restart Rewrite, Bugfix, etc

vkbo:
This PR is a major cleanup of the checkpoint restart module.
The commit history also includes the commits from PR #853, so no point reviewing it until the other one is merged.

Main points in this PR:
* Rename fort.91, 92, 93, 95 and 96 to some more descriptive file names.
* Reduce all the duplicated code for fort.95/96 to a do loop.
* Reduces the output to fort.93, and also uses the fort.3 `DEBUG` flag to determine how much to output.
* The fort.93 output now has tags that indicate where in the C/R code stuff is being logged, rather than having everything tagged with `SIXTRACR>`.
* Rewritten the copying of fort.92 to fort.6 code using existing features in Fortran rather than unnecessary code.
* Ensuring consistency with energy/momentum arrays between C/R and mod_particles.
* Removes the max CR value in the `TRAC` block. Will re-implement the continuation logic at a later point, It is a nice feature to have, but it can be done in a simpler way.
* When I checked the C/R code in `HIONS`, I also cleaned up the module header as it was leaking variables all over the place. Especially in `DYNK`.


source/track_thick.f90
```diff
@@ -555,25 +555,16 @@ subroutine thck4d(nthinerr)
...
-    end do
-  end if
-  do 490 n=numlcr,nnuml
+  nnuml  = numl
+  nfirst = cr_numl
 #else
-  do 490 n=1,numl
+  nfirst = 1
 #endif
+  do 490 n=nfirst,numl
```
kyrsjo: FYI the goto-labled do-loops may be replaced by construct names, e.g.
```turnloop: do n=nfirst,numl
    <stuff>
end do turnloop```

Comment (vkbo) [reply source/track_thick.f90]: I know. But I'm not going to fix the tracking code now.

File: source/checkpoint_restart.f90 [hunk 2]
```diff
@@ -165,728 +248,349 @@ subroutine cr_killSwitch(iTurn)
...
-    write(93,"(a)") "SIXTRACR> CRCHECK crbinrecs:"
-    do j=1,(crnapxo+1)/2
-      write(93,"(2(a,i0))") "SIXTRACR> ",j,": ",crbinrecs(j)
-    end do
-#endif
-    flush(93)
+    write(crlog,"(a)") "CR_CHECK> File "//cr_pntFile(nPoint)//" successfully read"
+    flush(crlog)
+    cr_pntRead(nPoint) = .true.
+    goto 100
```
kyrsjo: Can't this be replaced with a local flag which is flipped when checkpointing has sucessfully finished (followed by an `exit` to break the loop), and that flag checked and if it has not been sucessfully finished, print the error message below the loop and perror?

source/checkpoint_restart.f90
```diff
@@ -165,728 +248,349 @@ subroutine cr_killSwitch(iTurn)
...
-    ! This is not a FLUSH!
-    rewind(lout)
-    endfile(lout,iostat=ierro)
-    close(lout)
-    lout=6
-  endif
+  ! Set up flag for tracking routines to call CRSTART
+  cr_restart = .true.
+  write(lout,"(a)") "SIXTRACR> "//repeat("=",80)
+  write(lout,"(a)") "SIXTRACR>  Restarted"
```
kyrsjo: I would not here say "restarted", so far it's only "armed" the restart.

vkbo: Yeah, this was already copied to the end of crstart, and shouldn't even be here. Will delete.

source/checkpoint_restart.f90
vkbo: Since there was already a flag for this purpose for the file exist logic, I reused it for file check pass here.

PR #862, opened 2019-05-03
Make the error tests work when the supporting library is symlinked in

kyrsjo:
Running the error tests fails when compiling SixTrack as `./cmake_six BUILD_TESTING", since when running a symlinked python script it looks for modules in the folder the script is symlinked from, not in the working directory.


vkbo:
Interesting, because this works perfectly fine on my computer.

PR #866, opened 2019-05-05
Even More Bugfixes for 5.2.6

vkbo:
This PR addresses three issues spotted in nightly builds (two of them caught now that the fringe field test actually runs):

* Floating point exception in fringe field module where `abs(log10(x))` would error when `x` was negative. Changed to `abs(log10(abs(x)))`. This check is actually there to prevent floating point exceptions, so this is not without irony. :)
* The indexing of single elements for fringe fields did not check for negative indices, i.e. block elements. Caught by debug build bounds check.
* Although this has been working for a while, having the error code as an optional argument in the subroutine `prror` actually requires an explicit interface as the routine is not in a module. The compilers have been able to handle it anyway, put `ifort` face-planted into this during yesterday's nightly builds. Since all calls still using an error value used `-1` anyway, I just did a global replace of `call prror(-1)` with `call prror` and removed the optional argument.

In addition, I added backtrace and bounds check flags to the ifort debug build type. Nagfor also complained about unused stuff in the recently changed modules, which I cleaned up too. There is still a lot of that left in the older parts of the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -638,7 +638,7 @@ elseif(Fortran_COMPILER_NAME MATCHES "ifort.*")
 
   set(IFORT ON)
   set(CMAKE_Fortran_FLAGS_RELEASE "-fp-model source -fp-model strict -fp-model no-except -O3 -no-fma -std15 -assume realloc-lhs")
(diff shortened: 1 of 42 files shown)


vkbo:
Oh, and yes, saved again by Steve Lionel's answers on the Intel forum ... :)

PR #867, opened 2019-05-06
CRKILLSWITCH on elens, cheby

vkbo:
Added the CRKILLSWITCH to a bunch of elens and cheby tests


test/CMakeLists.txt
```diff
@@ -327,6 +327,8 @@ if(CR)
     dynk_operators
     dynk_polynomial
     dynk_random
+    elensidealthck4d
+    elensidealthck6d
```
amereghe: rather than the thick tests, I would have added the ion one, but is fine also like this.

vkbo: I added those because I've seen them fail on nightly builds, that's all.

vkbo:
> very nice choice of killing turns :P

Required the least amount of thinking :)

PR #870, opened 2019-05-07
More thickarray cleanup

vkbo:
Spotted more arrays that do not need to be arrays any more.

* For `ekv(j,l)` all `j` were set to `ek(l)` where `j` is particle id and `l` is single element id. `ekv(j,l)` replaced by `ek(l)`.
* `fokqv(j)` only contained `ekv(j,l)` times a sign that depended on horisontal or vertical magnet. Since `ekv` was repalced with `ek`, there is no longer a particle ID dependency, so it's a single variable with a sign per element.
* There is no need to checkpoint the arrays `dpd`, `dpsq`, `as`, and `al`. These are computed by the subroutine `synuthck`, and can just be recomputed from energy/momentum arrays after restart.


rdemaria:
Where do I see that ekv(j,l) is not particle dependent?

vkbo:
The only place it's ever set: in main_cr.

PR #872, opened 2019-05-08
Simu Block, Step One

vkbo:
This PR adds a SIMU block that replaces the current TRAC block.
The TRAC block is of course still supported, but both cannot be used simultaneously.

We now have a third (8 of 25) variables in the TRAC block disabled, but mostly still require dummy variables. It is getting a bit tedious to use it, so the keyword/value format of the SIMU block is a bit more convenient.

I've updated the manual as well.


rdemaria:
This mixes a bit the simulation and distribution described in the [wiki](https://github.com/SixTrack/SixTrack/wiki/new-DIST-and-SIMUL-block).  Would it be possible to design the user interface to match the one in wiki?

vkbo:
What part does it mix?

rdemaria:
The AMPLITUDE keyword belongs to the distribution bloc

vkbo:
Sure, the feature will be re-implemented in DIST, but we still need to support the current generator in maincr until then.

vkbo:
I can also just remove the AMPLITUDE keyword entirely from the SIMU block, which means the entire first line of the INIT block don't need to be supported either.

If people need to use the amplitude feature, they can still use the TRAC block.

rdemaria:
I would review REF_MASS, A, Z and Q.

For tracking mass we need a "reference mass" and a "reference charge" to compute `mtc` and related quantities. For scattering codes, as far as I know, we also need (A,Z,Q) or a PDG number or a fluka number. @james, @Alessio do you confirm?

amereghe:
yes, where A is the actual mass [GeV/c2].
...still have to go through the PR, though...

rdemaria:
Assuming that internally we have (following James #713 )
nucm0 nucm(j)
qq0  nqq(j)
aa0  naa(j)
zz0  nzz(j)
pdgid0 pdgid(j) 

I propose the interface to have:
PROTON_MASS=...
ELECTRON_MASS=...
(other constants)

REF_MASS=nucm0 (default PROTON_MASS)
REF_CHARGE=qq0 (default 1)
REF_A=aa0 (default 0)
REF_Z=zz0 (default 1)
REF_PDGID=pdgid0 (default 0)

where REF_PDGID is incompatible with any of charge, a,z and not implemented at the moment.

In this PR we also need to replace in the code nzz and zz0 with nqq and qq0 (following again #713) in the tracking part to implement the basic functionality.

amereghe:
agreed - if we default to protons, then:
```
REF_A=aa0 (default 1)
REF_PDGID=pdgid0 (default 2212)
```

rdemaria:
Ok for me

vkbo:
The default has to be proton for compatibility, so a Z Q = 1 1 1.
This is currently the case. I added all the needed arrays for this, and the REF_ION keyword can set them all. I just didn't add it to the manual because Q isn't used anywhere.

Any reason why you want to set all the constants in the SIMU block? We already have a constants file where these are defined as parameters.

vkbo:
`REF_PDGID` keyword added, and `pdgid0` variable added to `mod_common`.

rdemaria:
I would not use REF_ION because it is not clear what is the ordering of the numbers, but have separate keywords.

vkbo:
I want to keep them together to make it easier to parse the options for these values, but we can of course change the keyword. I named it `REF_ION` because it's the only keyword that is actually about ions, and I took it from the `HION` block.

rdemaria:
Can we have a key=value list, such as

REF_PARTICLE MASS=939 CHARGE=-1 A=4 Z=2?

vkbo:
Not with the current parser.

rdemaria:
I would then use separate keywords for each value or, if too complicated, use REF_Q_A_Z. Q should go first because who wants to track without collimation does not need to set A and Z.

vkbo:
It's not complicated, but either Q or Z can be optional, depending on how you want to set up your simulation, but sure, I can reorder it, and set it up like that.

Z should then default to abs(Q)? or should it stay at 1?

rdemaria:
perhaps abs(Q) is better. REF_QAZ is a bit ugly. What about having at least REF_CHARGE and REF_A_Z?

amereghe:
let's not forget cases like H-

vkbo:
Updated the SIMU block interface to what we agreed on today between me, @amereghe and @rdemaria 

That is, we add the keyword `PDG_YEAR` to set which year to take standard values from, and we replace `REF_MASS` with `REF_PARTICLE`. The `REF_PARTICLE` keyword takes 1, 2 or 4 parameters: mass, charge, A, Z.

Mass can either be `proton`, `pdg2212` or a float. We can then easily extend it to other names and PDGs. Charge is set to the `Q` arrays, and `A` and `Z` is the same as the `HION` block.

Until James' changes are merged or added in other ways, setting `charge` just sets `Z`, unless `Z` is specified.

vkbo:
Note: I will make the final changes to the manual when we have agreed on the code. So I think we could start doing the code review now.

PR #874, opened 2019-05-09
Cleanup in mod_common_main

vkbo:
There were a number of redundant arrays in the main particle array module, so I cleaned up a bit. Also added comments about what all the arrays are.

A bit difficult to see from the diff, but the deleted arrays are: `xsv`, `zsv`, `dp0v`, `sigmv6`, `dpsv6`, `xlv`, `zlv`, `nms`, and `ekkv`.

`dp0v` was always the same as `dpsv`, `nms` was always `one`, and `xlv` and `zlv` could be replaced by scalars. The rest weren't used at all.

In addition, `dpd`, `dpsq` and `ampv` can probably also be removed with a bit of tweaking of the code. They are all used as local temp arrays. So is `ejf0v`, but this one is used between particle loops.

### DIFF
--- a/source/common_modules.f90
+++ b/source/common_modules.f90
@@ -834,26 +834,26 @@ module mod_common_main
 
   implicit none
 
-  ! Main 1
-  real(kind=fPrec), allocatable, save :: smiv(:)      ! (nblz)
-  real(kind=fPrec), allocatable, save :: zsiv(:)      ! (nblz)
-  real(kind=fPrec), allocatable, save :: xsiv(:)      ! (nblz)
+  ! Number of Structure Elements (nblz)
+  real(kind=fPrec), allocatable, save :: zsiv(:)      ! Displacement of elements, including error
+  real(kind=fPrec), allocatable, save :: xsiv(:)      ! Displacement of elements, including error
+  real(kind=fPrec), allocatable, save :: smiv(:)      ! Magnetic kick, including error
 
-  real(kind=fPrec), allocatable, save :: xsv(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: zsv(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: xv1(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: yv1(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: xv2(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: yv2(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: dam(:)       ! (npart)
-  real(kind=fPrec), allocatable, save :: ekkv(:)      ! (npart)
-  real(kind=fPrec), allocatable, save :: sigmv(:)     ! (npart)
-  real(kind=fPrec), allocatable, save :: dpsv(:)      ! (npart)
+  ! Number of Particles (npart)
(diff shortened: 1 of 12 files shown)


source/common_modules.f90
```diff
@@ -832,81 +834,69 @@ module mod_common_main
...
+  real(kind=fPrec), allocatable, save :: sigmv(:)      ! Longitudinal coordinate: Position offset
+  real(kind=fPrec), allocatable, save :: dpsv(:)       ! Longitudinal coordinate: Momentum offset from reference momentum e0f
+  real(kind=fPrec), allocatable, save :: ejfv(:)       ! Particle momentum
+  real(kind=fPrec), allocatable, save :: ejv(:)        ! Particle energy
+
+  real(kind=fPrec), allocatable, save :: oidpsv(:)     ! 1/(1+dpsv)
+  real(kind=fPrec), allocatable, save :: rvv(:)        ! Beta_0 / Beta(j)
+  real(kind=fPrec), allocatable, save :: ejf0v(:)      ! Temporary array for momentum updates
+  real(kind=fPrec), allocatable, save :: dam(:)        ! Distance in phase space3
+  real(kind=fPrec), allocatable, save :: dpd(:)        ! Tick tracking only: 1+dpsv
```
rdemaria: typo: tick ->thick

vkbo: Are you sure? Tick > tick > BOOM! :P

vkbo:
Just did a quick count. With this PR we have reduced the amount of memory per particle to 460 bytes without collimation and about 5 kb with collimation.

For main SixTrack, this is reduced by at least a factor 4.

source/common_modules.f90
```diff
@@ -832,81 +834,69 @@ module mod_common_main
...
+  real(kind=fPrec), allocatable, save :: yv2(:)        ! Transverse coordinates: Vertical angle
+  real(kind=fPrec), allocatable, save :: sigmv(:)      ! Longitudinal coordinate: Position offset
+  real(kind=fPrec), allocatable, save :: dpsv(:)       ! Longitudinal coordinate: Momentum offset from reference momentum e0f
+  real(kind=fPrec), allocatable, save :: ejfv(:)       ! Particle momentum
+  real(kind=fPrec), allocatable, save :: ejv(:)        ! Particle energy
+
+  real(kind=fPrec), allocatable, save :: oidpsv(:)     ! 1/(1+dpsv)
+  real(kind=fPrec), allocatable, save :: rvv(:)        ! Beta_0 / Beta(j)
+  real(kind=fPrec), allocatable, save :: ejf0v(:)      ! Temporary array for momentum updates
+  real(kind=fPrec), allocatable, save :: dam(:)        ! Distance in phase space3
```
kyrsjo: phase space*3*?

rdemaria:
Additional array involving sqrt are there to save calculations, but the others in particular the one of the tilt I am not sure the reason...

vkbo:
> Additional array involving sqrt are there to save calculations, but the others in particular the one of the tilt I am not sure the reason...

Of course, but that does not mean they have to be global arrays. They are only used as temporaries.

vkbo:
> Awesome! Thanks :)
> 
> I admit to not having looked too hard at [7538cfe](https://github.com/SixTrack/SixTrack/commit/7538cfe16f63e378d4dd0e5393ace28cf0af6d98) , but otherwise everything looks good to me.

Oh, that one only removes redundant spaces and moves the scalars and fixed arrays to the top of the module.

PR #877, opened 2019-05-20
Boinc API, bugfix and progress

vkbo:
This PR:

* Repositions the BOINC submodule to latest master where my fortran_api changes were merged.
* Fixes building for ifort/nagfor (didn't like `stop exitCode` so changed to `stop 1`).
* If postprocessing is on, tracking progress stops at 99%, and jumps to 100% after.
* Reshuffles a little bit how BOINC is finished due to the log file needing to be closed before ZIPF.

**Additional**

Do we really need two flags for BOINC and API? @kyrsjo asked the same question. This is why we need dummy routines, like `boincrf`, and it complicates the service module. I can't see any reason why we shouldn't just link BOINC with the API always. I suppose there was a reason for the test suite before perhaps, but the tests pass fine now.

### DIFF
--- a/source/mod_boinc.f90
+++ b/source/mod_boinc.f90
@@ -174,7 +174,7 @@ subroutine boinc_finalise(exitCode)
   if(exitCode == 0) then
     stop
   else
-    stop exitCode
+    stop 1
   end if
 #endif
 
--- a/source/mod_boinc.f90
+++ b/source/mod_boinc.f90
@@ -20,6 +20,7 @@ module mod_boinc
   real(kind=fPrec),   private, save :: boinc_progInterval = 0.0            ! Number of seconds between progress updates
   real(kind=fPrec),   private, save :: boinc_lastCPReq    = 0.0            ! CPU time of last checkpointing request
   real(kind=fPrec),   private, save :: boinc_lastProgress = 0.0            ! CPU time of last progress report to the API
+  real(kind=fPrec),   private, save :: boinc_trackTotal   = 1.0            ! Max progress count in tracking
 
 contains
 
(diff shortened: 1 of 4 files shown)


source/mod_boinc.f90
```diff
@@ -137,7 +146,8 @@ subroutine boinc_post
 
   use checkpoint_restart
 
-  write(boinc_logBuffer,"(a)") "Final checkpoint after tracking"
+  write(boinc_logBuffer,"(a)") "Tracking completed. Final checkpoint."
+  write(boinc_logBuffer,"(a,f8.3,a)") "Progress: ",99.0," %"
```
amereghe: I guess that if no postprocessing is given, this message should report 100%, shouldn't it?

source/mod_boinc.f90
```diff
@@ -151,21 +161,27 @@ subroutine boinc_post
...
-subroutine boinc_finalise(exitCode)
+subroutine boinc_done
 
   use mod_units
 
-  integer, intent(in) :: exitCode
+#ifdef API
+  call boinc_fraction_done(1.0)
+#endif
+  write(boinc_logBuffer,"(a,f8.3,a)") "Progress: ",100.0," %"
```
amereghe: following comment above, don't you want to report, instead of 100%, a statement like:"call to boinc done"?

source/mod_boinc.f90
vkbo: This write is redundant. Will remove it.

source/mod_boinc.f90
vkbo: No, this is better I think as it actually does reach 100% in the log file. Otherwise the progress counter may be left hanging on 99%.

PR #878, opened 2019-05-21
Boinc API, Step 2

vkbo:
Note: This one continues from #877, so the diff includes both until the other is merged.

This PR does the following:

* Added a hard coded particle state file dump from the BOINC service module after tracking and postprocessing.
* Removed simulation time information from `fort.10`. This will be handled by the summary file below. This means the `fort.10` file can be diffed or hashed as long as the version number matches.
* The dummy `fort.10` in `abend` now only contains one record.
* The BOINC service module writes a file called `boinc_summary.dat` that has tracking time, and at leas a hash of the particle state file + various other files depending on simulation setup . See example below.

These changes were originally a part of this PR, but the changes copied to other PRs:

* Removes the `API` compiler flag. No point building BOINC without the API any more.
* The timing module `mod_time` is now fully checkpoint/restart compliant, removing the need for the old time variables and routines. These have been deleted.
* Changes the way the `part_writeState` routine is called so we can use it as a general particle dump routine. The flags in the settings block for final state and initial state have been modified to fit the new interface.

The summary file example:
```
000004147 00073 050207 932d2ee8078a8c73479c26616a24d33b 2bee525a39f59a7a9d5d63a1ef8eb9ed
```

The format consists of the time in milliseconds, and a five digit count of the number of characters that need to be verified between work units. The part that must match between work units is the version number, and the hash of the output files. In this case, 73 starting from the character immediately after the value. That  is character 16 to 73+16

The file will always contain a hash of the final state of the particles. In addition, it will hash fort.10 if postprocessing is enabled, and fma_sixtrack if fma is enabled. More can be added in the `boinc_done` subroutine.


vkbo:
After an email discussion, I've updated the summary file to also contain the progress (from 0 to 100000) and an exit code of 1 in the event of a caught early exit. The file is now written, and overwritten multiple times so that it is always present as long as SixTrack run long enough to initialise the BOINC module. The hashes are added at the end. Before this, it will only contain the data up until the version number.

vkbo:
The checkpointing of `mod_time` has been extracted from this PR and merged in a separate PR #884 .

amereghe:
A couple of comments before going through the PR:
* I think that it would be great if we could also remove the SixTrack version frtom fort.10. That would allow us to validate results from different SixTrack versions, as it is presently the case on BOINC, i.e. running with 5.02.05 except the MacOS exes, compiled as 5.02.10;
* of course, we cannot merge this until we have the new validator :)

source/mod_boinc.f90
```diff
@@ -141,37 +162,185 @@ subroutine boinc_post
...
+! ================================================================================================ !
+!  Sets the post tracking progress.
+!  V.K. Berglyd Olsen, BE-ABP-HSS
+!  Created: 2019-05-21
+!  Updated: 2019-05-21
+!  These steps are hardcoded, and should only be changed if a time consuming post processing routine
+!  is added in main_cr. If so, bump the mSteps parameter and recheck all calls to this routine.
+!  Currently there are 5 steps:
+!   - Step 0 : Called after tracking, called from main_cr
+!   - Step 1 : Called after reports, fort.12 and final state file, called from main_cr
```
amereghe: I don't see the Step 1 (at least in the diff)

source/mod_boinc.f90
```diff
@@ -141,37 +162,185 @@ subroutine boinc_post
...
+  write(boinc_logBuffer,"(a,f8.3,a)") "Progress: ",100.0*progFrac," %"
+  call boinc_writeLog
+end subroutine boinc_postProgress
+
+! ================================================================================================ !
+!  Writes the validation file header, and creates the file. Hash values must be set before this.
+!  V.K. Berglyd Olsen, BE-ABP-HSS
+!  Created: 2019-05-22
+!  Updated: 2019-05-22
+! ================================================================================================ !
```
amereghe: Maybe, in the header of `boinc_summary` and `boinc_done`, please state that the two are linked

vkbo:
Format updated. Currently the file reads, for scatter_aperture test:
```
000009238 100000 0 00110 050301 000028915 000000050 000000003 000002000 0000028217 000000000732361101 0d0a851731e5aa526dc255277063b407
```
The entries are as follows:
1. CPU time in milliseconds
2. Progress from 0 to 100000
3. Exit status, 0 or 1
4. The number of characters to be validated counting from the end of this number
5. Executable version
6. Number of structure elements, `iu`
7. Number of turns, `numl`
8. Number of particles, `napx`
9. Number of original particles, `napxo`
10. Number of particle-turns
11. Number of particle-turn-elements
12. and onward, file hashes.

source/mod_boinc.f90
vkbo: You're right. It was missing. Fixed.

source/mod_boinc.f90
vkbo: Those were out of date. Rewritten the comments for boinc_summary and boinc_done.

source/checkpoint_restart.f90
```diff
@@ -390,7 +390,7 @@ subroutine crcheck
     write(crlog,"(a)") "CR_CHECK>  * Tracking variables"
     flush(crlog)
     read(cr_pntUnit(nPoint),iostat=ioStat) crnumlcr,crnuml,crsixrecs,crbinrec,cril,crnapxo, &
-      crnapx,cre0,crbetrel,crbrho,crnucmda
+      crnapx,cre0,crbeta0,crbrho,crnucmda
```
amereghe: I guess that the change from `betrel` to `beta0` is done elsewhere in the code, and only CR was missing, right?

source/postprocessing.f90
```diff
@@ -12,79 +12,34 @@ module postprocessing
...
-      real(kind=real64) dummy64
-
-      !For the actual tracking data
-      real(kind=real64) b_tmp,c_tmp,d_tmp,e_tmp,f_tmp,g_tmp,h_tmp,p_tmp
-      real(kind=real64) c1_tmp,d1_tmp,e1_tmp,f1_tmp,g1_tmp,h1_tmp,p1_tmp
-
-      character(len=80) title(20),chxtit(20),chytit(20)
-      character(len=8) cdate,ctime,progrm ! Note: Keep in sync with maincr
-      character(len=80) sixtit,commen     ! Note: Keep in sync with mod_common
-                                          ! DANGER: If the len changes, CRCHECK will break.
```
amereghe: given the recent problems of Frank with post-processing, are you sure that deleting this block is fine? the diff of github (only most recent changes) is not clear on this...

source/checkpoint_restart.f90
vkbo: Not sure why this shows up as the variable is already `crbeta0` in master.

source/postprocessing.f90
vkbo: Not sure how you found this diff. It's not in the main changed files. I did not delete this code.

vkbo: Not sure how you found this diff. It's not in the main changed files. I did not delete this code.

source/checkpoint_restart.f90
vkbo: This change also does not exist in the main diff.

PR #882, opened 2019-05-24
ZLIB/MiniZip as a replacement for libArchive

vkbo:
This PR adds the `ZLIB` flag to CMake, and builds the c code from the `zlib/contrib/minizip` folder as well as the interface in source/minizip.

This provides a much more lightweight way of making and extracting zip files for BOINC and the `ZIPF` module.

The code here is under a compiler flag, so we should be able to test it without interfering with regular production code. If it works fine, we should set the `ZLIB` flag to `ON` by default and remove libArchive altogether.

Been tested on: Debian 10 64-bit, CentOS 7 64-bit, Ubuntu 16.04.6 32- and 64-bit, Windows 10 32- and 64-bit and on MacOS 10.14 64-bit.


vkbo:
Both zipping and unzipping now works on Debian and Windows10. Need to test a few more builds.

vkbo:
Didn't build properly without zlib being build by the old script, but that is fixed now.

Also confirmed that it builds and passes tests on Ubuntu 16.04.6 both 32 and 64 bit, and on lxplus7.

PR #885, opened 2019-05-28
Collimation Distribution Generator (Step 3)

vkbo:
Note: This PR includes the changes from PR #796, so the diff is not very readable until that one is merged.

This PR:
* Moves all of the distribution generator routines out of the main collimation module and into a module named `coll_dist`. They have then been cleaned up, and a lot of duplicate code removed.
* Since the order in which the different steps of the generators are called have changed, the order at which random numbers are used changes. That means the distributions generated by a given seed will not be numerically equal before and after this PR.
* A total of 9 tests have been added for the generators, including histograms from running with 10000 particles. The tests themselves contain 64 to reduce the size of canonicals.
* Cleaned out a lot of now redundant particle arrays, including all the "my..." arrays and "grd..." arrays.


vkbo:
One thing we should look at is the need for separate random seed for coll gaps errors. The way it is done now is that the random number generator is first initialised with the standard collimation seed. Then it is reinitialised with the collgaps seed, and then reinitialised again with the first seed.

However, the reinitialisation with the proper seed does not restore the random number generator to the state it would have been in before the generation of coll gap errors, it is instead reset to the start. Since the random numbers are used for different things after this point, this means we will have all the random numbers used for generating distributions reintroduced into the pool of numbers. That does not sound good.

The Benefit of the reset is of course that the collgaps random numbers will never interfere with the flow of the other numbers if the feature is switched off. However, that can be achieved by either using the ranecu generator which can continue on a previously stopped stream of numbers, or we can just allocate the numbers needed for the gaps regardless, and just not apply them if the feature is disabled.

amereghe:
I think that the problem is a bit more general. Probably what we would like to have is the following:
* to set only one seed for the entire simulation, no matter if the user will use multipolar and/or collimator errors randomly distributed, and random initial distributions - this is for user convenience;
* to have fully reproducible sequences for the same input and exe, i.e. we record the state of the seed - this is for user convenience;
* to have fully reproducible sequences even if we add new features which require random sampling - this is for user and admin convenience.

I would propose the following:
* the user provides one and only one seed for everything - this is our starting point;
* after `daten` and once all the input has been parsed, we initialise instead of 2 seeds, 2 (allocatable) arrays of seeds. Each place in the arrays is dedicated to a specific use (e.g. multipolar errors, collgaps errors, etc...);
* the initialisation of the arrays is done starting from the initial seed and extracting random numbers. At every extraction, we save the new seeds into a new place in the two arrays. In order to smear the sequences, we should apply an algebraic operation to the new seeds - maybe as simple as a shift of the two integers by 1;
* the seed arrays should be organised such that:
   * we anyway initialise the seeds of all the possible uses, no matter if they will be used or not;
   * the seeds necessary to `DYNK` are initialised as last ones;
   * all new applications implemented later than this proposal have their seeds initialised after those necessary to DYNK.

What do you think?

source/collimation.f90
```diff
@@ -862,69 +684,46 @@ subroutine collimate_init()
...
-  write(lout,"(a,e15.8)") 'COLL> Info: MYDEY               = ', mdey
-  write(lout,"(a,a)")     'COLL> Info: FILENAME_DIS        = ', trim(filename_dis)
-  write(lout,"(a,e15.8)") 'COLL> Info: ENERROR             = ', enerror
-  write(lout,"(a,e15.8)") 'COLL> Info: BUNCHLENGTH         = ', bunchlength
+  write(lout,"(a,i0)")    'COLL> Info: DIST_TYPES          = ', do_thisdis
+  write(lout,"(a,e15.8)") 'COLL> Info: DIST_NEX            = ', cdist_ampX
+  write(lout,"(a,e15.8)") 'COLL> Info: DIST_DEX            = ', cdist_smearX
+  write(lout,"(a,e15.8)") 'COLL> Info: DIST_NEY            = ', cdist_ampY
+  write(lout,"(a,e15.8)") 'COLL> Info: DIST_DEY            = ', cdist_smearY
+  write(lout,"(a,a)")     'COLL> Info: DIST_FILES          = ', trim(cdist_fileName)
```
amereghe: maybe `DIST_FILE`?

source/collimation.f90
```diff
@@ -986,108 +785,81 @@ subroutine collimate_init()
...
+    write(outlun,*)   "WARN> Distributions reset to pencil beam!"
+    xv1(1:napx) = zero
+    yv1(1:napx) = zero
+    xv2(1:napx) = zero
+    yv2(1:napx) = zero
   endif
 
   ! Optionally write the generated particle distribution
-  if(dowrite_dist .and. do_thisdis /= 0) then
+  if(dowrite_dist) then
```
amereghe: why removing the check on `do_thisdis`?

source/collimation.f90
```diff
@@ -2077,12 +1848,10 @@ subroutine collimate_start_sample(nsample)
 
   ! Copy new particles to tracking arrays. Also add the orbit offset at start of ring!
```
amereghe: maybe please update comment - we are no longer copying :)

source/coll_dist.f90
```diff
@@ -0,0 +1,443 @@
...
+  end if
+
+  ejv(:)   = cdist_energy
+  sigmv(:) = zero
+
+end subroutine cdist_makeDist_fmt1
+
+! ================================================================================================ !
+!  Generation of Distribution Format 2
+!  Uses format 1, and generates the transverse beam size in the direction set to zero in fort.3
```
amereghe: very disgusting comment - please, change it saying that the distribution is Gaussian in the other plane
moreover, please mention that format 1 is called outside of this routine, and it is a kind of pre-requisite

source/coll_dist.f90
```diff
@@ -0,0 +1,443 @@
...
+  use mathlib_bouncer
+  use mod_ranlux
+  use mod_common, only : napx
+  use mod_common_main, only : xv1, xv2, yv1, yv2, ejv, sigmv
+
+  real(kind=fPrec) :: long_cut, a_st, b_st
+  integer          :: j
+
+  ! For longitudinal phase-space, add a cut at 2 sigma
+  long_cut = 2
```
amereghe: move it to header?

source/coll_dist.f90
```diff
@@ -0,0 +1,443 @@
...
+    call cdist_makeDist_fmt3
+  case(4)
+    call cdist_makeDist_fmt4
+  case(5)
+    call cdist_makeDist_fmt3
+    call cdist_makeDist_fmt5
+  case(6)
+    call cdist_makeDist_fmt4
+    call cdist_makeDist_fmt6
+  case default
```
amereghe: the naming convention could be reviewed, as (in the end) the actual format is accomplished only if you have all the calls performed regularly

source/mainarrays.f90
```diff
@@ -257,18 +257,6 @@ subroutine shuffleLostParticles
...
-      xpgrd(j:tnapx)                = cshift(xpgrd(j:tnapx),                1)
-      ypgrd(j:tnapx)                = cshift(ypgrd(j:tnapx),                1)
-      pgrd(j:tnapx)                 = cshift(pgrd(j:tnapx),                 1)
-      ejfvgrd(j:tnapx)              = cshift(ejfvgrd(j:tnapx),              1)
-      sigmvgrd(j:tnapx)             = cshift(sigmvgrd(j:tnapx),             1)
-      rvvgrd(j:tnapx)               = cshift(rvvgrd(j:tnapx),               1)
-      dpsvgrd(j:tnapx)              = cshift(dpsvgrd(j:tnapx),              1)
-      oidpsvgrd(j:tnapx)            = cshift(oidpsvgrd(j:tnapx),            1)
-      dpsv1grd(j:tnapx)             = cshift(dpsv1grd(j:tnapx),             1)
-
```
amereghe: we like it :dancer:

source/collimation.f90
vkbo: I see no reason why you shouldn't be able to dump the distribution even if you're not using the internal generator. You may still want to know what it looks like.

amereghe: yes, but then one may want to do this no matter if the collimation block has been issued or not

vkbo: Not sure I follow.

I added this check when I added format 0 as a valid option, but on second thought I don't really see any reason to block the user from generating the dist files when they have requested them (for whatever reason). This also blocks the user from dumping a radial dist.

source/collimation.f90
```diff
@@ -122,7 +116,13 @@ module collimation
...
   real(kind=fPrec), public,  save :: myemitx0_collgap = zero
   real(kind=fPrec), public,  save :: myemity0_collgap = zero
-  real(kind=fPrec), save :: myalphay, mybetay, myalphax, mybetax, rselect, myemitx
+
+  real(kind=fPrec), private, save :: myalphay
+  real(kind=fPrec), private, save :: mybetay
+  real(kind=fPrec), private, save :: myalphax
+  real(kind=fPrec), private, save :: mybetax
+  real(kind=fPrec), private, save :: rselect = 64.0 ! Not set anywhere, but used in if statements
+  real(kind=fPrec), private, save :: myemitx
```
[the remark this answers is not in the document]
vkbo: Not sure. Didn't look into these in great detail.
I moved around a lot of variables in the header in order to properly format them and set initial values, but not all of them are related to this PR.

CMakeLists.txt
```diff
@@ -295,8 +295,8 @@ list(APPEND FORT90_LIB
   bouncy_castle plato_seq bdex fma dabnew end_sixtrack
   ranecu wire elens mod_dist scatter dump dynk cheby utils
   aperture mod_fluc beam6d_fox six_fox postprocessing
-  read_write ranlux collimation coll_db coll_dist mod_particles
-  mod_ffield mod_ffield_aux
+  read_write ranlux collimation coll_db mod_particles cheby utils
+  mod_ffield mod_ffield_aux mod_find coll_dist
```
amereghe: `cheby` and `utils` were already in the list and are now doubled with your addition

vkbo: It's not an "addition" as much as it was a merge conflict. There is no harm in this, but I'll clean it up anyway.

PR #891, opened 2019-06-06
x,y,s - rotations

tpersson:
The rotations (also known as patches) have been implemented in both the track and fox as new elements. It is following the same logic as MAD-X and the conversion from MAD-X is already in the master branch. 

Eventually this can be used also from the WIRE element which is not doing the correct rotation (not symplectic).

Now the trigonometric functions are recalculated every time in the code. This could be calculated at initialization and stored. I am not sure if the compiler is smart enough to understand that it is is only needed to calculate it once..


rdemaria:
It would be nice to use precomputed quantities for performance. If easy I would do it now, since your are fresh in the code. Otherwise it can wait whenever some application needs a performance boost.

tpersson:
Sure!  It is a very quick thing to do.

doc/user_manual/chMachineGeometry.tex
```diff
@@ -489,8 +489,45 @@ \subsection{Beam Position Monitor} \label{BPM}
...
+\paragraph{Format} \texttt{name type angle}
+
+\bigskip
+\begin{tabular}{@{}lp{0.8\linewidth}}
+    \texttt{name}  & May contain up to 47 characters. \\
+    \texttt{type}  & 44 \\
+    \texttt{angle} & The rotation angle in radians.  
+\end{tabular}
+\paragraph{Remarks}
+A possetive angle rotates the reference system in the clockwise direction.
```
vkbo: "Positive" not "possetive"

tpersson:
Ops I made a git mistake here.. I will clean it up or make another PR, sorry for that.

vkbo:
> It would be nice to use precomputed quantities for performance. If easy I would do it now, since your are fresh in the code. Otherwise it can wait whenever some application needs a performance boost.

I doubt allocating three arrays of several thousand elements for calculating a few trigonometric functions gives a performance boost worth the additional clutter and memory caching.

Both @kyrsjo and I have put a good deal of effort into reducing the number of such arrays. I think it is a very bad idea to start adding new ones back in.

vkbo:
> Ops I made a git mistake here.. I will clean it up or make another PR, sorry for that.

Seems fine to me?

PR #893, opened 2019-06-07
Particle State Files Update

vkbo:
This PR adds the Q column to the particle state files. It also adds reference particle, closed orbit, tune and tas matrix to the header.

The routine to dump these files have also been generalised and can in principle be called by any part of the code by providing a file name and a few options.

Updated a lot of canonicals, but mostly just their headers.


rdemaria:
Do we need an update of the manual?

vkbo:
Hmm, probably useful. I'll fix it now.

vkbo:
Done.

PR #901, opened 2019-06-14
CMake changes for arm and osx

JamesMolson:


vkbo:
Is the clang bit only valid for Apple, or can we have it as a general section?

JamesMolson:
Apple only, normal clang explodes.

vkbo:
COVERAGE and GPROF flags don't apply either I assume?

JamesMolson:
gprof for sure not.

vkbo:
Checked the documentation. The Coverage flags are fine, but -pg does not exist.

JamesMolson:
um, they are not in the man pages

vkbo:
https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-fprofile-arcs

Assuming this is the correct documentation?

JamesMolson:
ok, it does work, but the fortran part explodes with it enabled so I can't test it. -pg explodes.


AppleClang != Clang. They forked it a long time ago and have done terrible dark things to it since.

vkbo:
Ah, I see. In that case.

vkbo:
Could you please just delete the flags that do not apply? Commented out code that doesn't work anyway is just clutter ...

PR #902, opened 2019-06-14
EQV Fix

vkbo:
.eqv. operator doesn't have the precedence parts of the code assumes.
Also removed a few that were a bit inconsistent.

### DIFF
--- a/source/coll_db.f90
+++ b/source/coll_db.f90
@@ -332,7 +332,7 @@ subroutine cdb_writeDB_ROOT
   character(len=5)          :: this_material = C_NULL_CHAR
   integer j
 
-  if(root_flag .eqv. .false. .or. root_CollimationDB /= 1) return
+  if((root_flag .eqv. .false.) .or. root_CollimationDB /= 1) return
 
   do j=1,cdb_nColl
     this_name     = trim(adjustl(cdb_cNameUC(j)))//C_NULL_CHAR
--- a/source/mod_ffield_aux.f90
+++ b/source/mod_ffield_aux.f90
@@ -509,7 +509,7 @@ contains
           enddo
         endif
 
-        if (CoefSave.eqv..False.) then ! Create new coef
+        if (CoefSave.eqv..false.) then ! Create new coef
           tpij_Ax(1,tlinex,sline)=expx
           tpij_Ax(2,tlinex,sline)=expy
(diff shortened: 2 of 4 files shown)


vkbo:
The precedence of .eqv. is indeed a little muddy.

http://jick.net/Manuals/FORTRAN/g77/Equivalence-Versus-Equality.html

source/mod_ffield_aux.f90
```diff
@@ -509,7 +509,7 @@ subroutine ReadVectPotCoeff(this,lun,norm,iErr)
           enddo
         endif
 
-        if (CoefSave.eqv..False.) then ! Create new coef
+        if (CoefSave.eqv..false.) then ! Create new coef
```
amereghe: what about `if (.not.CoefSave)` ?

source/mod_ffield_aux.f90
```diff
@@ -531,7 +531,7 @@ subroutine ReadVectPotCoeff(this,lun,norm,iErr)
...
```
amereghe: what about `if (.not.CoefSave)` ?

source/mod_ffield_aux.f90
```diff
@@ -553,7 +553,7 @@ subroutine ReadVectPotCoeff(this,lun,norm,iErr)
...
```
amereghe: what about `if (.not.CoefSave)` ?

vkbo:
> fine.
> not clear to me though if you wanted to purge `.eqv.` and `.neqv.` statements, though

I replaced the ones in mod_alloc because they were inconsistent, and I was the one who put them there. I left the ones in the fringe field code be because those modules are maintained by someone else.

.eqv. is technically the correct way to compare logicals though. But for simple if statements it's perhaps a bit overkill.

source/coll_db.f90
```diff
@@ -332,7 +332,7 @@ subroutine cdb_writeDB_ROOT
   character(len=5)          :: this_material = C_NULL_CHAR
   integer j
 
-  if(root_flag .eqv. .false. .or. root_CollimationDB /= 1) return
+  if((root_flag .eqv. .false.) .or. root_CollimationDB /= 1) return
```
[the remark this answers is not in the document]
vkbo: No, the one above is as it should be.

PR #903, opened 2019-07-04
Collimation DB New Format (Part 4)

vkbo:
This PR adds the new collimator database format, and a test to check that it produces the same results as the old format.


source/coll_db.f90
```diff
@@ -49,28 +53,28 @@ subroutine cdb_allocDB
...
-  call alloc(cdb_cRotation,           cdb_nColl, zero,    "cdb_cRotation")
-  call alloc(cdb_cBx,                 cdb_nColl, zero,    "cdb_cBx")
-  call alloc(cdb_cBy,                 cdb_nColl, zero,    "cdb_cBy")
-  call alloc(cdb_cFound,              cdb_nColl, .false., "cdb_cFound")
+  call alloc(cdb_cName,     mNameLen, cdb_nColl, " ",           "cdb_cName")
+  call alloc(cdb_cNameUC,   mNameLen, cdb_nColl, " ",           "cdb_cNameUC")
+  call alloc(cdb_cMaterial, 4,        cdb_nColl, " ",           "cdb_cMaterial")
+  call alloc(cdb_cFamily,             cdb_nColl, 0,             "cdb_cFamily")
+  call alloc(cdb_cNSig,               cdb_nColl, cdb_defColGap, "cdb_cNSig")
+  call alloc(cdb_cNSigOrig,           cdb_nColl, cdb_defColGap, "cdb_cNSigOrig")
```
amereghe: didn't we say that the default collgap was 0.0?

source/coll_db.f90
```diff
@@ -38,6 +41,7 @@ module coll_db
   ! Family Arrays
   character(len=:), allocatable, public, save :: cdb_famName(:)     ! Family name
   real(kind=fPrec), allocatable, public, save :: cdb_famNSig(:)     ! Family sigma
+  real(kind=fPrec), allocatable, public, save :: cdb_famNSigOrig(:) ! Family sigma
```
amereghe: maybe add a comment line saying that the original nSigma is useful for DYNKing the collimator settings - or do I wrongly remember?

source/coll_db.f90
```diff
@@ -484,19 +591,52 @@ subroutine cdb_writeFam
 
   integer j, famUnit
 
-  call f_requestUnit("coll_families.dat", famUnit)
-  call f_open(unit=famUnit,file="coll_families.dat",formatted=.true.,mode="w",status="replace")
+  call f_requestUnit("coll_families_dump.dat", famUnit)
```
amereghe: don't you want to have the name of the file in a variable in the module header?

source/coll_db.f90
```diff
@@ -484,19 +591,52 @@ subroutine cdb_writeFam
...
+!  Write collimator database to file
+! ================================================================================================ !
+subroutine cdb_writeDB
+
+  use mod_units
+
+  character(len=cdb_fNameLen) famName
+  integer j, dbUnit
+
+  call f_requestUnit("coll_db_dump.dat", dbUnit)
```
amereghe: don't you want to have the name of the file in a variable in the module header?

source/coll_db.f90
vkbo: I think so? Can't remember now. I will change it to zero anyway.

source/coll_db.f90
vkbo: I forgot now. Will handle all of that when I finish the last PR in this chain where the DYNK stuff is applied. I think I need to make some changes to this (internally) anyway.

source/coll_db.f90
vkbo: I don't really see the value in a moule-wide variable used only one place? I can make it a parameter in the subroutine, but that too seems a bit unnecessary.

PR #905, opened 2019-07-05
New DIST Block (Part 1)

vkbo:
Adding the FORMAT keyword to the DIST block with support for various column types and units. This is the first step of the DIST block upgrades preparing for linking with the dist generator library.

The format is backwards compatible with the old 14 column DIST files. If no FORMAT is specified, this file is assumed. Alternatively, the old format can be selected with the FORMAT specifier OLD_DIST.


test/thin6d_ions/fort.3
```diff
@@ -75,7 +75,8 @@ HION
 208 82 1.93687690162648124E+02 82
 NEXT
 DIST
-READ initial.dat
+! FORMAT ID none none x[m] y[m] none xp[rad] yp[rad] none ion_a ion_z mass[GeV] P[GeV] dT[s]
```
rdemaria: If xp[rad] is Px/P the units are not radians.

source/mod_dist.f90
```diff
@@ -98,159 +231,586 @@ subroutine dist_parseInputLine(inLine, iLine, iErr)
...
+
+  character(len=*), intent(in)    :: fmtName
+  character(len=*), intent(in)    :: fmtUnit
+  integer,          intent(in)    :: unitType
+  real(kind=fPrec), intent(out)   :: unitScale
+  logical,          intent(inout) :: uErr
+
+  unitScale = zero
+
+  if(unitType == 1) then ! Positions
```
rdemaria: How do you specify multipliers for unitless quantities. Like delta in 1e-3?

source/mod_dist.f90
```diff
@@ -98,159 +231,586 @@ subroutine dist_parseInputLine(inLine, iLine, iErr)
...
+  case("YP") ! Vertical angle
+    call dist_unitScale(fmtName, fmtUnit, 2, uFac, fErr)
+    call dist_appendFormat(dist_fmtYP, uFac)
+  case("PX") ! Horizontal momentum
+    call dist_unitScale(fmtName, fmtUnit, 3, uFac, fErr)
+    call dist_appendFormat(dist_fmtPX, uFac)
+  case("PY") ! Vertical momentum
+    call dist_unitScale(fmtName, fmtUnit, 3, uFac, fErr)
+    call dist_appendFormat(dist_fmtPY, uFac)
+  case("PX/P0","PXP0") ! Relative horizontal momentum
```
rdemaria: I find "PXP0" is a bit confusing, but I see the need. Would it be possible to use P0 as unit, like PX[P0]?

source/mod_dist.f90
```diff
@@ -20,45 +23,138 @@
...
+  integer, parameter :: dist_fmtPX          = 15 ! Horizontal momentum
+  integer, parameter :: dist_fmtPY          = 16 ! Vertical momentum
+  integer, parameter :: dist_fmtPXP0        = 17 ! Relative horizontal momentum
+  integer, parameter :: dist_fmtPYP0        = 18 ! Relative vertical momentum
+  integer, parameter :: dist_fmtSIGMA       = 19 ! Longitudinal relative position
+  integer, parameter :: dist_fmtDT          = 20 ! Time delay
+  integer, parameter :: dist_fmtE           = 21 ! Particle energy
+  integer, parameter :: dist_fmtP           = 22 ! Particle momentum
+  integer, parameter :: dist_fmtDEE0        = 23 ! Relative particle energy (to reference particle)
+  integer, parameter :: dist_fmtDPP0        = 24 ! Relative particle momentum (to reference particle)
```
rdemaria: we also need DE/P0c which is the value that comes from mad

source/mod_dist.f90
```diff
@@ -20,45 +23,138 @@
...
+
+  ! Normalised Coordinates
+  integer, parameter :: dist_fmtX_NORM      = 31 ! Normalised horizontal position
+  integer, parameter :: dist_fmtY_NORM      = 32 ! Normalised vertical positiom
+  integer, parameter :: dist_fmtXP_NORM     = 33 ! Normalised horizontal angle
```
rdemaria: I would call 
`dist_fmtPX_NORM     = 33 ! Normalised horizontal momentum`

source/mod_dist.f90
```diff
@@ -20,45 +23,138 @@
...
+  integer, parameter :: dist_fmtYP_NORM     = 34 ! Normalised vertical angle
+  integer, parameter :: dist_fmtSIGMA_NORM  = 35 ! Normalised longitudinal relative position
```
rdemaria: I would call
`dist_fmtZETA_NORM  = 35 ! Normalised longitudinal relative position`

source/mod_dist.f90
```diff
@@ -290,58 +850,41 @@ subroutine dist_finaliseDist
...
+  call f_open(unit=fUnit,file=dist_echoFile,mode='w',err=cErr,formatted=.true.)
   if(cErr) goto 19
 
-  rewind(dist_echoUnit)
-  write(dist_echoUnit,"(a,1pe25.18)") "# Total energy of synch part [MeV]: ",e0
-  write(dist_echoUnit,"(a,1pe25.18)") "# Momentum of synch part [MeV/c]:   ",e0f
-  write(dist_echoUnit,"(a)")          "#"
-  write(dist_echoUnit,"(a)")          "# x[mm], y[mm], xp[mrad], yp[mrad], sigmv[mm], ejfv[MeV/c]"
+  rewind(fUnit)
+  write(fUnit,"(a,1pe25.18)") "# Total energy of synch part [MeV]: ",e0
```
rdemaria: I would replace "synch" with "reference"

test/thin6d_ions/fort.3
vkbo: This is a comment, will delete the line.

source/mod_dist.f90
vkbo: You don't.

source/mod_dist.f90
vkbo: As discussed, we'll keep it this way.

source/mod_dist.f90
vkbo: Good idea to support all Mad-X formats, so I'll add this under the keywords `DE/P0C` and `DEP0C`.

source/mod_dist.f90
vkbo: Will add this as well.

source/mod_dist.f90
vkbo: Sure. I can review the echo file in general. The 1pe25.18 format is also a decimal too many.

source/mod_dist.f90
```diff
@@ -98,159 +265,726 @@ subroutine dist_parseInputLine(inLine, iLine, iErr)
...
+  case default
+    write(lerr,"(a)") "DIST> ERROR Unknown column format '"//trim(fmtName)//"'"
+    fErr = .true.
+
+  end select
+
+end subroutine dist_setColumnFormat
+
+! ================================================================================================ !
+!  Parse File Column Units
```
amereghe: Maybe make the comment a bit more descriptive of what you do here, i.e. find the proper unit conversion

source/mod_dist.f90
```diff
@@ -98,159 +265,726 @@ subroutine dist_parseInputLine(inLine, iLine, iErr)
...
+    case("[mrad]")
+      unitScale = one
+    case("[1]")
+      unitScale = one
+    case default
+      goto 100
+    end select
+  elseif(unitType == 3) then ! Energy/Momentum/Mass
+    select case(chr_toLower(fmtUnit))
+    case("[ev]","[ev/c]","[ev/c^2]")
```
amereghe: shall we add also `[ev/c2]` and `[ev/c**2]`?

source/mod_dist.f90
```diff
@@ -98,159 +265,726 @@ subroutine dist_parseInputLine(inLine, iLine, iErr)
...
+    call dist_appendFormat(dist_fmtMASS,     c1e3, 0)
+    call dist_appendFormat(dist_fmtCHARGE,   one,  0)
+    call dist_appendFormat(dist_fmtIonA,     one,  0)
+    call dist_appendFormat(dist_fmtIonZ,     one,  0)
+    call dist_appendFormat(dist_fmtPDGID,    one,  0)
+
+  case("OLD_DIST") ! The old DIST block file format
+    call dist_appendFormat(dist_fmtPartID,   one,  0)
+    call dist_appendFormat(dist_fmtParentID, one,  0)
+    call dist_appendFormat(dist_fmtNONE,     one,  0)
```
amereghe: no, please, do not kill the parsing of the statistical weight.
In the context of the coupling, the statistical weight is currently operational

source/mod_dist.f90
```diff
@@ -20,45 +23,172 @@
...
+  logical, private, save :: dist_norm6D     = .false.
+
+  ! Arrays to send to DISTlib
+  real(kind=fPrec), allocatable, private, save :: dist_partCol1(:)    ! Ends up in array xv1
+  real(kind=fPrec), allocatable, private, save :: dist_partCol2(:)    ! Ends up in array yv1
+  real(kind=fPrec), allocatable, private, save :: dist_partCol3(:)    ! Ends up in array xv2
+  real(kind=fPrec), allocatable, private, save :: dist_partCol4(:)    ! Ends up in array yv2
+  real(kind=fPrec), allocatable, private, save :: dist_partCol5(:)    ! Ends up in array sigmv
+  real(kind=fPrec), allocatable, private, save :: dist_partCol6(:)    ! Ends up in array dpsv, evj, and ejfv
+  integer,                       private, save :: dist_partFmt(6) = 0 ! The format used for each column
```
amereghe: in order to be synched with `dist_postParticles`, I think you should initialise `dist_partFmr(6)=dist_fmtNONE`

source/mod_dist.f90
```diff
@@ -98,159 +265,726 @@ subroutine dist_parseInputLine(inLine, iLine, iErr)
...
+
+  logical doAction, doNormal
+
+  doAction = .false.
+  doNormal = .false.
+
+  select case(dist_partFmt(6))
+  case(dist_fmtNONE)
+    ejv(1:napx)  = e0
+    call part_updatePartEnergy(1,.false.)
```
amereghe: why not `call part_updatePartEnergy(dist_updtEFrom,.false.)`?

source/mod_particles.f90
```diff
@@ -108,30 +108,25 @@ end subroutine part_updateRefEnergy
...
-  logical, optional, intent(in) :: updateAngle
-
-  logical :: doUpdateAngle = .false.
+  integer, intent(in) :: refArray
+  logical, intent(in) :: updateAngle
 
   !if(part_isTracking .and. refArray /= 1) then
   !  write(lerr,"(a)") "PART> ERROR During tracking, only energy updates are allowed in part_updatePartEnergy."
   !  call prror
   !end if
```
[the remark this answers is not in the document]
vkbo: We can't yet. It would break Beam-Beam. The code for it needs a rewrite.

source/mod_dist.f90
vkbo: Sure, can do.

source/mod_dist.f90
vkbo: Ah, I'll add it back in.

source/mod_dist.f90
vkbo: That is the intention, yes. Still not finished the rewriting after the latest round of changes.

source/mod_dist.f90
vkbo: dist_updtEFrom will be deleted. After the last rewrite it is redundant.

source/mod_dist.f90
vkbo: Yeah, it's just out of date. Units weren't there at the start.

source/mod_dist.f90
```diff
@@ -20,45 +23,172 @@
...
+  ! ================
+  integer, parameter :: dist_fmtNONE        = 0  ! Column ignored
+  integer, parameter :: dist_fmtPartID      = 1  ! Paricle ID
+  integer, parameter :: dist_fmtParentID    = 2  ! Particle parent ID (for secondary particles, otherwise equal particle ID)
+
+  ! Physical Coordinates
+  integer, parameter :: dist_fmtX           = 11 ! Horizontal position
+  integer, parameter :: dist_fmtY           = 12 ! Vertical positiom
+  integer, parameter :: dist_fmtXP          = 13 ! Horizontal momentum ratio
+  integer, parameter :: dist_fmtYP          = 14 ! Vertical momentum ratio
```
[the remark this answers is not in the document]
vkbo: Done

source/mod_dist.f90
```diff
@@ -20,45 +23,172 @@
...
+  integer, parameter :: dist_fmtPXP0        = 17 ! Relative horizontal momentum
+  integer, parameter :: dist_fmtPYP0        = 18 ! Relative vertical momentum
+  integer, parameter :: dist_fmtZETA        = 19 ! Longitudinal relative position (canonical)
+  integer, parameter :: dist_fmtSIGMA       = 20 ! Longitudinal relative position
+  integer, parameter :: dist_fmtDT          = 21 ! Time delay
+  integer, parameter :: dist_fmtE           = 22 ! Particle energy
+  integer, parameter :: dist_fmtP           = 23 ! Particle momentum
+  integer, parameter :: dist_fmtDEE0        = 24 ! Relative particle energy (to reference particle)
+  integer, parameter :: dist_fmtDPP0        = 25 ! Relative particle momentum (to reference particle)
+  integer, parameter :: dist_fmtPT          = 26 ! Delta enery over reference momentum (Pt)
```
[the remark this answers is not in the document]
vkbo: Done

source/mod_dist.f90
vkbo: What do you mean by "operational", in the old block the value was discarded. This is why I set it to NONE.

vkbo:
> For the moment just some minor changes. On the other hand, it is not clear to me if `FORMAT` with a keyword different from `IONS` properly sets the pdgid...
> Flexible implementation, thanks a lot!
> 
> I haven't gone through the C part, though

Thanks. Went through and fixed the issues pointed out.
There is no C part to this module.

amereghe:
> Not sure what you mean. PDGID is available as a column format and also automatically included with IONS.

Yes, imagine that the user requests the FORMAT for 6D-normalised coordinates. I don't think we have the place in the DIST block where pdgid is set to that of proton

vkbo:
> > Not sure what you mean. PDGID is available as a column format and also automatically included with IONS.
> 
> Yes, imagine that the user requests the FORMAT for 6D-normalised coordinates. I don't think we have the place in the DIST block where pdgid is set to that of proton

That is set in the `SIMU` block, or if the old blocks are used, it belongs in the `HION` block.

vkbo:
To clarify: The `pdgid` array defaults to `pdgid0`, which again defaults to `2212`. All of this is set in `common_modules.f90`.

source/mod_particles.f90
```diff
@@ -56,7 +55,7 @@ subroutine part_applyClosedOrbit
     xv2(1:napx)   = xv2(1:napx)   +   clo(2)*real(idz(2),fPrec)
     yv2(1:napx)   = yv2(1:napx)   +  clop(2)*real(idz(2),fPrec)
```
tpersson: Not sure about the units of the clo(2) is it x'?

vkbo: I think it is. The code here was moved from main_cr some time last year.

vkbo:
We will merge the changes made at this point even if the module needs more development. merging now makes it easier for @tpersson to work on DISTlib and use the new input parsing.

The block is backwards compatible with the old DIST block, so while we develop, we will keep the new features undocumented until it's stable.

PR #913, opened 2019-07-09
A step towards the dist block

tpersson:
This  PR only contains the part of reading a file and converting coordinates to the correct ones. At the moment only physical coordinates are available since there is no support for tas matrix. However, all this is already done and just need to be fitted into this slightly modified structure. It was done in this way to be able to have several smaller PR. 

It is still not linked with SixTrack since there a few things still to discuss.


vkbo:
Hmm, this is not a clean diff. It contains several changes made by both me and @JamesMolson  ...

tpersson:
The call from sixtrack could then look something like 
     call dist_initializedistribution(1)
      call dist_sete0andmass0(mass0,energy0 )
      call dist_setemitt12(e1,e2)
      call dist_setemitt3(e3)
      call dist_settasmatrix(tas)
      call dist_setnormalizedcoords(xn,xnp,yn,ynp,zn,znp,npart)
      call dist_get6trackcoord(x,xp,y,yp,sigma,deltap, npart)

What do you think @vkbo ?

vkbo:
Closing this PR as DISTlib now has its own repository, and PR #905 adds it as a submodule of SixTrack.

PR #916, opened 2019-07-09
Add arrays for future spin tracking.

JamesMolson:


rdemaria:
Ok for me in principle. Can you add a compilation flags around them such that they don't eat memory until there is actually physics using them?

vkbo:
I'll add the flag when I add the parsing in the DIST block for these columns.

vkbo:
This will be merged after 5.3 release

PR #919, opened 2019-07-12
General particle transport for FLUKA

JamesMolson:
This starts to add support for sending anything between SixTrack <--> FLUKA.

Do not merge (yet).

This needs to be merged at the same time as a number of other repos:

The FLUKA IO repo
The FLUKA coupling repo

otherwise everything will explode, since the communication is changing.

TODO here: make the IO repo a submodule.


JamesMolson:
There probably also should be a fortran interface defined to the IO C functions.

JamesMolson:
This is done for now

PR #920, opened 2019-07-12
Kill libArchive

vkbo:


rdemaria:
Ok for me. Any other feedback?

source/zipf.f90
```diff
@@ -150,7 +141,7 @@ subroutine zipf_dozip
     write(lerr,"(a,i0)") "ZIPF> WARNING MiniZip returned error code ",iErr
   end if
 #else
-  write(lout,"(a)") "ZIPF> *** No Archive Libraries in this SixTrack *** "
+  write(lout,"(a)") "ZIPF> *** No ZLIB in this SixTrack *** "
```
kyrsjo: Is there a reason why this is not a fatal error? It's a bit late (if I remember correctly it comes after all the computation, and it is checked for in the input parsing, but still... That would make it more likely that the relevant error message is actually spotted.

PR #922, opened 2019-07-13
Further Manual Updates

vkbo:
Just looked at the variables definitions in the DUMP section.

I added the scaling factor for xp & yp, but for consistency, I removed those and updated the units instead. Also, the array variable for momentum was wrong.

The delta array should also be scaled by 1000, right?

Also, if ejfv is "Pc", shouldn't e0f be "P0 c"?


rdemaria:
At least internally dpsv should be unscaled if I remember correctly....

vkbo:
You're right. It's in the tas matrix the factor 1000 comes in, not the array itself.

Although dpsv1 is scaled.

rdemaria:
I would use the following rule:

Value  = Formula  / Unit

ejfv  [MeV/c]  $P$  means that if I take P in SI units and MeV/c in SI units P/(MeV/c) gives the numerical value in ejfv.

This works also for xv

xv  [1/1000] Px/P  means that xv=(Px/P)/(1/1000)

PR #923, opened 2019-07-16
G4 update + output formatting

JamesMolson:
Add the spin input/output to G4.
Fix the alignment of a few headers, including the aperture loss file.


source/dump.f90
```diff
@@ -1874,6 +1874,71 @@ subroutine dump_beam_population(nturn, i, ix, unit, fmt, lhighprec, loc_clo, tas
...
+    if(i == 0 .and. ix == 0) then
+      localDcum   = zero
+      localKtrack = 0
+    else
+      localDcum   = dcum(i)
+      localKtrack = ktrack(i)
+    end if
+#ifdef HDF5
+    if(h5_useForDUMP) then
+      call h5_prepareWrite(dump_hdf5DataSet(ix), napx)
```
vkbo: Yeah, this will not work. You have to create the dataset first in the routine that initialises all files. Here, `dump_hdf5DataSet(ix)` is probably 0.

PR #929, opened 2019-07-29
[WIP] Scatter updates

vkbo:
This PR updates to a newer version of Pythia which can now take momentum vectors directly into the event generator's `next()` call. To take advantage of this, two new profile types have been added to allow for collisions with particles at reference energy and sample a full beam profile.

The beam profile can be sampled by a beam that is the mirror of beam one, or with a given sett of twiss (beta, alpha, crossing angle and offset). Emittance is set separately for beam 2 in the main scatter block.

In order to extract the twiss parameters, a hook has been added into scatter from the linopt routines. In the process, the three main linopt routines were moved to their own file named `linear_optics.f90`.

Also, in the process, a number of `two*pi` were replaced with the numerical constant `twopi`.

Additional information has also been added to the `scatter_summary.dat` file, and a more exact calculation of changes to particle orbit from scattering angle has been added. The old calculation assumed small angles, which is not always true.

**ToDo:**

* Complete the tests for the Pythia implementation. The tests are mostly ready, but need to check that the results are reasonable.
* Update documentation.


tpersson:
I just wanted to see what was changed.. Didn't want to start the review if it wasn't ready ..

vkbo:
No problem. I just added `[WIP]` to the title.

vkbo:
Closing this PR and making a new one from scratch.

PR #934, opened 2019-08-05
Compact DUMP TAS

vkbo:
This PR replaces the `dumptas`, `dumptasinv`, and `dumpclo` arrays in `DUMP` with an array of structs that is expanded only when there is an element with a dump assigned to it. An integer map points between the element index and the storage array index.

This saves quite a bit of memory. For n FCC test with 60 particles it saves 33%, and for a HL-LHC test with 2000 particles it save 24%.

### DIFF
--- a/source/dump.f90
+++ b/source/dump.f90
@@ -124,7 +124,7 @@ end subroutine dump_expand_arrays
 subroutine dump_setTasMatrix(elemID, tasData, cloData)
 
   use crcoall
-  use mod_common, only : bezs
+  use mod_common, only : bez
   use mod_settings
 
   integer,          intent(in) :: elemID
@@ -149,7 +149,7 @@ subroutine dump_setTasMatrix(elemID, tasData, cloData)
     storeID = dump_nNormMat
     if(st_debug) then
       if(elemID > 0) then
-        write(lout,"(a)") "DUMP> Saving normalisation matrix for element '"//trim(bezs(elemID))//"'"
+        write(lout,"(a,i0,a)") "DUMP> Saving normalisation matrix for element ",elemID," '"//trim(bez(elemID))//"'"
       else if(elemID == -1) then
         write(lout,"(a)") "DUMP> Saving normalisation matrix for element 'StartDUMP'"
       end if
@@ -158,7 +158,7 @@ subroutine dump_setTasMatrix(elemID, tasData, cloData)
     storeID = dump_nMatMap(elemID)
     if(st_debug) then
       if(elemID > 0) then
-        write(lout,"(a)") "DUMP> Updating normalisation matrix for element '"//trim(bezs(elemID))//"'"
+        write(lout,"(a,i0,a)") "DUMP> Updating normalisation matrix for element ",elemID," '"//trim(bez(elemID))//"'"
       else if(elemID == -1) then
         write(lout,"(a)") "DUMP> Updating normalisation matrix for element 'StartDUMP'"
       end if
@@ -208,7 +208,8 @@ subroutine dump_lines(n,i,ix)
     ! Dump at all SINGLE ELEMENTs
     if (ndumpt(0) == 1 .or. mod(n,ndumpt(0)) == 1) then
       if ((n >= dumpfirst(0)) .and. ((n <= dumplast(0)) .or. (dumplast(0) == -1))) then
-        call dump_beam_population(n, i, ix, dumpunit(0), dumpfmt(0), ldumphighprec, dumpclo(ix,1:6),dumptasinv(ix,1:6,1:6))
+        call dump_beam_population(n, i, ix, dumpunit(0), dumpfmt(0), ldumphighprec, &
+          dump_normMat(dump_nMatMap(ix))%orbit, dump_normMat(dump_nMatMap(ix))%invtas)
       end if
     end if
   end if
@@ -218,7 +219,8 @@ subroutine dump_lines(n,i,ix)
       ! Dump at this precise SINGLE ELEMENT
       if (ndumpt(ix) == 1 .or. mod(n,ndumpt(ix)) == 1) then
         if ((n >= dumpfirst(ix)) .and. ((n <= dumplast(ix)) .or. (dumplast(ix) == -1))) then
-          call dump_beam_population(n, i, ix, dumpunit(ix), dumpfmt(ix), ldumphighprec, dumpclo(ix,1:6),dumptasinv(ix,1:6,1:6))
+          call dump_beam_population(n, i, ix, dumpunit(ix), dumpfmt(ix), ldumphighprec, &
+            dump_normMat(dump_nMatMap(ix))%orbit, dump_normMat(dump_nMatMap(ix))%invtas)
         end if
       end if
     end if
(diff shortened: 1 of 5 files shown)


source/dump.f90
```diff
@@ -78,37 +67,121 @@ module dump
...
+!  Updated: 2019-08-05
+!  Set the normalisation matrix, compute its inverse, and save it in a compact array of structs.
+! ================================================================================================ !
+subroutine dump_setTasMatrix(elemID, tasData, cloData)
+
+  use crcoall
+  use mod_common, only : bez
+  use mod_settings
+
+  integer,          intent(in) :: elemID
```
kyrsjo: Point out somewhere that this is the *single* element ID, not structure.

source/fma.f90
```diff
@@ -169,10 +169,11 @@ subroutine fma_postpr
   real(kind=fPrec), allocatable :: nxyzv(:,:,:)
   ! Normalised emittances
   real(kind=fPrec), allocatable :: epsnxyzv(:,:,:)
+  real(kind=fPrec) dumptas(6,6), dumpclo(6), dumptasinv(6,6)
```
kyrsjo: Describe the temp variable in a comment, as is done with the other variables.

kyrsjo: Just "Temp variables for loading normalization matrices" would suffice fine.

source/main_cr.f90
```diff
@@ -731,20 +731,16 @@ program maincr
 
   ! save tas matrix and closed orbit for later dumping of the beam
   ! distribution at the first element (i=-1)
-  ! dumptas(*,*) [mm,mrad,mm,mrad,1] canonical variables
```
kyrsjo: Keep this information where `type, public :: dump_normType` is defined?

kyrsjo: I.e. just the units etc. -- the "matrix is different for each particle" is outdated now, right?

source/six_fox.f90
```diff
@@ -2771,55 +2771,56 @@ end subroutine clorda
 !           note: inversion method copied from subroutine postpr        *
 !-----------------------------------------------------------------------*
 subroutine invert_tas(fma_tas_inv,fma_tas)
```
kyrsjo: Note the comment -- this subroutine is copied from postpr, and should probably be shared somehow...

source/dump.f90
```diff
@@ -78,37 +67,121 @@ module dump
...
+    dump_nMatMap(elemID) = dump_nNormMat
+    storeID = dump_nNormMat
+    if(st_debug) then
+      if(elemID > 0) then
+        write(lout,"(a,i0,a)") "DUMP> Saving normalisation matrix for element ",elemID," '"//trim(bez(elemID))//"'"
+      else if(elemID == -1) then
+        write(lout,"(a)") "DUMP> Saving normalisation matrix for element 'StartDUMP'"
+      end if
+    end if
+  else
```
kyrsjo: Why is this even allowed? Seems like a bug (uncaught input error) if it happens - the results won't be valid...

source/include/umlalid.f90
```diff
@@ -93,34 +94,34 @@
...
-      dumptas(ic(i)-nblo,ii  ,i3  )=au(i3  ,i3  )
-!    closed orbit in canonical variables x,px,y,py,sig,delta [mm,mrad,mm,mrad,mm,1.e-3]
-!    convert to x,xp,y,yp,sig,delta [mm,mrad,mm,mrad,mm,1]
-!     -> check units used in dumpclo (is x' or px used?)
-      dumpclo(ic(i)-nblo,2*j-1)=c(j)
-      if (j.eq.3) then !dp/p
-        dumpclo(ic(i)-nblo,2*j)  =cp(j)*c1m3
+  ! Store tas matrix (normalisation of phase space) and closed orbit for FMA and DUMP normalization.
+  ! Variable added to DUMP block module variables;
+  ! Units tasData: mm,mrad,mm,mrad,mm,1.e-3 -> convert later to 1.e3
```
kyrsjo: Is this converted to 1e3 or 1e0?

source/main_cr.f90
vkbo: No idea what you're referring to here. Please explain.

source/dump.f90
vkbo: I added it originally because the setting of the matrix was called three times for each element due to being in an existing loop with the other code. Now that I deleted the old code, the call is outside the loop in a different location than before.

In any case, the old code did not prevent the overwriting of the values at saving time, so I just replicated the behaviour for compact storage. I was trying to to impose new restrictions on the existing logic because it would trigger more work in rewriting DUMP.

source/include/umlalid.f90
vkbo: ???

vkbo:
> Very nice! Mostly just comments that I think are needed to be added/updated, plus one potential logic change (the "overwrite" behaviour in case there are multiple tas for one element).

Didn't want to alter the logic at this point. Just replace the storage arrays. I think this should be noted for further updates to the DUMP module. Also, there is no longer a need to restrict the dump to single elements since we're now using compact arrays.

But again, logic change should be done in connection with a full review of the DUMP module.

> Regarding the inversion routine, I just wanted to remind that this is one of the duplicate routines that would be nice to clean up at some point.

Make an issue?

source/dump.f90
vkbo: As far as I can see, there is nothing that checks that the element used for normalised dump is unique. We should probably require that for normalised dumps. Not sure where is the best place to add that. Adding it here also means we have to check for dump format, otherwise we will block all non-unique elements.

source/main_cr.f90
kyrsjo: I meant to put the description about the units (corrected if necessary) into where the dump_normType is described, as there is no precise description of the variables there now.

source/dump.f90
kyrsjo: Yeah, that's not really ideal in DUMP. At least an almost-bug :)

Maybe it's OK to just let it crash here then, with a suggestive error message?

source/include/umlalid.f90
kyrsjo: See line 161 to 164

vkbo: Just deleted the two comment lines as the information is duplicated anyway. Corrected the comment further down.

source/main_cr.f90
vkbo: Cleaned up these comments anyway, and added a note to check conversion to canonicals later. Something doesn't look quite right.

source/fma.f90
vkbo: Done

source/dump.f90
vkbo: Done

kyrsjo:
Regarding invert_tas, I don't know exactly where everything is anymore, so it's better if you do it (or not). I saw it was mentioned (as in moved around) in PR #616 also.

vkbo:
I was having a chat with @amereghe earlier about cleaning up the utilities routines a bit and move them into a set of tools modules. I think we can include this one then.

source/dump.f90
vkbo: We cannot restrict this it turns out. The routine where this is saved from is called during 6D closed orbit search, so potentially multiple times. The functionality to allow updating the values need to stay until we look into this in more detail. Otherwise it breaks existing functionality.

kyrsjo:
Thanks!

PR #935, opened 2019-08-06
Bugfixes for Compact TAS PR

vkbo:
Forgot to check that the index is non-zero when reading the tas matrix in the actual dump routines. These now use the dum_getTasMatrix call which does the check automatically.

Also, when cleaning up the header, forgot to check that the subroutines had the use statements it needed also under the CR flag.

Additional note: The tas matrix should probably be saved in a global module instead of under DUMP. especially since it is already shared with FMA. Will probably move this when PR #929 is merged, because there all the linopt stuff is moved to a separate file.

### DIFF
--- a/source/dump.f90
+++ b/source/dump.f90
@@ -187,21 +187,15 @@ subroutine dump_lines(n,i,ix)
   use mod_common_track
 
   integer, intent(in) :: n,i,ix
-  real(kind=fPrec) invTas(6,6), cloOrb(6)
+  real(kind=fPrec) tmpTas(6,6), tmpInv(6,6), tmpClo(6)
 
-  if(dump_nMatMap(ix) > 0) then
-    invTas = dump_normMat(dump_nMatMap(ix))%invtas
-    cloOrb = dump_normMat(dump_nMatMap(ix))%orbit
-  else
-    invTas = zero
-    cloOrb = zero
-  end if
+  call dump_getTasMatrix(ix, tmpInv, tmpTas, tmpClo)
 
   if(ldump(0)) then
     ! Dump at all SINGLE ELEMENTs
     if(ndumpt(0) == 1 .or. mod(n,ndumpt(0)) == 1) then
       if ((n >= dumpfirst(0)) .and. ((n <= dumplast(0)) .or. (dumplast(0) == -1))) then
-        call dump_beam_population(n, i, ix, dumpunit(0), dumpfmt(0), ldumphighprec, cloOrb, invTas)
+        call dump_beam_population(n, i, ix, dumpunit(0), dumpfmt(0), ldumphighprec, tmpClo, tmpInv)
       end if
     end if
   end if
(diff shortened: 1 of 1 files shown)


source/dump.f90
```diff
@@ -186,27 +186,26 @@ subroutine dump_lines(n,i,ix)
 
   use mod_common_track
 
-  implicit none
```
kyrsjo: Is this simply redundant because it's in the module already?

source/dump.f90
```diff
@@ -186,27 +186,26 @@ subroutine dump_lines(n,i,ix)
...
+  call dump_getTasMatrix(ix, tmpInv, tmpTas, tmpClo)
+
+  if(ldump(0)) then
     ! Dump at all SINGLE ELEMENTs
-    if (ndumpt(0) == 1 .or. mod(n,ndumpt(0)) == 1) then
+    if(ndumpt(0) == 1 .or. mod(n,ndumpt(0)) == 1) then
       if ((n >= dumpfirst(0)) .and. ((n <= dumplast(0)) .or. (dumplast(0) == -1))) then
-        call dump_beam_population(n, i, ix, dumpunit(0), dumpfmt(0), ldumphighprec, &
-          dump_normMat(dump_nMatMap(ix))%orbit, dump_normMat(dump_nMatMap(ix))%invtas)
+        call dump_beam_population(n, i, ix, dumpunit(0), dumpfmt(0), ldumphighprec, tmpClo, tmpInv)
```
kyrsjo: Does tmpClo and tmpInv actually make any sense here, or are you just sending them as placeholders (and somewhere else -- possibly just in the manual -- taking care that a normalized dump for ALL is not attempted?)

source/dump.f90
```diff
@@ -573,7 +571,8 @@ subroutine dump_initialise
       end if
 
       ! Write format-specific headers
-      if (dumpfmt(i) == 1) then
+      call dump_getTasMatrix(i, tmpInv, tmpTas, tmpClo)
```
kyrsjo: What happens here for i=0, i=-1? (maybe add a small comment to remind)

source/dump.f90
vkbo: Yes

kyrsjo:
> Additional note: The tas matrix should probably be saved in a global module instead of under DUMP. especially since it is already shared with FMA. Will probably move this when PR #929 is merged, because there all the linopt stuff is moved to a separate file.

Possibly. However it shouldn't really be shared with FMA, that "feature" was just the first implementation of normalized dumps (before they were moved into dump itself, where they belong more).

source/dump.f90
vkbo: Huh?

I'm using the getTasMatrix routine to get the tmpTas, tmpInv and tmpClo before passing them on. The rotuine checks that they are set first to avoid out of bounds issues. That is the entire point of this fix.

source/dump.f90
vkbo: What happens for 0 and -1 depends on initialisation. Generally -1 is populated, and 0 is not.

vkbo:
> > Additional note: The tas matrix should probably be saved in a global module instead of under DUMP. especially since it is already shared with FMA. Will probably move this when PR #929 is merged, because there all the linopt stuff is moved to a separate file.
> 
> Possibly. However it shouldn't really be shared with FMA, that "feature" was just the first implementation of normalized dumps (before they were moved into dump itself, where they belong more).

Well, they are no longer shared as the data arrays are private. The benefit of extracting the logic is that it isn't really a DUMP feature, just used by DUMP and FMA. There are possibly other places where this may be useful.

PR #936, opened 2019-08-08
Linopt module

vkbo:
This PR moves the linopt routines to a new module, and also the input parsing for it.

This is extracted from my scatter branch as I there inject code to extract twiss parameters from writelin. I want to add some additional features to this module later to save these parameters for specific elements. This is needed both in Scatter and in Collimation.

Note: I needed to edit the include files used for the solenoid, and noticed that the three of them were always included in sequence, so I just merged them to one file.


source/linear_optics.f90
```diff
@@ -0,0 +1,1262 @@
...
+  use mod_common
+  use sixtrack_input
+
+  character(len=*), intent(in)    :: inLine
+  integer,          intent(in)    :: iLine
+  logical,          intent(inout) :: iErr
+
+  character(len=:), allocatable   :: lnSplit(:)
+  character(len=mNameLen) mode
+  integer nSplit,i,ilin0
```
kyrsjo: ilin0 is not used?

source/linear_optics.f90
```diff
@@ -7,6 +7,10 @@ module mod_linopt
 
   implicit none
 
+  ! Linear optics dump file, formerly fort.11
+  character(len=15), private, parameter :: linopt_fileName = "linopt_dump.dat"
```
kyrsjo: Updated manual?

kyrsjo:
Could you add header to the output file as well?

vkbo:
Updates:

There was a region where the indentation was one step out, so fixed that. Also renamed fort.34 to a more sensible file name. It now has a header, but I did not add a header to former fort.11 (coupled parameters file) because I don't really know what everything is.

I also cleaned up the printout a little bit.

PR #938, opened 2019-08-12
Keeping track of particle pairs

vkbo:
This PR is an attempt to solve an issue pointed out by @amereghe 

The current implementation doesn't properly keep track of which particles belong together when going into postprocessing. This PR adds an array preserving the original pair number of each particle, and uses this array (which is reshuffled with all the others on the event of a lost particle) to keep an updated map between a given pair number and their current index in the particle arrays. This map is recomputed after each shuffle, but it's done in a single pass, so it's fast.

The downside to the single pass is that it doesn't check that all entries are unique. I can add a bunch of if-statements in the loop, but if we call lost particles often, that may be slow. Instead the content of the map is checked when the track files are written. If something goes wrong when building the map, some of the indices should be 0, which I check for,

~~The changes causes four of the fast tests to fail due to changes in pair combination, causing minor differences to `fort.10` and `singletrackfile.dat`. We need to double check that this now is correct.~~

This was due to the loop over pairs growing shorter when particles were lost, so with the pair numbers not changing, `writebin` would sometimes miss the last one or two pairs. This has now been fixed, and the tests pass.


vkbo:
The tests failing are:
* lostnotilt
* wire_thck6d
* s316
* wire_thin6d
* lostevery
* eric

Haven't checked the diff on all, but on lostnotilt and lostevery it's a difference in which pair is considered as lost as writebin excludes any pair where one or both particles are lost. These tests have aperture losses, which triggers a reshuffling, breaking up the pairs (the old code assumed each pair was always one index apart).

rdemaria:
I would have expected more test to fail if it was a serious issues. Can someone check if there are fort.10.canonical with unstable particles, which are still passing the tests?

vkbo:
Well, a lot of the tests don't use a singletrackfile or fort.10 for reference, so those will not be affected. 

Also, a lot of tests don't have aperture losses, so the fact that these still pass was how I verified that my changes didn't break existing functionality. In fact, at first most of those tests did fail because I accidentally swapped particle 1 and 2 of each pair when building the pair map.

The tests that do fail now are the ones with aperture losses where a pair is split. The consequence of a pair being split is that the postpr routine will still just pick whatever particle is next in the array as the second particle, meaning the distance calculation will see a jump.

vkbo:
Also: The prob3 and prob1 tests also passed, so the above list is complete.

rdemaria:
Do you think it is an old bug, or was it introduced by sixtrsck 5, when merging the aperture check from collimation? Do prob1 prob3 have lost particles?

vkbo:
> Do you think it is an old bug, or was it introduced by sixtrsck 5, when merging the aperture check from collimation?

Probably best to let @amereghe answer that one. I didn't do much work on SixTrack before that.

I would suspect the fraction of particles this applies to is fairly small. It only happens when the even numbered particle of a pair is lost, but the odd numbered one isn't.

vkbo:
In any case, I've spent most of the day trying to figure out how the old parts of the code keep track of information compared to the new. Since we'd like to have the particle ID be more or less a passthrough parameter, that is, whatever is sent in, comes back out again unchanged, I have rewritten all the parts of the code that uses particle ID as index in arrays. This is now cleaned out.

The pair structure is handled by a two part ID now. It is assumed that the particles after initialisation is stored pair by pair in the arrays, and each pair is assigned an incremental ID as one value, and a 1 or 2 as a second value. This is now written to the particle state files, so it can be verified that these come out as expected. In addition, the number of turns each particle survives in the machine is also recorded in these files.

This means I can now go through the 6 tests that fail after the fix and verify that the results do make sense.

rdemaria:
I still miss the mechanism. I would need to run some diff of the fort.10, Can you update a fort.10.canonocal in the PR?

vkbo:
> I still miss the mechanism. I would need to run some diff of the fort.10, Can you update a fort.10.canonocal in the PR?

I was going to update those tomorrow, so you should see it then.

The way this was done was that writebin had a loop over all particles `j=1,napx-1`, extracted the particle ID from particle `j` such that `ia=partID(j)`, then it would assume the next particle of the pair was `ie=partID(j)+1`, and write that pair to the track file. In most cases in DA studies, that would work just fine because all particles were stored sequentially such that `j == partID(j)` even at the end of tracking, but in the case the second particle of the pair was lost, and had been moved to the end of the array, the ID of partID(j)+1 would be a particle from another pair, mostly such that `partID(j)+1 == partID(j+2)`.

vkbo:
Ok, writing the last response got me to look at the code once again. It's a bit tricky to follow as the different particle arrays are indexed differently, but I realised a couple of things that were unclear before (thanks in part to code comments.that are misleading).

Firstly, the assumption that two particles belonging to the same pair are stored sequentially is definitely dodgy, as there is no guarantee this is preserved through losses, nor can you assume that running the loop to `napx-1` will correctly cover all particles as the highest particle ID isn't necessarily the last one in the array (I've had this segfault before).

*However*, since the writebin routine actually skips writing a pair if just one of them is lost, the assumption *within* that if statement that they are still sequential holds, provided no particles are added in-between. So the writebin logic is otherwise fine, aside from the out of bounds issue above. The distance calculation is run after lost particles, so I think it is fine too, even though it is not shuffled when particles are moved around.

With the extended particle state files, it was also easy to figure out what caused the 6 tests to fail after I changed the logic of writebin. It was just an issue of looping over `napxo` rather than `napx` since the IDs still run to the original size even if particles are lost.

So, the last tests now also pass.

While the writebin results are unchanged, the particle pair reports at the end of tracking, as well as what went into fort.12, was all over the place. I fixed some of it a while back so that the particle coordinates written were at least correct, just not in actual pairs. This is now corrected as well.

vkbo:
It would be good if we could have this PR reviewed and merged as-is. At the current state, this PR doesn't make any user side changes to SixTrack aside from fixing the fort.12 output and the particle pair summary at end of tracking.

Cleaning up the use of particle ID as a lookup index is an important step to move forward with other developments. It also removes an out-of-bounds trap in writebin that may cause issues on some simulations.

vkbo:
Since shuffling arrays is computationally expensive, and some modules still need to track additional particle data, I added a function in mod_particles that computes the original particle index from the pairID. Of course, if it is called too often, the cost/benefit compared to shuffle my be lost.

The function is currently used by scatter to log the statistical weight of multiple scatters.

amereghe:
Sorry for the late reply.
Honestly I don't remember by heart - I should check the codes...
I've started looking at the PR, but have to leave pretty soon - will continue on Monday

amereghe:
Ok - for the record, in v4.5.18 (the latest one that was merged in the FLUKA coupling SVN repo before merging the coupling into the SixTrack github repo), all dumps are correct. In particular:
- the one in `fort.6` loops until `napxo`, i.e. it access particle arrays (tracking AND losses) in their original pair positions;
- the one in `writebin` accesses particle arrays (only the tracking ones, of course) based on napx, assuming that both particles in the pair are either lost or remain stable at the same time - which is basically correct if the second particle is very close to the first one.

rdemaria:
Just a comment: for typical DA studies the fact of particles start close does not imply they are lost on the same turn, there are often large differences of the order of hundred thousand turns sometimes.

amereghe:
then you can conclude on what writebin was duming :)

vkbo:
Up until we started moving particles around in the arrays, everything was correct. It's been a bit mixed since then, but writebin looks to have always been correct. At least aside from the fact that it can potentially go out of bounds in the if-statement checking `pstop`.

I'm merging this now since it doesn't change any of the results in post-processing.

amereghe:
@vkbo ,
I do not agree with your statements. I think that this piece of code from `writebin` in v4.5.18 says it clearly:
```
        do 10 ia=1,napx-1
!GRD
          if(.not.pstop(nlostp(ia)).and..not.pstop(nlostp(ia)+1).and.   &
     &(mod(nlostp(ia),2).ne.0)) then
            ia2=(nlostp(ia)+1)/2
            ie=ia+1
            if(ntwin.ne.2) then
              write(91-ia2,iostat=ierro)                                &
     &numx,nlostp(ia),dam(ia),                                          &
     &xv(1,ia),yv(1,ia),xv(2,ia),yv(2,ia),sigmv(ia),dpsv(ia),e0
              endfile 91-ia2
              backspace 91-ia2
+if cr
              binrecs(ia2)=binrecs(ia2)+1
+ei
            else
              write(91-ia2,iostat=ierro)                                &
     &numx,nlostp(ia),dam(ia),                                          &
     &xv(1,ia),yv(1,ia),xv(2,ia),yv(2,ia),sigmv(ia),dpsv(ia),e0,        &
     &nlostp(ia)+1,dam(ia),                                             &
     &xv(1,ie),yv(1,ie),xv(2,ie),yv(2,ie),sigmv(ie),dpsv(ie),e0
              endfile 91-ia2
              backspace 91-ia2
+if cr
              binrecs(ia2)=binrecs(ia2)+1
+ei
            endif
```

vkbo:
Saying what clearly? The if-statement seems to me to be correct, aside from the risk of assuming `pstop(nlostp(ia)+1)` is a valid index if `nlostp` has been reshuffled.

amereghe:
sorry - yes you are right, the `if` statement is correct. Please ignore my comment.
Concerning accessing `pstop`, I think that in the perspective of DA simulations (where we don't generate new particles), the indexing is fine.
Thanks for merging.

vkbo:
The if-statement is saved by the check for odd/even particle ID in terms of what is written to file, but not in what lookups are made in `pstop`. It is not easy to follow the logic in the code, so at least there is some benefit in making it clearer by using an explicit map of particle pairs.

The possibility that we had erroneous results for pairs that were broken up on different turns had Frederik a bit worried, but it seems that was not an issue in the end.

amereghe:
I do think that the lookup of `pstop` is perfectly fine in a DA perspective even with re-shuffling - unless something goes very badly in memory, but that's bad anyway.

I will anyway have a look at the merged code, to verify that everything is consistent and this aspect can be marked as definitively solved.

vkbo:
Just to wrap it up for the record, as we looked at this in Alessio's office, the danger with the old logic is if you have particle IDs that are either not in increasing order or that are larger than `napxo`, you may reach an out of bounds lookup. Both of these are possible in the current version of SixTrack. In current DA simulations this will not happen, but the writebin routine may also be called in non-DA simulations.

amereghe:
Hi Veronica,
sorry for being late - PR fine with me. My only remark is when, at the end of tracking, in the particle summary in `fort.6`, you check that `ia.eq.0 .or. ie.eq.0` but then you let the code go - and we don't have index `0` in the array used for reporting.
Thanks for the work.

vkbo:
That is exactly why we check for index 0. Because we don't have those in the array, so that would be an error.

vkbo:
Oh, wait, I misread your comment. Yes, there should be a cycle statement in that if-statement. I will add this.

PR #942, opened 2019-08-15
HDF5 module has no interface for int16

vkbo:
Recent changes in DUMP broke HDF5 build. This is a quick fix.


JamesMolson:
should this not be int(nzz,int32) to be future proof?

vkbo:
That would make it *not* future proof. The writeData routine is whatever size integer is, so it must be int(value) not int(vale,int32).

PR #945, opened 2019-08-21
Count particles per turn per element

vkbo:
Currently, the `META` module counts the total number of particle-turns. The particle-turn-element number was computed by multiplying that number by the number of elements. This PR now sums this up explicitly after lostpart.

This is needed for the new BOINC validation file, but the information is also written to the `sim_meta.dat` file as well as used for the average timing per particle, per turn, per element in `sim_time.dat`.


vkbo:
Making it a separate PR as it can go in right away, but will merge the branch into PR #878 directly as well.

Issue #956, opened 2019-08-29
Error function normalisation factor too low precision


vkbo:
The factor 2/sqrt(pi) used in subroutines `errf` and `errff` (FOX version) uses too low precision for bit math. Currently the are hardcoded to 1.12837916709551, but should really use `two/pisqrt`. The subroutine `wwerf` uses a higher precision value and could be replaced with`two/pisqrt` without changing the tests.

Of course, for 128 bit math, all of these are too low precision.

PR #957, opened 2019-08-29
Move initialise_element and wzsub routines out of sixtrack.f90

vkbo:
This PR moves `initialise_elements` to the bottom of `mod_geometry.f90`. Due to inter-dependency with DYNK, it cannot sit inside the module. Alternatively, it would also make sense to put the routine in `mainarrays.f90`.

Also, the `errf`, `wwerf`, `wzsub`, `wzsubv` and `wzset` routines are moved to the utils module. One of them was still using a common block, instead of the module `parbeam` which contains these variables!

Also, the normalisation factor for the `errf` and `errff` subroutines is too low precision. See issue #956. The error function `wwerf` had higher precision, and has now been replaced by the appropriate numerical constants `two/pisqrt`. Making this change in `errf` and `errff` breaks tests, which is why I didn't update this at the moment.


rdemaria:
I agree on the principle, but I cannot look into the code in detail in the next few days. I tentatively approve to unblock if necessary

vkbo:
The moving of the code is trivial, but changing the normalisation constant of the errf function changes results, so that is something I want a second opinion on.

vkbo:
Had a discussion with @amereghe too, and we concluded that merging this cleanly with just code being moved and the potential issue resolved with the common blocks is the best option now.

Fixing the normalisation of the error functions is something we should do in a separate pull request and with a careful check of the changes it produces.

PR #958, opened 2019-08-30
Onesided Collimators

vkbo:
Since both the new jaw fit implementation and the crystals module need additional settings added to the database, I started adding the general parsing for this using onesided collimators as a starting point since this is mostly straight forward.

The PR is now complete and replaces the old feature for select onesided collimators.


vkbo:
The PR is now complete and replaces the old feature for select onesided collimators.

source/coll_db.f90
```diff
@@ -406,7 +444,59 @@ subroutine cdb_readDB_oldFormat
...
+  write(dbNew,"(a)") "# Automatically converted collimator DB from old format file '"//trim(cdb_fileName)//"'"
+  write(dbNew,"(a)") "# Families"
+  do j=1,cdb_nFam
+    write(dbNew,"(a,1x,a16,1x,f13.6)") "NSIG_FAM",cdb_famName(j),cdb_famNSig(j)
+  end do
+  write(dbNew,"(a)") "#"
+  write(dbNew,"(a)") "# Collimators"
+
+  write(dbNew,"(1a,a47,1x,a16,1x,a4,5(1x,a13))") "#",chr_rPad(" name",47),&
+    "opening","mat.","length[m]","angle[deg]","offset[m]","beta_x[m]","beta_y[m]"
```
amereghe: I would write `"opening/fam"`

source/coll_db.f90
```diff
@@ -506,6 +732,29 @@ subroutine cdb_writeDB_HDF5
...
+subroutine cdb_rotateCollimator(collID, rAngle)
+
+  use numerical_constants
+
+  integer,          intent(in) :: collID
+  real(kind=fPrec), intent(in) :: rAngle
+
+  cdb_cRotation(collID) = cdb_cRotation(collID) + rAngle
+  if(cdb_cRotation(collID) > twopi) then
+    cdb_cRotation(collID) = cdb_cRotation(collID) - twopi
```
amereghe: why not:
```
tmpAngle = cdb_cRotation(collID) / twopi
cdb_cRotation(collID) = ( tmpAngle - int(tmpAngle) ) * twopi
```

source/coll_db.f90
```diff
@@ -506,6 +732,29 @@ subroutine cdb_writeDB_HDF5
...
+  else if(cdb_cRotation(collID) < zero) then
+    cdb_cRotation(collID) = cdb_cRotation(collID) + twopi
```
amereghe: why not:
```
tmpAngle = cdb_cRotation(collID) / twopi
cdb_cRotation(collID) = ( tmpAngle + int(tmpAngle) ) * twopi
```

source/coll_db.f90
```diff
@@ -1038,7 +1038,7 @@ subroutine cdb_setLHCOnesided(doOneSide)
 
   do i=1,cdb_nColl
     cdb_cSides(i) = 0
-    if(cdb_cNameUC(i)(1:3) == "TCP" .and. doOneSide .or. cdb_cNameUC(i)(1:4) == "TCDQ" .or. cdb_cNameUC(i)(1:5) == "TCXRP") then
+    if((cdb_cNameUC(i)(1:3) == "TCP" .and. doOneSide) .or. cdb_cNameUC(i)(1:4) == "TCDQ" .or. cdb_cNameUC(i)(1:5) == "TCXRP") then
```
vkbo: What's the point of this?

vkbo:
> Please note that I have made a couple of changes to the doc, and added parentheses in the conditional if-statement concerning LHC single sided collimators, to avoid mis-interpretations by the compiler (or humans).

There is no confusion for the compiler. The precedence of logical operators is well established and unambiguous.

source/coll_db.f90
vkbo: Can do that, although that code will not build.

vkbo:
Your second last commit breaks build. Not sure why the final status shows all checks passing. I will have a look at this on Monday.

source/coll_db.f90
vkbo: This does not work. Should be `tAngle - floor(tAngle)`

PR #960, opened 2019-09-02
Bugfix illegal floating point operation in wwerf

vkbo:
Wrong operator in if-statement causing division by zero.

Or may have been infinity actually ...

### DIFF
--- a/source/mod_utils.f90
+++ b/source/mod_utils.f90
@@ -176,7 +176,7 @@ subroutine wwerf(x, y, wr, wi)
 
   xa = abs(x)
   ya = abs(y)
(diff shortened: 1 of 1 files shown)


amereghe:
not gone through the code in detail, hence I cannot verify what are the consequences. Fully trust @vkbo

vkbo:
It's a typo by me from PR #957. It doesn't change the numerical results, just chooses a simpler method of calculating the error function at the tails, which seem to end up in a region that causes a floating point error when building with stricter compliance in DEBUG.

amereghe:
thanks for the explanation. I think now it is clearer also for others :)

PR #961, opened 2019-09-03
Jaw profiling

amereghe:
This PR aims at adding a couple of collimation tests to check the functionalities concerning jaw fitting, one-sided collimators and random tilting/offsetting/opening. To do so in an optimal way, a dump of the particle positions at the entrance and exit of jaws has been implemented.

Tests are also provided with a plotting script, for a visual inspection of the results.


amereghe:
What's written in the manual concerning one-sided collimators is bullshit:
* the knob in `fort.3` affects only TCP collimators;
* only positive jaws are treated. Negative ones *cannot* be treated by changing the skew angle by pi, as claimed by the documentation;
* the `NAME_SEL` knob in `fort.3` has nothing to do with single sided collimation.

amereghe:
Hi @vkbo ,
the idea is to merge this into master, then update the test in your PR https://github.com/SixTrack/SixTrack/pull/933, and then merge it. What do you think?

source/collimation.f90
```diff
@@ -1834,6 +1837,7 @@ subroutine collimate_start
           "7=x_in(b!)[m], 8=xp_in, 9=y_in, 10=yp_in, 11=x_out [m], 12=xp_out, 13=y_out, 14=yp_out"
         write(coll_scatter_unit,"(a)") "# 1=icoll, 2=iturn, 3=np, 4=nabs (1:Nuclear-Inelastic,2:Nuclear-Elastic,3:pp-Elastic, "//&
           "4:Single-Diffractive,5:Coulomb), 5=dp, 6=dx', 7=dy'"
+        write(profiling_jaw_unit,"(a)") "# 1=icoll, 2=iturn, 3=np, 4=x, 5=xp, 6=y, 7=yp, 8=s, 9=[1:in,2:out,3:back]"
```
vkbo: Can we please have the column headers aligned instead of numbered? We've been updating a lot of files to do this, so I would prefer it if we didn't add new files that needs later updating.

source/collimation.f90
```diff
@@ -1821,11 +1822,13 @@ subroutine collimate_start
       call f_requestUnit('all_absorptions.dat', all_absorptions_unit)
       call f_requestUnit('Coll_Scatter.dat', coll_scatter_unit)
       call f_requestUnit('FirstImpacts.dat', FirstImpacts_unit)
+      call f_requestUnit('JawProfiles.dat', profiling_jaw_unit)
```
vkbo: Please add the filenames as a constant in the module header. I think my updates on this are in the branch not yet merged. Also, we should avoid using mix of upper and lower case characters in new file names as different operating systems have different rules for case sensitivity.

source/collimation.f90
```diff
@@ -2291,6 +2296,8 @@ subroutine collimate_do_collimator(stracki)
   real(kind=fPrec), intent(in) :: stracki
 
   integer j,jjj
+  logical, allocatable :: linside(:) !(npart)
```
vkbo: Why not just allocated it as `logical linside(napx)` right away?

vkbo:
Your new test is now failing.

amereghe:
yes, I know, I am committing/pushing things in steps - I will push the updated canonical in few minutes

PR #967, opened 2019-09-10
Kill non-STF build option

vkbo:
After this has come up again, Riccardo and I agreed that it was time to discuss deleting the non-STF option. It keeps causing issues with the test suite as we have to disable tests that use more than 64 particles but otherwise checks important parts of the code. It also means we no longer need to maintain a duplicate set of track files for all tests that uses these.

This PR removes all of the STF flags from the code and locks it to always use the singletrackfile option. It's been the default for a while now too. Documentation is also updated.


kyrsjo:
Great! Only comment is that it could be useful to have a STF-to-FORT90ish converter, for legacy users.

vkbo:
Maybe it would be better to just have a standalone executable. I doubt anyone relying on this uses SixTrack 5.x anyway. I assume the code can be extracted from pospr or read90?

kyrsjo:
For what it's worth, the converter was written and can be seen in:
https://github.com/vkbo/SixTrack/pull/4

It works, but needs to be integrated with the build and test system. It should probably also accept a flag to write the old-style fort.91-i filenames.

kyrsjo:
The PR to Veronica's branch is now ready to be merged.
One thing that should be changed (and I think e.g. Veronica is better situated to do it) is to use a dedicated `singletrackfile.dat` for testing. I basically picked a random one, however it should ideally be one that (a) has > 32 pairs, and (b) has a few 100 turns.

It should probably also be tested with ntwin=1...

kyrsjo:
Thanks for merging that.

Only thing that one should look at is that the test for the conversion tool is currently referencing the `singletrackfile.dat.canonical` in the test `dump7_fma`. It should really have it's own little file, ideally with > 64 particles and ~100 turns (losses are cool), but I feel that Veronica @vkbo  would be better placed to choose or generate such a test file; given the churn of these files I did not want to add one.

PR #968, opened 2019-09-10
Elens check integration radial profile

amereghe:
This PR aims at fixing an error in the integration of the radial profiles from files. The error was two-folded:
* wrong unit conversion from A/cm2 to A/mm2;
* integration was not deploying the rule of the trapezoid on the radius, but the lower square;

In addition:
* changes in integrating the radial profile:
  * slightly changed parenthesisation;
  * removed offset to values of J given by the first value;
* non-positive values are now loaded. User manual changed accordingly;
* echo of profiles in `fort.6` now comes with column headers and units; moreover, it no longer skips null entries;
* added error message in case profile does not contain any point;
* added a couple of comments in module header;
* updating tests. At the same time, the script plotting the radial profile superimpose also the output from sixtrack in `fort.6`, such that it can be compared to what expected. Moreover, I have restored the `aperture_losses.dat.canonical` in the `_ramp_DYNK` test.


source/elens.f90
```diff
@@ -558,34 +558,37 @@ subroutine parseRadialProfile(ifile)
...
+  elens_radial_profile_R(ii,ifile) = tmpR
+  elens_radial_profile_J(ii,ifile) = tmpJ/c1e2 ! from [A/cm2] to [A/mm2]
 
   goto 10
 
 20 continue
 
   call f_freeUnit(fUnit)
-  write(lout,"(a,i0,a)") "ELENS> ...acquired ",elens_radial_profile_nPoints(ifile),"points."
+  if ( elens_radial_profile_nPoints(ifile).eq.0 ) then
```
vkbo: This is just nitpicky, but could we at least avoid using fortran 77 and older logical operations like `.eq.` in new code?

amereghe: promised.
Could you merge this? I cannot

PR #969, opened 2019-09-10
Jaw Fit - New DB Format

vkbo:
Implements the new format for jaw fit profiles in the new collimations database. The old way of applying jaw fits is now only supported for the old block.

ToDo: Update manual


source/coll_db.f90
```diff
@@ -642,6 +632,82 @@ subroutine cdb_readDBSettings
...
+    if(nSplit /= 5 .and. nSplit /= 7 .and. nSplit /= 9) then
+      write(lerr,"(a,i0)") "COLLDB> ERROR JAW_FIT expects 4, 6 ot 8 values, got ",nSplit-1
+      write(lerr,"(a)")    "COLLDB>       JAW_FIT collname|famname nslices fit1 fit2 [scale1 scale2 [recentre1 recentre2]]"
+      goto 30
+    end if
+
+    rParam(:) = one
+    iParam(:) = -1
+    bParam(:) = .false.
+    if(nSplit > 2) call chr_cast(lnSplit(3), iParam(1), cErr)
```
amereghe: if we check at the beginning that `nSplit >= 5`, we don't need this specific one

source/coll_db.f90
```diff
@@ -642,6 +632,82 @@ subroutine cdb_readDBSettings
...
+    if(nSplit > 5) call chr_cast(lnSplit(6), rParam(4), cErr)
+    if(nSplit > 6) call chr_cast(lnSplit(7), rParam(5), cErr)
+    if(nSplit > 7) call chr_cast(lnSplit(8), rParam(6), cErr)
+
+    call jaw_addJawFit(trim(lnSplit(2)), rParam(1:6), iFit, cErr)
+
+    if(cErr) goto 30
+
+  case("JAW_FIT") ! Apply Jaw Fit Profile
+    if(nSplit /= 5 .and. nSplit /= 7 .and. nSplit /= 9) then
```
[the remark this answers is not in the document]
vkbo: As we discussed, 5, 7 and 9 are the valid options as the additional ones come in pairs.

source/coll_db.f90
vkbo: Still need to check before we read the array, but in principle they could be grouped. However, if we make any changes to the earlier if statement, keeping these this way means we don't have to keep them in sync.

vkbo:
> thanks for the work - fine with me.
> There is only an if statement that is not fully consistent with the code immediately after

It is consistent. Checking that the element in the array exists before accessing it, regardless of the above conditions, means the code is easier to maintain because that check will always work.

Anyway, still need to update manual before merging.

PR #970, opened 2019-09-11
Un-LHC Collimation

vkbo:
Make collimation no longer depend on LHC collimator naming conventions.

This PR does:
* Removes the hardcoded name check to enter collimation from tracking. If the element is in the collimator database, it is treated as a collimator. If not, it's a drift.
* Give a number of files a proper write format and extend precision on integers to allow for larger number of particles. A lot of files only allowed 5 digit integers.
* Remove the database array with an upper case version of the collimator name. All entries are now treated as the element name is written in fort.2 and database.

More to come ...


source/collimation.f90
```diff
@@ -2180,94 +2122,132 @@ subroutine collimate_start
...
-          nampl(ie)    = nampl(ie) + 1
-        else
-          nspx = zero
-          nspy = zero
+      if(firstrun) then
+        if(iturn == 1 .and. j == 1) then
+          sum_ax(ie) = zero
+          sum_ay(ie) = zero
         end if
+      end if
```
amereghe: shouldn't we bring this loop out of the `do j=1,napx` one? still checking `if(firstrun)` and `if (iturn==1)`

source/collimation.f90
```diff
@@ -2180,94 +2122,132 @@ subroutine collimate_start
...
+  use mod_common_main
+  use mod_common_track
+
+  real(kind=fPrec), intent(in) :: stracki
+
+  integer j
+
+  icoll    = cdb_elemMap(myix)
+  nsig     = cdb_cNSig(icoll)
+  c_length = zero
```
amereghe: I think that the actual `collimate_start_collimator` subroutine should finish here.
The following part should be some kind of `collimate_stats`

amereghe:
Hi Veronica,
I am going through the PR. There is one thing that I don't get the reason why you set it up like that.
In `track_thin.f90` you have:
```
...
      case (1)
        stracki = strack(i)
￼￼        ! Check if collimation is enabled, and call the collimation code as necessary
￼        if(do_coll .and. cdb_elemMap(myix) > 0) then
￼          ! Collimator is in database, and we're doing collimation
￼          call collimate_trackThin(stracki,.true.)
...
          if(do_coll) then
            ! Not a collimator, but collimation still need to perform additional calculations
            call `
          end if
...
```

If I correctly understand, `call collimate_trackThin(stracki,.false.)` is used to perform only some statistics on particles being tracked (after the drift has been applied). I would actually create a `collimate_stats` subroutine which perform the stats, call it here explicitly instead of `call collimate_trackThin(stracki,.false.)` and in `call collimate_trackThin(stracki,.true.)`. This would save you from duplicating the stats code, using a flag in the interface and have a counter-intuitive call to `collimate_trackThin(stracki,.false.)` in a piece of code where you would not expect it...
What do you think?

source/collimation.f90
```diff
@@ -2347,29 +2329,18 @@ subroutine collimate_do_collimator(stracki)
   end if
 
 !++  Write beam ellipse at selected collimator
-  if (((cdb_cNameUC(icoll).eq.name_sel(1:mNameLen)) .or. (cdb_cName(icoll).eq.name_sel(1:mNameLen))) .and. do_select) then
-    do j = 1, napx
-      write(coll_ellipseUnit,'(1X,I8,6(1X,E15.7),3(1X,I4,1X,I4))') ipart(j),xv1(j), xv2(j), yv1(j), yv2(j), &
-     &        ejv(j), sigmv(j),iturn,secondary(j)+tertiary(j)+other(j)+scatterhit(j),nabs_type(j)
+  if(chr_toLower(cdb_cName(icoll)) == name_sel .and. do_select) then
```
amereghe: shouldn't we lower `name_sel` instead of `cdb_cName(icoll)`? I think that the latter is already in lower case...

source/collimation.f90
```diff
@@ -2498,28 +2453,10 @@ subroutine collimate_do_collimator(stracki)
     end if
   end if
 
-!++  Assign aperture which we define as the FULL width (factor 2)!!!
-!JUNE2005 AGAIN, SOME SPECIFIC STUFF FOR RHIC
-  if(cdb_cNameUC(icoll)(1:4).eq.'COLM') then
-    c_aperture = two*calc_aperture
-    nom_aperture = two*nom_aperture
```
amereghe: ok here we break the user interface to L-shaped collimators.
We have to think about how to re-implement L-shaped collimators...

source/collimation.f90
```diff
@@ -3257,27 +3187,23 @@ subroutine collimate_end_collimator(stracki)
 
 ! should name_sel(1:11) extended to allow longer names as done for
 ! coll the coll_ellipse.dat file !!!!!!!!
-  if(((cdb_cNameUC(icoll).eq.name_sel(1:mNameLen)).or.&
-      (cdb_cName(icoll).eq.name_sel(1:mNameLen))) .and. iturn.eq.1  ) then
+  if(chr_toLower(cdb_cName(icoll)) == name_sel .and. iturn == 1) then
```
amereghe: same comment as before about what should be lowered in order to perform the comparison

source/collimation.f90
```diff
@@ -3333,39 +3259,41 @@ subroutine collimate_end_collimator(stracki)
...
   use hdf5_tracks2
 #endif
+#ifdef G4COLLIMATION
+  use geant4
+#endif
 
   implicit none
 
-  integer, intent(in) :: j
+  ! integer, intent(in) :: j
```
amereghe: I would completely kill the line

amereghe:
The deletion of the `collimator-temp.db` implies that I have to update the way `plotImpacts.py` scripts get the offset of the collimators. No big deal, just not to forget :)

amereghe:
BTW: I guess that the request to test the build for the fluka coupling and geant4 still holds, right?

vkbo:
> Hi Veronica,
> I am going through the PR. There is one thing that I don't get the reason why you set it up like that.
> In `track_thin.f90` you have:
> 
> ```
> ...
>       case (1)
>         stracki = strack(i)
> ￼￼        ! Check if collimation is enabled, and call the collimation code as necessary
> ￼        if(do_coll .and. cdb_elemMap(myix) > 0) then
> ￼          ! Collimator is in database, and we're doing collimation
> ￼          call collimate_trackThin(stracki,.true.)
> ...
>           if(do_coll) then
>             ! Not a collimator, but collimation still need to perform additional calculations
>             call collimate_trackThin(stracki,.false.)
>           end if
> ...
> ```
> 
> If I correctly understand, `call collimate_trackThin(stracki,.false.)` is used to perform only some statistics on particles being tracked (after the drift has been applied). I would actually create a `collimate_stats` subroutine which perform the stats, call it here explicitly instead of `call collimate_trackThin(stracki,.false.)` and in `call collimate_trackThin(stracki,.true.)`. This would save you from duplicating the stats code, using a flag in the interface and have a counter-intuitive call to `collimate_trackThin(stracki,.false.)` in a piece of code where you would not expect it...
> What do you think?

Hi,

This is just a temporary solution so that I could move the code out of tracking. It is not meant to stay like that. When all of the other steps are complete, I will reorganise the routines in the main collimation module. It was just a minimal effort at this stage.

source/collimation.f90
vkbo: The very last step in the line of steps will be reorganising the main collimation module. I don't want to make any major changes to the order of things until then.

source/collimation.f90
vkbo: This is done so that the `name_sel` feature is not case sensitive. The `name_sel` variable is converted to lower case during input parsing, and here that must be done on the other side as well to ensure the comparison passes in case the `fort.2` / `collDB` has any upper case characters.

I had to do it that way since the old code required the upper case version, which means older input files will be upper case.

source/collimation.f90
vkbo: Yes, we have to implement this in a different way.

source/collimation.f90
vkbo: It must be this way to work with both new and old inputs.

source/collimation.f90
vkbo: Yeah, it's done in a later PR. There's a lot to clean up when we're done.

source/collimation.f90
vkbo: Added a comment explaining the reason for this

source/collimation.f90
vkbo: Fixed

vkbo:
There was only one thing that was applicable, which I have fixed. The restructuring of the subroutines I would prefer to do at the very end. The whole statistics bit I want to extract from the main code as a whole and put under various flags.

PR #976, opened 2019-09-19
Minor Bugfix on Collimation Exit

vkbo:
The exit routine in collimation is called with a different condition than the init routine, causing the final summary files to not be written when running a thin 4D setup.


vkbo:
The init call was fixed in PR #931

PR #977, opened 2019-09-20
Merging of trauthin and trauthck

vkbo:
This PR adds a new module for tracking, and merges `trauthin` and `trauthck`.


vkbo:
Setting up thick tracking the same way as thin tracking works just fine, but some ktrack values are not supported in the thick tracking routines. Should probably look into adding them or having them trigger an error. One difference is the xrot, yrot and srot features which could be implemented in thick as well.

PR #983, opened 2019-10-03
Beam--Beam Ibeco=0 Bugfix

vkbo:
Sometimes, on some systems, and on certain combinations of compiler flags, the wire_thin6d_ibeco0 test would fail with particle coordinates ending up as `NaN`. I looked at everywhere the `ibeco` flag was used (and cleaned up hard to read code in the process) and eventually tracked it down to the beam-beam initialisation code.

In `initialise_elements`, the `beamoff` values were computed from uninitialised variables if the `ibeco` flag was set to 0. There was also a lot of other inconsistencies with if statements for `ktrack` 42 and 43 elements.

In order to understand how the beam beam element kicks are calculated, I combined the numerous beam-beam include files into just three files, one for each of the three `ktrack` values. I also optimised these quite a bit by moving if statements out of loops and using vectorisation where practical.

The files were then compared to the code in `initialise_elements` and the logic reshuffled and rewritten such that no value is used uninitialised. I am a little bit uncertain on how to handle the `ibtyp == 1` case, which is an optimisation step using a vectorised error function, which doesn't really apply during initialisation. So I removed the if condition, but left a comment about this.

Everything still passes tests (changing the beam beam code easily breaks a large number of tests), so things should be fine.

I also track down a place where the beam-beam code causes an underflow (at the end of the include files). It's not really an issue, but it is caused by an exp call in the beam-beam code. It is properly handled by the subnormal implementation, so not an error, but could still probably be avoided.


vkbo:
As discussed with @rdemaria, I implemented the `ibtyp` check again in `initialise_elements` and now call the `wzsub` error function instead when `ibtyp /= 0`.

PR #989, opened 2019-10-09
Add FLUSH statement before stop in splitSingetrack

kyrsjo:
When compiled with Nagfor, the `STOP: <code>` output arrived before the actual error message, which caused the self test of splitSingletrack to fail.

This fixes it.


kyrsjo:
Note: The actual test still depends on a random singletrackfile I found in another test. It should really not.

vkbo:
> Note: The actual test still depends on a random singletrackfile I found in another test. It should really not.

Can't you just copy the file into the test folder where the script lives?

Issue #982, opened 2019-10-02
make `NOFILE` in `DYNK` module the default behaviour


amereghe:
When using `DYNK`, the `dynksets.dat` file is always written unless the user explicitly issue a `NOFILE` keyword in the `DYNK` block. It happened that on BOINC a series of dynked jobs was killed with error `EXIT_DISK_LIMIT_EXCEEDED`, since the `dynksets.dat`was growing too much - with 1e7 turns...

It would be great to change the user interface such that:
- `dynk_noDynkSets` is set to `.true.` by default;
- `dynk_noDynkSets` is set to `.false.` by a specific keyword (e.g. `DYNKSETS`);

vkbo:
I think that's a good idea. In general, large files mainly used for debugging test setup should be off by default.

I can update the parsing for this.

kyrsjo:
Yeah, agreed.

Hopefully whoever used DYNK for 1e7 turns also didn't use FUN FILE ...

vkbo:
Solved in PR #992.

PR #992, opened 2019-10-09
Dynk NOFILE > DYNKSETS

vkbo:
This PR addresses issue #982

The `NOFILE` flag in `DYNK` is removed and replaced by the `DYNKSETS` flag. That is, the new flag need to be present for the `dynksets.dat` file to be written. If it isn't, the file contains a single line stating how to enable this.

Manual was also updated.

All tests using DYNK have been updated, and I cleaned up the `fort.3` files while I was at it.


doc/user_manual/chSpecialElements.tex
```diff
@@ -783,22 +781,22 @@ \subsection{Examples}
...
 \begin{cverbatim}
 DYNK
-NOFILE
-FUN RIPP-dmqx1f50l5+2 COSF_RIPP 3.2315D-10 224.9 0.0
-SET dmqx1f50l5+2 average_ms RIPP-dmqx1f50l5+2 1 -1 0
-FUN RIPP-dmqx2af50l5+2 COSF_RIPP -3.2315D-10 224.9 0.0
-SET dmqx2af50l5+2 average_ms RIPP-dmqx2af50l5+2 1 -1 0
-FUN RIPP-dmqx1f20kl5+2 COSF_RIPP 2.5246D-12 0.56225 0.0
-SET dmqx1f20kl5+2 average_ms RIPP-dmqx1f20kl5+2 1 -1 0
+  DYNKSETS
```
kyrsjo: Here `NOFILE` has been replaced by `DYNKSETS`, inverting the logic. And unless you're debugging FUN:RIPP, you probably don't want a humongous dynksets file for this!

source/dynk.f90
```diff
@@ -183,8 +183,11 @@ subroutine dynk_parseInputLine(inLine,iErr)
     write(lout,"(a)") "DYNK> Debugging is ENABLED"
 
   case("NOFILE")
-    dynk_noDynkSets = .true.
-    write(lout,"(a)") "DYNK> Disabled writing '"//dynk_fileName//"'"
+    write(lout,"(a)") "DYNK> The NOFILE flag no longer has any effect"
```
kyrsjo: This should maybe be an error?

doc/user_manual/chSpecialElements.tex
```diff
@@ -765,8 +763,8 @@ \subsection{Additional Flags}
 
 \subsection{Output File \texttt{dynksets.dat}}
 
-When a \texttt{DYNK} block is present in the input file, a file \texttt{dynksets.dat} is created and in the current working directory.
-Unless a \texttt{NOFILE} statement is present, this file contains first a header
+When the \texttt{DYNKSETS} flag is present in the \texttt{DYNK} block, a file \texttt{dynksets.dat} is created and in the current working directory.
```
kyrsjo: Keep a mention of what the meaning of "NOFILE" was, e.g. a sentence "The DYNKSETS flag was introduced in version xx.xx.xx, replacing the NOFILE flag which had the opposite effect."

source/dynk.f90
vkbo: No, it will not be made an error. The dynkjsets file explains why it's empty.

vkbo:
> Fine with me, thanks @vkbo !
> Only one minor thing: you still keep in the manual the explanation of `NOFILE`, whereas in the code you basically disable it...

Missed that one spot. Fixing it before merging.

source/dynk.f90
vkbo: No, it will not be made an error.

When asking for NOFILE, you get exactly what you ask for since NOFILE is now default behaviour. Error on this is just unnecessarily breaking input files that is asking for exactly what they're getting.

doc/user_manual/chSpecialElements.tex
kyrsjo: Thanks!

PR #995, opened 2019-10-10
adding `#` char to header of aperture losses

amereghe:
The header of the aperture losses file does not have a char marking it is a header.
This may result in analysis tools not being able to correctly crunch the file - e.g. `np.loadtxt`.


kyrsjo:
Is this also the case for some of the collimation files, or has this now been fixed? I have a vague memory of this being an issue in the past...

amereghe:
mmm should check it, and I have to fix the tests, of course...
...running out of time... can we post-pone the BOINC release?

vkbo:
> mmm should check it, and I have to fix the tests, of course...
> ...running out of time... can we post-pone the BOINC release?

I can push new canonicals, if that's all that's missing.

vkbo:
Also: Please don't make any changes to files in collimation right now.

vkbo:
> excellent, thanks a lot!
> Yes, I agree, let's wait to have the Collimation PRs in and then we check the format of the headers

Yes, I've been modifying a lot of them while going through the code. Many have issues with for instance assuming number of particles and turns is <= 9999.

PR #998, opened 2019-10-14
Minor bug in thin tracking for collimator check

vkbo:
Should only look up cdb_elemMap array if do_coll is true

Fails on boundary check with Debug build.


source/track_thin.f90
```diff
@@ -728,7 +728,14 @@ subroutine thin6d(nthinerr)
         if(bdex_enable .and. kz(ix) == 0 .and. bdex_elementAction(ix) /= 0) call bdex_track(i,ix,n)
       end if
 
-      if(do_coll .and. cdb_elemMap(myix) > 0) then
+      is_coll = .false.
+      if(do_coll) then
+        if(cdb_elemMap(myix) > 0) then
+          is_coll = .true.
```
kyrsjo: Add comment:
The array cdb_elemMap is only defined when do_coll is true, so should only check its value in that case.
For some combinations of compilers and flags (gfortran/debug) the logical operation is not short-circuited,
so the if has to be explicitly done in two steps in order to avoid read of unallocated memory.

vkbo: This entire thing will be removed soon. This is just a temporary fix to not break current master.

amereghe: don't you want to set `is_coll=.false.` in the collimation module and to `.true.` after parsing the coll DB?
That would be one stable change...

Issue #962, opened 2019-09-03
optimising time of exact drifts


amereghe:
I think we can optimise a bit the code of exact drifts - from:
```
            xv1(j)=xv1(j)*c1m3
            xv2(j)=xv2(j)*c1m3
            yv1(j)=yv1(j)*c1m3
            yv2(j)=yv2(j)*c1m3
            pz=sqrt(one-(yv1(j)**2+yv2(j)**2))
            xv1(j)=xv1(j)+stracki*(yv1(j)/pz)
            xv2(j)=xv2(j)+stracki*(yv2(j)/pz)
            xv1(j)=xv1(j)*c1e3
            xv2(j)=xv2(j)*c1e3
            yv1(j)=yv1(j)*c1e3
            yv2(j)=yv2(j)*c1e3
```
to:
```
            stracki=c1e3*stracki
            yv1(j)=yv1(j)*c1m3
            yv2(j)=yv2(j)*c1m3
            pz=sqrt(one-(yv1(j)**2+yv2(j)**2))
            xv1(j)=xv1(j)+stracki*(yv1(j)/pz)
            xv2(j)=xv2(j)+stracki*(yv2(j)/pz)
```

Then, there is also the never-ending question about improving the accuracy of an `sqrt` of the difference of squares, i.e. turning:
```
            pz=sqrt(one-(yv1(j)**2+yv2(j)**2))
```
into:
```
            pz=sqrt(one/((one-(yv1(j)+yv2(j)))*(one+(yv1(j)+yv2(j)))))
```
Of course, `pz` should be renamed (e.g. `opz`) and we should compute `yv1(j)*opz` instead of `yv1(j)/pz` (same for the vertical plane). @rdemaria @tpersson , what do you think?

rdemaria:
For the conversion I agree.  For the last line:  1-(a**2+b**2) is not equal to (1-(a+b))(1+(a+b))

amereghe:
sorry you are right on the second point - total bullshit, my mistake, please ignore it

PR #999, opened 2019-10-14
Exact Drift Fix

vkbo:
This PR changes the three places where exact drift is calculated and remove the scaling up and down of all the variables. The scaling up and down is unnecessary and potentially adds numerical errors as well as waste CPU time.

The current change doesn't change the results of the transverse calculation, but changes the last two decimals in the `sigmv` array. It also changes the tune in the last decimals and also the tas matrix.

This means all of the tests using exact drift changes canonicals, but I don't want to update those until we agree this code is correct.


amereghe:
here we go:
https://github.com/SixTrack/SixTrack/issues/962

vkbo:
I think it's a very bad idea to modify `yv1` and `yv2` unnecessarily, as it introduces a numerical error because the multiplications are not round binary numbers. My version of `pz` avoids this. We may consider dividing by `c1e3` at the end instead of multiplying with `c1m3` though, but I did it this way because it didn't alter the `xv1` ad `xv2` coordinates from the old method.

PR #1001, opened 2019-10-15
Cavity/Trombone Update

vkbo:
This PR updates the handling of ktrack type 2 and 3 in tracking; cavity and phase trombone.

The phase trombone implementation was a bit inconsistent, but discussed with @tpersson which one is correct for 4D and 6D, and copied those around.

Cavities loop has been optimised a bit.

Both elements now uses the `part_updatePartEnergy` routine, however, since the cavities element used the other alternative version of computing `dpsv1`, the change broke a lot of tests, and a lot of very large canonicals. It also changed results for very long tracking quite a bit, although only in terms of final coordinates, not in terms of various physics results like in postpr, FMA, etc.

In any case, it was simpler to just swap the calculation of `dpsv1` in the update routine to the one used for the element, and fix the other set of tests that changed, mostly DYNK, crab cavities and beam--beam. The coordiante changes were minimal.

The two different ways of computing `dpsv1` are:
a) `dpsv1(1:napx) = (dpsv(1:napx)*c1e3)/(one + dpsv(1:napx))`
b) `dpsv1(1:napx) = (dpsv(1:napx)*c1e3)*oidpsv(1:napx)`

We're now using the latter.


source/track_thick.f90
```diff
@@ -723,71 +726,61 @@ subroutine thck6d(nthinerr)
...
+            pttemp   = pttemp   + temptr(kxxa)*rrtr(irrtr,6,kxxa)
+          end do
+
+          ! Transforming back to the tracked coordinates of SixTrack
+          ejv(j) = (e0f*pttemp/(c1e3*(e0/e0f))+e0)/mtc(j)
+        end do
+        call part_updatePartEnergy(1,.false.)
+
+        ! We have to go back to angles after we updated the energy.
+        yv1(1:napx) = yv1(1:napx)*mtc(1:napx)/(one+dpsv(1:napx))
```
tpersson: We not just multiply by moipsv(j) ?

vkbo: It had that calculation commented out, but I can change them to this one. I agree it makes more sense, at least code wise. Not sure what the numerical difference is.

PR #1003, opened 2019-10-17
Change aperture tilt to degree 

tpersson:
I did the conversion in the input parsing. I saw that at some other places it is done for each particle. Is there a reason for that?


amereghe:
@tpersson thanks for the modifications!
I don't get your question - could you point me to the concerned lines?

tpersson:
@amereghe It is all fine, my mistake it looks like it is all in the initialization.

rdemaria:
I would make the conversion at initialize elements

amereghe:
@rdemaria it is not needed - the code is already in good shape :D

vkbo:
@rdemaria One check seems to be hanging. Can you use the admin privileges to override and merge? It is ready.

amereghe:
@vkbo which test was failing and where?

tpersson:
Tried to add another commit but still not working. I have no idea what is going wrong. The tests pass but still no report.

vkbo:
Riccardo should be able to override. I do this on my own repository sometimes when I have the same issue, but let me change policy back and forth and do it that way.

PR #1005, opened 2019-10-17
Thin CF does not give the same result as Malte's MAD-X code

tpersson:
There was previously a difference in sign.

Studies if this is the best approach should be done but since it has been a long standing request to have the same as in Malte's version it is now done.


vkbo:
For some reason there is always 2 checks when @tpersson makes a PR, and one always hangs. Not sure why this happens. It block merging ...

tpersson:
This never got merged.. This explains the difference in sign!!

PR #1014, opened 2019-11-08
SplitTrack header fix

vkbo:
Added padding of 44 x zero to the header of the splitSingletrack tool. Now matches code in postpr.


kyrsjo:
Hmm, Table C2 says that there should be 43, not 44, dummies at the end of the file. Seems there may be a problem  in the manual too then.

rdemaria:
If it works with the script of Frank we can merge for me...

vkbo:
I just replicated what postpr does. The number of variables padded also matches the number needed to be commented out in Frank's code when I testet it last night.

If the documentation says otherwise, it probably needs updating.

PR #1023, opened 2019-12-11
Collimation Particle Updates

vkbo:
This PR removes the logic which copies back particles that did not interact with the collimator. The code should really just skip these particles to avoid adding numerical noise.

This PR also changes the energy update on copying back particles to use the routine from mod_particles.

This PR also ensures that the numxv array is properly updated when particles are absorbed in collimation.


source/collimation.f90
```diff
@@ -1797,43 +1798,32 @@ subroutine coll_doCollimator(stracki)
...
+  call part_updatePartEnergy(1,.true.)
 #else
   ! Copy particle data back and do path length stuff; check for absorption
-  ! Add orbit offset back.
+  ! Add orbit offset back
   do j=1,napx
-    ! In order to get rid of numerical errors, just do the treatment for impacting particles
     if(part_hit_pos(j) == ie .and. part_hit_turn(j) == iturn) then
-      ! For zero length element track back half collimator length
+      ! In order to get rid of numerical errors, just do the treatment for impacting particles
```
vkbo: Here I seem to have added back in the wrong comment (I first deleted the one that was there by accident). Will fix.

amereghe:
Hi @vkbo ,
thanks for the changes - I think they are in general fine.

I have anyway a couple of questions:
- final state file of the `collimation_cry_ge_am_hor_b2` test: I guess that the update of the turn number of the lost particles (line 591 and below) is thanks to the update of the `numxv` var. But the last one is lost at turn 0... Moreover, it is a bit suspicious that the loss turn number decreases with the ID of the particle, as if the initial conditions of tracking were sampled along a line instead of randomly...
- in addition to the previous point, the final state file of the `collimation_cry_si_ch_hor_b1` test also shows that the lost particles get delta set to 0.0 and p to p0 (synchronous particle). Do you understand why? The same remarks apply to (more or less all) the other final state files...

vkbo:
> Hi @vkbo ,
> thanks for the changes - I think they are in general fine.
> 
> I have anyway a couple of questions:
> 
> * final state file of the `collimation_cry_ge_am_hor_b2` test: I guess that the update of the turn number of the lost particles (line 591 and below) is thanks to the update of the `numxv` var. But the last one is lost at turn 0... Moreover, it is a bit suspicious that the loss turn number decreases with the ID of the particle, as if the initial conditions of tracking were sampled along a line instead of randomly...

The `numx` variable counts the number of completed turns, not the current turn. When turns column is 0, it means the particle was lost in the first turn.

The reason the loss turn decreases is due to the way the shuffle routine works. The particles lost in a specific turn will always be in increasing order of particle ID, and the turns in decreasing order.

> * in addition to the previous point, the final state file of the `collimation_cry_si_ch_hor_b1` test also shows that the lost particles get delta set to 0.0 and p to p0 (synchronous particle). Do you understand why? The same remarks apply to (more or less all) the other final state files...

This is done in the collimation code when lost particles are set to e0. The update particle routine propagates that to the other coordinates. I think it's a bad idea to alter the coordinates of lost particles in collimation in general.

amereghe:
> The `numx` variable counts the number of completed turns, not the current turn. When turns column is 0, it means the particle was lost in the first turn.

ok - even though I think that reporting the turn at which the particle is lost is more intuitive than the total number of accomplished turns. I would restore it.

> The reason the loss turn decreases is due to the way the shuffle routine works.

Ah yes - in the final state file, lost particles appear ordered by decreasing turn, increasing s-position  (not visible in the file, though) for the same turn, and increasing ID for the same s-position.

> This is done in the collimation code when lost particles are set to e0. The update particle routine propagates that to the other coordinates. I think it's a bad idea to alter the coordinates of lost particles in collimation in general.

Yes, it is a bad idea - but its correction I think goes beyond the scope of this PR. I was surprised to see that particles identified as lost in the master version did not have the reset of p whereas with your changes this is the case - as if your changes corrected a bug...

vkbo:
> > The `numx` variable counts the number of completed turns, not the current turn. When turns column is 0, it means the particle was lost in the first turn.
> 
> ok - even though I think that reporting the turn at which the particle is lost is more intuitive than the total number of accomplished turns. I would restore it.

Restore? The numxv array has always been like that. It was just never populated in collimation on losses, which I fixed here.

> > The reason the loss turn decreases is due to the way the shuffle routine works.
> 
> Ah yes - in the final state file, lost particles appear ordered by decreasing turn, increasing s-position (not visible in the file, though) for the same turn, and increasing ID for the same s-position.

Yes. For collimation this is the case as losses are recorded at end of turn.

> > This is done in the collimation code when lost particles are set to e0. The update particle routine propagates that to the other coordinates. I think it's a bad idea to alter the coordinates of lost particles in collimation in general.
> 
> Yes, it is a bad idea - but its correction I think goes beyond the scope of this PR. I was surprised to see that particles identified as lost in the master version did not have the reset of p whereas with your changes this is the case - as if your changes corrected a bug...

I propose we drop resetting variables other than the transverse coordinates, which is what is checked against elsewhere in the collimation code. Resetting energy just seems unnecessary.

amereghe:
> Restore? The numxv array has always been like that. It was just never populated in collimation on losses, which I fixed here.

Sorry, yes, it is fine - I was thinking about losses as recorded by the online aperture checking or the losses in collimators, whereas here it is the final state file. Anyway, in the other two files the same proton will have a different number of turns than in the final state file - as you said, in the final state file the turn number has a different meaning.

> I propose we drop resetting variables other than the transverse coordinates, which is what is checked against elsewhere in the collimation code. Resetting energy just seems unnecessary.

I would stop resetting at all, but for later PRs.

vkbo:
Thanks for the review.

I can add an issue on the resetting of coordinates for absorbed particles, and also on the combined losses file we discussed in my office.

PR #1026, opened 2019-12-12
Fix root output compiling

JamesMolson:
This had stopped working and I could not work out why.

Anyway change floats -> doubles and it builds.


vkbo:
Yes, all timestamps are now fPrec as well. This was changed with the latest BOINC updates.

PR #1032, opened 2019-12-18
More DA tests

vkbo:
This PR adds even more DA tests from Frank.

We will go through them this afternoon to make sure everything is fine, so not merging yet.


rdemaria:
I think it is useful to add the NORM block as well in all examples, which is an extra step on top of DIFF.

vkbo:
> I think it is useful to add the NORM block as well in all examples, which is an extra step on top of DIFF.

The more we can test, the better. I'll ask Frank for this.

PR #1037, opened 2019-12-18
delete unused variables

JamesMolson:


source/coll_common.f90
```diff
@@ -122,13 +122,12 @@ module coll_common
 
 contains
 
-subroutine coll_expandArrays(npart_new, nblz_new)
```
amereghe: I guess that here `nblz_new` was added to the interface just for the sake of completeness, such that in case one need to add stuff depending on the new number of lattice structures, the interface is already general enough

source/main_da.f90
```diff
@@ -38,13 +38,13 @@ program mainda
...
 
-  integer i,iation,itiono,idate,im,itime,ix,izu,j,k,kpz,kzz,l,ll,ncorruo,ndim,nlino,nlinoo,nmz
+  integer i,iation,itiono,im,ix,izu,j,k,kpz,kzz,l,ll,ncorruo,ndim,nlino,nlinoo,nmz
   real(kind=fPrec) alf0s1,alf0s2,alf0s3,alf0x2,alf0x3,alf0z2,alf0z3,amp00,bet0s1,bet0s2,bet0s3,     &
     bet0x2,bet0x3,bet0z2,bet0z3,clo0,clop0,dp0,dp10,eps,epsa,gam0s1,gam0s2,gam0s3,gam0x1,gam0x2,    &
     gam0x3,gam0z1,gam0z2,gam0z3,phag,qw,qwc,r0,r0a,rv,tas,tas16,tas26,tas36,tas46,tas56,tas61,tas62,&
     tas63,tas64,tas65
 
-  character(len=8)  cdate,ctime ! Note: Keep in sync with maincr. If the len changes, CRCHECK will break.
+!  character(len=8)  cdate,ctime ! Note: Keep in sync with maincr. If the len changes, CRCHECK will break.
```
amereghe: why commenting this line out and not killing it, if it is not needed?
Is there some code that should be injected depending on these vars?

source/coll_common.f90
vkbo: It's probably no longer needed after we killed a bunch of collimation arrays.

PR #1039, opened 2019-12-19
Reposition files routine

vkbo:
This PR just moves the reposition files routine from the scatter module to mod_units, so it can be used by any module. This is especially useful for collimation.

The routine has been cleaned up a bit, made general (also should work outside of C/R) and more comments added.

### DIFF
--- a/source/mod_units.f90
+++ b/source/mod_units.f90
@@ -523,7 +523,7 @@ subroutine f_positionFile(fileName, fileUnit, filePos, newPos)
   inquire(unit=fileUnit,opened=isOpen)
   if(isOpen) then
 #ifdef CR
-    write(crlog,"(a,i0,a)") "UNITS> ERROR Failed while repsositioning '"//trim(fileName)//"', "//&
+    write(crlog,"(a,i0,a)") "UNITS> ERROR Failed while repositioning '"//trim(fileName)//"', "//&
       "unit ",fileUnit," already in use"
     flush(crlog)
 #endif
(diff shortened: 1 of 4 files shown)


source/mod_units.f90
```diff
@@ -483,6 +483,100 @@ subroutine f_flush(unit)
...
+  logical isOpen, fErr
+  integer iError
+  integer j
+  character(len=mInputLn) aRecord
+
+  call f_requestUnit(fileName,fileUnit)
+  inquire(unit=fileUnit,opened=isOpen)
+  if(isOpen) then
+#ifdef CR
+    write(crlog,"(a,i0,a)") "UNITS> ERROR Failed while repsositioning '"//trim(fileName)//"', "//&
```
amereghe: repositioning

source/mod_units.f90
```diff
@@ -483,6 +483,100 @@ subroutine f_flush(unit)
...
+  integer,          intent(inout) :: fileUnit
+  integer,          intent(inout) :: filePos
+  integer,          intent(in)    :: newPos
+
+  logical isOpen, fErr
+  integer iError
+  integer j
+  character(len=mInputLn) aRecord
+
+  call f_requestUnit(fileName,fileUnit)
```
amereghe: shouldn't this call be under if statement? `f_requestUnit` initialises fileUnit to -1

source/mod_units.f90
```diff
@@ -483,6 +483,100 @@ subroutine f_flush(unit)
...
+    if(fErr) goto 10
+    filePos = 0
+    do j=1,newPos
+      read(fileUnit,"(a)",end=10,err=10,iostat=iError) aRecord
+      filePos = filePos + 1
+    end do
+    endfile(fileUnit,iostat=iError)
+    call f_close(fileUnit)
+
+    call f_open(unit=fileUnit,file=fileName,formatted=.true.,mode="w+",err=fErr,status="old")
```
amereghe: Maybe add a comment explaining why we need to end the file, close it and re-open it again

source/mod_units.f90
vkbo: What should the if-statement check?

amereghe: clarified by discussion - fine

vkbo:
Noticed that the handling of `iostat` in `f_open` could return an uninitialised value, so fixed that as well.

PR #1041, opened 2020-01-06
enabling norm for 5D (from Frank)

rdemaria:
Enabling normal form calculation on 5D (from F. Schmidt).

(Testing in progress...)


JamesMolson:
Frank says there are some more bits missing so don't merge this just yet

source/sixda.f90
```diff
@@ -116,21 +118,21 @@ subroutine daliesix
...
-  call dadal(rot,nd2)
-  call dadal(xy,nd2)
+  call dadal(rot,nv)
+  call dadal(xy,nv)
   call dadal(h,1)
   call dadal(h4,1)
   call dadal(hc,1)
   call dadal(hs,1)
-  call dadal(df,nd2)
+  call dadal(df,nv)
```
JamesMolson: apparently all this (the nv) should not be there, the rest is fine

PR #1043, opened 2020-01-16
Fix bug in pencil beam

amereghe:
When using the pencil beam of type 3, the optics functions at the entrance face of the collimator were always used. On the contrary, if the actual collimator cut is provided by the downstream edge of the jaw, then a mis-matched beam is generated, which will never interact with the collimator. Strange behavior noticed by A. Fomin - thanks for spotting it!

In addition to the bug fix, I have added also a suitable test, based on the input provided by A. Fomin.

The bug came in with commit https://github.com/SixTrack/SixTrack/commit/952d4a549b5f0eac76b18de4edea35e5061ba7e6 , in the context of PR https://github.com/SixTrack/SixTrack/pull/987
I missed it when reviewing the code.


source/coll_common.f90
```diff
@@ -63,6 +63,7 @@ module coll_common
   character(len=11), parameter :: coll_tracksFile     = "tracks2.dat"
   character(len=17), parameter :: coll_positionsFile  = "CollPositions.dat"
   character(len=20), parameter :: coll_pencilFile     = "pencilbeam_distr.dat"
+  character(len=23), parameter :: coll_pencilFileRB   = "pencilbeam_distr_RB.dat"
```
vkbo: Why "RB"?

JamesMolson: I assume this is because roderik intially wrote this bit of code

vkbo: Yeah, not a good naming convention. I suggested renaming it to `pencilbeam_distr_3.dat` instead.

amereghe: ok done

PR #841, opened 2019-04-18
more elens changes

amereghe:
This PR is a continuation of PR https://github.com/SixTrack/SixTrack/pull/839 - separated, though, not to mix modifications necessary by fox with other stuff.

Changes:
   * write statements involving exponential notation are all converted to `1pe22.15`;
   * remaining signs of chebyshev maps in `elens` module have been removed;
   * gamma_electron added to computation of kick by elens. Documentation and tests updated accordingly;
   * fixing plotting scripts and plots of radial integrated currents and kicks of elenses with measured radial profiles; fix in SixTrack code will actually come with PR https://github.com/SixTrack/SixTrack/pull/839, where it was found and fixed immediately in order not to forget to do it;
   * `elens` module fully dynamic allocatable;
   * give possibility to express `R_1` and `R_2` in sigma;
   * add any ion species to be defined as possible lens beam;
   * perform checks on sensible values of `R_1` and `R_2`, eg they fall inside the domain of the ASCII map;
   * review plots of tests and optimise storage space in git;

A change in the printout format of `dynkset.dat` was necessary - hence, all the tests with this file among the extra checks have been updated.

PR #1044, opened 2020-01-21
[WIP] wrapping up with current development of electron lenses and Chebyshev polynomials

amereghe:
This PR is aimed at inserting in master the main changes that the following PRs would have brought:
* PR https://github.com/SixTrack/SixTrack/pull/839
* PR https://github.com/SixTrack/SixTrack/pull/841
* PR https://github.com/SixTrack/SixTrack/pull/849

The development behind those 2 PRs has been stopped, since the eLens and Chebyshev modules are being re-visited for having a fully 6D symplectic model, and hence there was no need to continue the foxification of those two modules as they are. Nevertheless some side modifications are anyway useful.

Hence, I am importing the code anyway via this PR, but always without performing the FOX calculations on the two modules.


amereghe:
I have decided to go ahead with the original three PRs

PR #1045, opened 2020-01-21
Update lxplus detection and compiler + cmake paths with the lxplus flag.

JamesMolson:
This should hopefully fix an issue @amereghe was having.

Also the paths used for the lxplus flag were still pointing to sl6 versions.

Everything is updated to the latest now.


rdemaria:
I understand it is better to check for lxplus explicitly instead of AFS/CVMFS because it forces the version of the compilers, right?

JamesMolson:
It doesn't force the compiler version, but only tells you how to do it.

The previous method was freezing when it tried to access cvmfs on alessio's computer - /cvmfs/ existed and was mounted, but wasn't connected resulting in a potentially infinite time out.

rdemaria:
ok for me

amereghe:
I cannot merge :)

amereghe:
just checked out the source branch, and it works perfectly on my machine.
Thanks for the fix, @JamesMolson

PR #1049, opened 2020-01-28
Noskip

tpugnat:
Two bug are fixed:
  - The flag FFSKIP was not taken into account because of change of the operator in track_thin
  - For small x or y, particle are lost quickly because of abs(log()) in mod_ffield_aux

### DIFF
--- a/source/mod_ffield.f90
+++ b/source/mod_ffield.f90
@@ -365,7 +365,7 @@ contains
   subroutine ffield_mod_link(iErr)
     ! Mod from SixTrack
     ! ---------------------------------------------------------------------------------------------- !
-    use mod_common,          only : e0f
+    use mod_common,          only : e0f, qq0
     use physical_constants,  only : clight
     use numerical_constants, only : c1e6
 
@@ -385,7 +385,7 @@ contains
       ffNLn = ffNLn-1
       ffMSn = ffMSn-1
       ffNLFile = ffNLFile-1
-      norm=clight/(e0f*c1e6)   ! [c/eV] = 1/p0
+      norm=(qq0*clight)/(e0f*c1e6)   ! [c/eV] = Q0/p0
 
       ! Generate the array of type(ffTable_n_Track)
       ! -------------------------------------------------------------------------------------------- !
@@ -594,7 +594,7 @@ contains
     use parpro,              only : nblo
     use mod_common,          only : napx, ic, tiltc, tilts
     use mod_common_track,    only : strack
-    use mod_common_main,     only : xv1, xv2, yv1, yv2, oidpsv, dpsv, zsiv, xsiv, llostp !, rvv, ejv
+    use mod_common_main,     only : xv1, xv2, yv1, yv2, oidpsv, dpsv, zsiv, xsiv, llostp, mtc !, rvv, ejv
     use numerical_constants, only : half, one, c1e3, c1m3, c1m6
 
     implicit none
--- a/source/mod_ffield.f90
+++ b/source/mod_ffield.f90
@@ -586,11 +586,12 @@ contains
   ! ================================================================================================ !
   !  Lie 2 Tracking if particles enter the Quadupole
   !  B. Dalena, T. Pugnat and A. Simona from CEA
-  !  Last modified: 2020-01-31
+  !  Last modified: 2020-03-26
   ! ================================================================================================ !
-  subroutine ffield_enterQuad(ffi)
+  subroutine ffield_enterQuad(ffn,ffi,ffix)
     ! Mod from SixTrack
     ! ---------------------------------------------------------------------------------------------- !
+    use aperture,            only : apflag,lbacktracking,aperture_reportloss,aperture_savelastcoordinates
     use parpro,              only : nblo
     use mod_common,          only : napx, ic, tiltc, tilts
     use mod_common_track,    only : strack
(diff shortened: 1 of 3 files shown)


source/mod_ffield_aux.f90
```diff
@@ -820,13 +820,13 @@ subroutine TLie2(this,x,px,y,py,zb,deltap1)
 
     ! Check size vectors xpow and ypow   (Prevent SIGFPE)
     ! ---------------------------------------------------------------------------------------------- !
-    log_tmp=abs(log10_mb(abs(x)))
+    log_tmp=log10_mb(abs(x))
     if (log_tmp*(this%n)>230) then
```
amereghe: I don't know the logics behind (hence maybe my question is very naive), but since you have removed the `abs` on `log_tmp`, are you sure that the if statement still makes sense as is?

tpugnat: This part was to prevent the signal SIGFPE. n and m are the maximum exponent for respectively x and y in the horner routine and can have the value 30. This prevent an overflow but adding the abs outside the log generate an error for on axis (x or y) particles.
If the SIGFPE signal is not taken into account for the release anymore then this section can be removed.

tpugnat: As a side not, ffindex(:) is equal to 0 most of the time, positive if the code use the fringe field procedure and negative if we want to skip a multipole.

amereghe: Hi @tpugnat ,

I think that you still get a `SIGFPE` when `x=0` or `y=0` (on-axis particles) - from `log10_mb(zero)`.

Moreover, the left-hand part of the check under `if` statement was always positive before this bug fix, whereas now it can be either positive or negative, and it is not clear to me that comparing the left-hand side (positive or negative) with a number (always) positive is what you want to do...

I don't have any other objection - in the end, you are the main developer and user of this part

vkbo: Since subnormal numbers are handled anyway, can you rewrite this to drop the log10 call in the if-statement and just check against a large number? The log function is relatively slow.

tpugnat: I will make the change.

tpugnat:
The correction has been made. I check on my side and the code work normally.

vkbo:
Great!

Before we merge, can you also delete the commented out code? The history of the code is preserved in the repository so there is no need to keep it. if you need to restore it at some point, you can always go to an earlier version.

tpugnat:
Commented part and useless variable removed. Some comment code are still present but I might need for further update.

source/mod_ffield.f90
```diff
@@ -394,9 +386,8 @@ subroutine ffield_mod_link(iErr)
       ffNLn = ffNLn-1
       ffMSn = ffMSn-1
       ffNLFile = ffNLFile-1
-      gamma0r=pmap/e0 !      gamma0=e0/pmap
-      beta0=sqrt(one-gamma0r*gamma0r) ! = sqrt(one-one/(gamma0*gamma0))
-!      p0=beta0*e0*c1e6/clight
+      gamma0r=pmap/e0 
+      beta0=sqrt(one-gamma0r*gamma0r)
```
vkbo: Hmm, just noticed this.

First of all, do not use `pmap` to compute gamma. `pmap` is NOT the mass of the reference particle. The variable `nucm0` is. `pmap` is a constant that may or may not be used to set `nucm0`, depending on the input file settings.

Secondly, both the variables for gamma0, beta0 and gammar already exist in common modules.

source/mod_ffield.f90
```diff
@@ -779,13 +770,13 @@ subroutine ffield_exitQuad(ffi)
     ! ---------------------------------------------------------------------------------------------- !
     integer          :: iFile
     integer          :: ffj,  k                    ! iterator
-    integer          :: itDlt                      !
+    integer          :: itDlt                      !
```
vkbo: Please strip trailing spaces from code lines.

tpugnat: Ok

source/mod_ffield.f90
tpugnat: Ok I will check them and make modification

vkbo: They are all in module `mod_common`, so should be straight forward to use.

source/mod_ffield.f90
tpugnat: Done

source/mod_ffield.f90
tpugnat: Done

rdemaria:
you should probably use (but please check)

norm=mtc(j) * clight/(e0f*c1e6) 

where mtc is the mass to charge ration with respect to the reference particle. Now Sixtrack can track muliple particles types at the same time.

amereghe:
@tpugnat could you, please, comment on @rdemaria 's suggestion? I think it is an important point to clarify, even if it is not strictly speaking a show-stopper for this PR, but it would be a pity if the module could be used only for protons.
Otherwise, since there are three approvals, I can merge this PR straight away.

tpugnat:
Sorry for my late answer. To summarise my offline discussion with @rdemaria:
  - I can't use mtc(j) in this place because the subroutine in question is not for the tracking but to prepare the data for the tracking. I have to check the impact of tracking multiple particles types at the same time. For now, I will make a check to return an error when using multiple particles types and the FFIELD routine. 
  - I forgot to put the charge in the equation and I will make the change before the end of the week.

amereghe:
Hello @tpugnat ,
no worries, I just wanted to be sure that I was pressing 'merge' at the right moment, which does not seem to be the case right now.
No worries, take your time to implement what's missing. We will wait for you to ping us before merging.

amereghe:
Hi @tpugnat ,
why have you closed the PR? please note that the PR has NOT been merged yet...
Let me know if you really want to disregard all these changes or if you want them in.
Cheers,
A.

PR #1050, opened 2020-01-30
Bug fix: update pairMap only in case of primary beam particles

amereghe:
In case secondary particles are injected in the tracking arrays, `pairID` and `pairMap` arrays are extended, with an initialisation to 0. This leads the `updatePairMap` subroutine to crash, as `pairMap(0,0)` would be updated for all secondary particles, out of bounds.

The fix is simple, i.e. to update the `pairMap` only for what concerns primary particles. The `pairMap` is anyway regularly expanded to keep the reshuffling algorithm smooth, without may exceptions.

Thanks to @JamesMolson for spotting the problem, in the context of the Fluka-SixTrack coupling.

### DIFF
--- a/source/mainarrays.f90
+++ b/source/mainarrays.f90
@@ -292,7 +292,7 @@ end subroutine shuffleLostParticles
 ! ~~~~~~~~~~~~~~~~~~~~~~~~~
 !  V.K. Berglyd Olsen, BE-ABP-HSS
 !  Created: 2019-08-12
-!  Updated: 2019-08-12
+!  Updated: 2020-01-30, A. Mereghetti (BE-ABP-HSS)
 !
 !      This map allows for reverse lookup from a pairID to the index of its two particles in the
 !  main particle arrays. This is used for the distance calculation and for post-processing.
@@ -301,6 +301,10 @@ end subroutine shuffleLostParticles
 !      If, for some reason, each pair ID is not represented exactly twice in the array, the final
 !  map will contain zeros. Any routine using this map for lookup must therefore check for 0 values
 !  and trigger necessary error handling.
+!      In case of new particles added to the tracking arrays (e.g. secondary particles out of
+!  inelastic interactions), these must not enter the logics of couples, since the original particle
+!  died. Hence, keep the pairMap table of the length of the tracked particles (to keep particle
+!  reshuffling simple), but do not update the map for non-primary particles.
 ! ================================================================================================ !
 subroutine updatePairMap
 
@@ -311,6 +315,8 @@ subroutine updatePairMap
 
   pairMap(:,:) = 0
   do j=1,npart
+    ! do not update the map in case the particle is not a primary one
+    if (pairID(1,j).eq.0.and.pairID(2,j).eq.0 ) cycle
     pairMap(pairID(2,j),pairID(1,j)) = j
   end do
 
--- a/source/mainarrays.f90
+++ b/source/mainarrays.f90
@@ -316,7 +316,7 @@ subroutine updatePairMap
   pairMap(:,:) = 0
   do j=1,npart
     ! do not update the map in case the particle is not a primary one
-    if (pairID(1,j).eq.0.and.pairID(2,j).eq.0 ) cycle
+    if (pairID(1,j)==0.and.pairID(2,j)==0) cycle
     pairMap(pairID(2,j),pairID(1,j)) = j
   end do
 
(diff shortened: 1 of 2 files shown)


vkbo:
Yeah, I added the check on 0 to catch such issues. Thanks for thje fix.

source/mainarrays.f90
```diff
@@ -311,6 +315,8 @@ subroutine updatePairMap
 
   pairMap(:,:) = 0
   do j=1,npart
+    ! do not update the map in case the particle is not a primary one
+    if (pairID(1,j).eq.0.and.pairID(2,j).eq.0 ) cycle
```
vkbo: I'd suggest using modern Fortran operators `==` instead of the F77 and earlier format `.eq.`. I've been trying to clean up the old format when I modify code, since we no longer try to preserve F77 compatibility.

There should also not be a space before the closing bracket.

vkbo:
Note that this is not a complete fix.

If you want to enable postpr for FLUKA integration jobs, 0 entries will blow up there. That's where I added the checks.

vkbo:
Just a note on code style:

I never found a proper code style for Fortran, but I've generally followed the Python one where it can be applied. It would probably be useful to write one up based on PEP 8 and remove the bits that don't apply. I've generally tried to stay withing 120 character lines though, as opposed to PEP 8 recommendation of 79, or alternatively, 99.

amereghe:
> Note that this is not a complete fix.
> 
> If you want to enable postpr for FLUKA integration jobs, 0 entries will blow up there. That's where I added the checks.

Good point - I have removed the use of `napxo` in the FLUKA coupling, as the concerned code actually doing things with that is basically gone. The only remaining part was the echo of the surviving particles in `fort.6`, which can cycle up to `napx` with no need of using `napxo`. Hence, the `#ifdef FLUKA` in `source/include/lostpart.f90` (the only place where the `FLUKA` compilation flag was injecting an update of `napxo`) can disappear. I will commit these changes after the long tests will be over on my local machine.

Just to check -- If I correctly understand, the `part_writeState` will dump also secondary particles, which will have `isPrim=.false.` and `0.0` in the `pairID` column, right?

> I never found a proper code style for Fortran, but I've generally followed the Python one where it can be applied. It would probably be useful to write one up based on PEP 8 and remove the bits that don't apply. I've generally tried to stay withing 120 character lines though, as opposed to PEP 8 recommendation of 79, or alternatively, 99.

Yes, I agree with you. In particular, the recommendation of 79 chars on a single line is practically useless - it helps in having many windows or sub-windows of text editors side by side, but it messes up reading complex lines

vkbo:
> Just to check -- If I correctly understand, the `part_writeState` will dump also secondary particles, which will have `isPrim=.false.` and `0.0` in the `pairID` column, right?

`isPrim` is essentially a true/false check on whether parentID and partID is the same, so as long as parentID is properly set, then yes.

parID depends on what you implement. The coupling code should probably assign a pairID I suppose? I don't know if that makes sense or not.

> > I never found a proper code style for Fortran, but I've generally followed the Python one where it can be applied. It would probably be useful to write one up based on PEP 8 and remove the bits that don't apply. I've generally tried to stay withing 120 character lines though, as opposed to PEP 8 recommendation of 79, or alternatively, 99.
> 
> Yes, I agree with you. In particular, the recommendation of 79 chars on a single line is practically useless - it helps in having many windows or sub-windows of text editors side by side, but it messes up reading complex lines

Well, the PEP8 standard only requires 79 characters for docstrings in function headers, and is flexible on code lines with a maximum of 99. For Python, that actually works quite well, and I tend to follow these rules. For Fortran, I generally had 100 characters as an ideal length, and 120 as a maximum. But I do think 120 should be the limit for PR readability. Text is generally more readable if shorter.

amereghe:
As written in my comment, I still had a couple of changes to push, but I was waiting to compile once the code, and I had to wait the long tests on my machine to be over...

amereghe:
> parID depends on what you implement. The coupling code should probably assign a pairID I suppose? I don't know if that makes sense or not.

No, I think that it makes no sense to try to assign secondary particles to couples - the concept of couples belongs to DA studies, where beam particles are of interest, not secondary particles out of inelastic interactions.

JamesMolson:
Oops, my fault for poking riccardo about it

rdemaria:
My apologies, I thought it was completed! Please just reopen a new PR...

amereghe:
> My apologies, I thought it was completed! Please just reopen a new PR...

sure, no worries :)

vkbo:
> > parID depends on what you implement. The coupling code should probably assign a pairID I suppose? I don't know if that makes sense or not.
> 
> No, I think that it makes no sense to try to assign secondary particles to couples - the concept of couples belongs to DA studies, where beam particles are of interest, not secondary particles out of inelastic interactions.

My point was to assign them anyway to prevent out of bounds errors and garbage values from memory in the analysis. I added checks to catch this to avoid exactly this. Running PostPR on secondaries will therefore hit my bug trap, as intended. A more sensible solution may be to just filter out secondaries of course and skip them in the loops. Either way, one of these approaches need to be implemented.

PR #1052, opened 2020-02-10
Fix pair map (continued)

amereghe:
This is a continuation of PR #1050.

Here we remove any update of `napxo` that was still done in the context of the Fluka-SixTrack coupling, presently useless as the concerned loop can be run on `napx`.

Continuing on the comment https://github.com/SixTrack/SixTrack/pull/1050#issuecomment-580291211 by @vkbo :
> My point was to assign them anyway to prevent out of bounds errors and garbage values from memory in the analysis.

I agree with you on this, that's good, and it stays like that with these PRs.

> I added checks to catch this to avoid exactly this. Running PostPR on secondaries will therefore hit my bug trap, as intended.

I agree with you also on this, that's good, and it stays like that with these PRs. The loop in post-processing goes with `napxo`, hence having removed its (useless) update has put the Fluka-SixTrack coupling in the position of running DA studies (if ever wanted...). Therefore, your check will actually trap no secondary particles, but only problems in the mapping :)

> A more sensible solution may be to just filter out secondaries of course and skip them in the loops.

Not sure this is a smart idea - we should have two integers for accessing the tracking arrays...

> Either way, one of these approaches need to be implemented.

I think we are there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 * Do not update the pair mapping for non-primary particles. PR #1050 (A. Mereghetti)
 * Removed updating napxo variable in the context of the Fluka-SixTrack coupling. This allows not to screw-up pair mapping in the context of DA studies. PR # 1052 (A. Mereghetti)
 
+**User Side Changes**
+
+* When specifying `XP` and `YP` in the `FORMAT` statement of the `DIST` block, the units are parsed. Accepted values are [1], [1000], [MRAD], [RAD].
+
 ### Version 5.4.3 [19.12.2019] - Release
 
 **Bugfixes**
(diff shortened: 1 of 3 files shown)


vkbo:
Not familiar enough with the coupling to advice on this. All I can say is what my approach was when I added the pair mapping. :)

amereghe:
> Not familiar enough with the coupling to advice on this. All I can say is what my approach was when I added the pair mapping. :)

Sure, I just wanted to check with you that:
> Either way, one of these approaches need to be implemented.

We have an implementation :) If you agree, I will merge straight away

amereghe:
This PR also includes the forgotten update in the `CHANGELOG.md` mentioned in PR #1054

PR #1053, opened 2020-02-10
Changes to CollDB format conversion subroutine and bug fixes

MarcoDAndrea:
The `cdb_writeDB_newFromOld` subroutine always converts an old format CollDB file to the new format using the collimator family settings specified in the `fort.3` file, even when actual collimator settings for the simulation are required to be taken directly from the CollDB instead. This pool request implements some changes to the subroutine to address this, namely:
- The separate subroutine `cdb_reconstructNSigFam` is called if collimator settings are required to match those of an old format CollDB. Family settings are reconstructed based on the most frequent collimator setting for each family;
- The separate subroutine `cdb_nullEmptyFam` checks for family with no collimator and sets the corresponding aperture to zero.

Additionally, this pool request includes some minor fixes:
- The maximum particle ID that can be printed in the FirstImpact file has been increased;
- The number of digits of collimator lengths in coll_summary file has been increased to properly display crystal collimators, which are usually a few mm long;
- First impacts on crystal collimator are now correctly flagged inside the crystal collimation module, by setting `isImp` to true every time a hit on the crystal collimator is recorded;
- A missing check on the `dowrite_impact` flag when writing the coll_scatter file has been added.


source/coll_db.f90
```diff
@@ -586,6 +586,16 @@ subroutine cdb_writeDB_newFromOld
     return
   end if
 
+  do j=1,cdb_nColl
+    ! Fixing family setting to that of the first collimator in the family when writing old CollDB in new format
+    ! Allows to take into account if settings were defined in fort.3 or old CollDB
+    if(j > 1 .and. cdb_cFamily(j) == cdb_cFamily(j-1)) then
+      continue
```
vkbo: Is `continue` a placeholder for coming code, or do you want it to continue the loop like in Python? If it's the latter, then it will not work as you intended, although since there is no more code in the do loop (at the moment) it will anyway. However, this can be a stumbling block in the future if code is added after the if-statement.

`continue` in Fortran is the same as `pass` in Python. `cycle` is the Fortran equivalent of Python's `continue`.

MarcoDAndrea: Thanks for the remark, I fixed it in the latest commit.

source/coll_crystal.f90
```diff
@@ -244,6 +244,8 @@ subroutine cry_doCrystal(ie,iturn,j,mat,x,xp,z,zp,s,p,x0,xp0,zlm,s_imp,nhit,nabs
   real(kind=fPrec) xp_temp,xp_shift,xp_rot,xp_int,xp_tangent
   real(kind=fPrec) tilt_int,shift,delta,a_eq,b_eq,c_eq
 
+  logical isImp
```
amereghe: Could you maybe add here the `intent` statement for this var? It helps clarifying what's the role of the variable...
You can find examples in other modules.

source/coll_db.f90
```diff
@@ -586,6 +586,16 @@ subroutine cdb_writeDB_newFromOld
     return
   end if
 
+  do j=1,cdb_nColl
+    ! Fixing family setting to that of the first collimator in the family when writing old CollDB in new format
+    ! Allows to take into account if settings were defined in fort.3 or old CollDB
+    if(j > 1 .and. cdb_cFamily(j) == cdb_cFamily(j-1)) then
```
amereghe: Here you assume that collimators in the same family are all contiguous in memory. I guess that this is not the case e.g. for IR TCTs... Then you would store in memory the settings of the first collimator in the last batch of consecutive collimators in the same family

amereghe:
Thanks, @MarcoDAndrea , for the work!

One thing that we have been discussing lately is to change the logic of inferring the settings of a family based on the settings of the collimators in that family - presently those of the first collimator found in the family are taken. This can be mis-leading in the case eg all TCPs are at 5.0 sig and the first one in the family has exceptionally the settings at 6.0 sig.

Probably, it would be better to implement some kind of settings statistics inside each family, and set as family setting those recurring most frequently. @vkbo what do you think? I think it should not be too complicated - we may need a couple of subroutines, one doing the analysis on a given family, and the other one being the wrapper cycling over all families.

source/coll_crystal.f90
vkbo: Yes, all function variables should have the intent statement.

MarcoDAndrea: Thanks for the remark, the `intent` statement has been added in one of the latest commits.

source/coll_db.f90
MarcoDAndrea: The logic of the family settings reconstruction has been reworked in a separate subroutine. As you suggested in your comments, the reconstruction is now based on the most frequent setting for each family.

source/coll_db.f90
```diff
@@ -632,6 +637,72 @@ subroutine cdb_writeDB_newFromOld
...
+    end do
+
+    freq_max = 0
+    do k=1,count_sig
+      freq = count(NSig_collFam == NSig_collFam(k))
+      if(freq > freq_max) then
+        freq_max = freq
+        cdb_famNSig(j) = NSig_collFam(k)
+      end if
+    end do
```
amereghe: probably this part should be under:
`if(count_sig/=0)then`

source/coll_db.f90
```diff
@@ -632,6 +637,72 @@ subroutine cdb_writeDB_newFromOld
...
+
+    if(count_sig /= 0) then
+      call dealloc(NSig_collFam,"NSig_collFam")
+    else
+      cdb_famNSig(j) = 0
```
amereghe: probably we can skip this, as the `cdb_nullEmptyFam` subroutine is called afterwards

PR #1054, opened 2020-02-12
parsing unit of `xp` and `yp` in `FORMAT` keyword in `DIST` block

amereghe:
Presently, if the `xp` or `yp` columns are present in a `FORMAT` statement, the unit stated by the user is ignored and the numbers are read in mrad. This is a problem in case of a file generated with `gpdist` (in the context of the Fluka-SixTrack coupling) in case of partially stripped ions or unstable hadrons, where the format of the file is non standard and a `FORMAT` statement is hence needed, but `xp` and `yp` are expressed in rad and not in mrad.

This PR aims at giving the possibility to parse the units of `xp` and `yp` in a `FORMAT` statement. Accepted values are `[1], [1000], [RAD], [MRAD]`.

User manual changed accordingly.


rdemaria:
For me it is ok. I merge...

amereghe:
The changelog was not changed...
...I will modify it in one of the open PRs...

rdemaria:
Correct! Thanks a lot!

PR #1056, opened 2020-02-24
Foxification of e-lenses (final)

amereghe:
This PR aims at finalising the foxification of e-lenses, as discussed with @rdemaria recently.
For Gaussian and uniform electron beam distribution, I have added the explicit expression for R1=0, both in the FOX and in the tracking part.
In addition, for the Gaussian distribution, I have added a linearisation of the kick of the full lens when r<1E-7 mm.

I have added few tests, each checking that the de-tuning for a single full e-lens with a specific profile and kick is sound.
The check is done reducing all the beam distributions to a constant one, and comparing to the tune-shift of a quad with the same normalised kick.

At the same time, I allow the use of the elens in the very degenerate case of the wire.
I have added a test for that - checking only kick during tracking and in FOX, but no specific test on tune-shift.

At the same time:
* more human readable output at various levels;
* removed bug on Gaussian and radial profiles, where the normalisation geometrical factor was wrongly taken into account twice.


amereghe:
Hi @rdemaria ,
I think this is the final version.
I haven't updated the original elens tests - I will wait for your green light before doing that, such that I update the tests only once.
That's why circleci fails.

Please find 
[here](https://github.com/SixTrack/SixTrack/files/4255533/2020-02-19_slides4RDeMaria.pdf) the updated .pdf with all the math - now it should be correct.
I think that the most interesting slides for you are 2 and 6.
Moreover, you can find on slide 8 the tune shifts given by the different types of e-lenses tested as calculated analytically and as computed by FOX.
Are they acceptable?

amereghe:
ok - I have updated the [slides](https://github.com/SixTrack/SixTrack/files/4262330/2020-02-19_slides4RDeMaria.pdf)
, showing also that (last slide):
* theta_prime(r) at R1 for the uniform distribution is different if the lens is hollow or full;
* theta_prime(r) at R2- for the Gaussian distribution tends to the limit at R2+ if R2 tends to infinity;

amereghe:
ok, @rdemaria , I am done here.
You can accept the PR.
I will draft all the math in a specific report, and add in the documentation a reference :)

amereghe:
Hi @tpersson ,
Thanks for the feedback. Shall I wait for your detailed review or shall I merge the PR?
In addition, FYI, following the suggestion of @rdemaria , I am de-facto drafting a report on the physics behind the implementation, but the repo will come much later than this PR...
Thanks!

amereghe:
ok to merge, as agreed with @tpersson

PR #1057, opened 2020-02-27
Update FLUKA IO ref.

JamesMolson:
update the fluka io module to the latest version


amereghe:
Shall we wait also for the same PR on the fluka_coupling repo before merging?

JamesMolson:
no need to wait, it shouldn't change any physics results with/without, the latest version only helps with debugging.

amereghe:
excellent, thanks a lot!

PR #1058, opened 2020-03-05
Previous interaction with a crystal collimator added to cry_interaction.dat

MarcoDAndrea:
This pull request adds an extra column to the cry_interaction.dat file. If a particle interacts with a crystal collimator after having already interacted with another or the same crystal collimator, the process ID of the previous interaction is stored in this column. This is meant to be a temporary addition in order to maintain the previously established post-processing and analysis setup for ongoing crystal collimation studies.


source/collimation.f90
```diff
@@ -270,6 +270,26 @@ subroutine coll_shuffleLostPart
...
+  use coll_common
+
+  integer j, tnapx
+
+  tnapx = napx
+  do j=napx,1,-1
+    if(llostp(j) .eqv. .false.) cycle
+
+    cry_proc_prev(j:tnapx) = cshift(cry_proc_prev(j:tnapx), 1)
+    cry_proc_tmp(j:tnapx) = cshift(cry_proc_tmp(j:tnapx), 1)
```
amereghe: Why not adding these lines directly to `coll_shuffleLostPart`

amereghe:
Hi @MarcoDAndrea ,
thanks for the changes. Just a minor suggestion from my side.

As already anticipated by @MarcoDAndrea , this PR is aimed at restoring that column on a temporary basis, not to have to re-write the analysis scripts already in place for a system that recently has been made baseline for HL-LHC. Thus it is important to stay productive on the short time scale.

The more appropriate implementation, implying a thorough re-thinking of parsing at the analysis level, can be done slowly in parallel to production.

PR #1059, opened 2020-04-08
Update geant4 based collimation

JamesMolson:
Missing #ifndefs as previously added for FLUKA (for pair map related topics).
Treat all particles as secondaries for the pair mapping with geant4 (i.e. just disable it).
Add Si as a material for geant4.


source/g4collimation/CollimationMaterials.cpp
```diff
@@ -42,6 +43,8 @@ CollimationMaterials::CollimationMaterials()
 	AddMaterial("Mo", Mo);
 	AddMaterial("W", W);
 	AddMaterial("Pb", Pb);
+	AddMaterial("Si", Si);
+	AddMaterial("SI", Si);
```
amereghe, line 47: why twice? I would leave only `Si`. In case the string can be input by the user, I would consider string comparison where the code lowers everything...

JamesMolson: on my ever growing to do list

PR #1060, opened 2020-04-22
Minor geant4 updates

JamesMolson:
1: Fix building with the latest g4 release
2: Fix a warning about a missing quote.


JamesMolson:
The message I was sent:

"There's a significant update in 10.6 that moves all Geant4 CMake targets to link to their dependencies through CMake imported targets. The upshot of that is Geant4 include directories, definitions will not contain any flags related to external dependencies. However, the flags will be applied to the build as the imported targets set these automatically for consumers
of the Geant4 targets."

amereghe:
Hi @JamesMolson ,
many thanks! I will merge the PR straight away.
I think we need to add some tests with Geant4 collimation - the same applies to the coupling with Fluka, though...

Issue #1064, opened 2020-07-09
Latest changes to elens module breaks C/R


vkbo:
Some error logging to the lerr output and crlog under the CR flag in the elens module is breaking CR build.

```
/home/vkbo/Code/SixTrack/Fork/build/SixTrack_cmakesix_CR_PYTHIA_gfortran_debug/elens.f90:2012:13:

 2012 |   write(crlog,"(a,i0,a)") "CR_POINT> ERROR Writing C/R file fort.",fileUnit," in ELENS"
      |             1
Error: Symbol ‘crlog’ at (1) has no IMPLICIT type
/home/vkbo/Code/SixTrack/Fork/build/SixTrack_cmakesix_CR_PYTHIA_gfortran_debug/elens.f90:2011:12:

 2011 |   write(lerr, "(a,i0,a)") "CR_POINT> ERROR Writing C/R file fort.",fileUnit," in ELENS"
      |            1
Error: Symbol ‘lerr’ at (1) has no IMPLICIT type
/home/vkbo/Code/SixTrack/Fork/build/SixTrack_cmakesix_CR_PYTHIA_gfortran_debug/elens.f90:1990:13:

 1990 |   write(crlog,"(a,i0,a)") "CR_CHECK> ERROR Reading C/R file fort.",fileUnit," in ELENS"
      |             1
Error: Symbol ‘crlog’ at (1) has no IMPLICIT type
/home/vkbo/Code/SixTrack/Fork/build/SixTrack_cmakesix_CR_PYTHIA_gfortran_debug/elens.f90:1989:12:

 1989 |   write(lerr, "(a,i0,a)") "CR_CHECK> ERROR Reading C/R file fort.",fileUnit," in ELENS"
      |            1
Error: Symbol ‘lerr’ at (1) has no IMPLICIT type
```

@amereghe

vkbo:
I needed to push a fix for SCATTER block parsing, so fixing this, and one of the elens tests as well.

amereghe:
Closing issue as it has been solved with PR #1065

PR #1065, opened 2020-07-09
Fix CR build, elens test for BOINC, and SCATTER parsing bug

vkbo:
This PR fixes three bugs:

* The C/R build for the elens module (Issue #1064)
* One of the elens tests had a non-existent file in extra_inputs.txt, which is used to generate the Sixin.zip file. This broke BOINC build.
* The check of parameters for the FIXED target profile type in the SCATTER block was set to the wrong value.

### DIFF
--- a/source/elens.f90
+++ b/source/elens.f90
@@ -1811,6 +1811,8 @@ end subroutine elens_kick_fox
 #ifdef CR
 subroutine elens_crcheck(fileUnit,readErr)
 
+  use crcoall
+
   integer, intent(in)  :: fileUnit
   logical, intent(out) :: readErr
 
@@ -1831,6 +1833,8 @@ subroutine elens_crcheck(fileUnit,readErr)
 end subroutine elens_crcheck
 
 subroutine elens_crpoint(fileUnit, writeErr)
+
+  use crcoall
 
   integer, intent(in)  :: fileUnit
   logical, intent(out) :: writeErr
--- a/source/scatter.f90
+++ b/source/scatter.f90
@@ -698,7 +698,7 @@ subroutine scatter_parseProfile(lnSplit, nSplit, iErr)
 
   case("FIXED")
     proType = scatter_proFixed
-    if(nSplit /= 3) then
+    if(nSplit /= 4) then
       write(lerr,"(a,i0)") "SCATTER> ERROR PROfile type FIXED expected 1 argument, got ",nSplit-3
       write(lerr,"(a)")    "SCATTER>       PRO name FIXED density[targets/m^2]"
       iErr = .true.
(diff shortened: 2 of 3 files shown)


vkbo:
> @vkbo , thanks a lot for spotting all these!

No problem. I just noticed the builds failing on the nightly builds. They were simple enough to fix ...

> I guess that the bug in the scatter module was related to the issue of the NaNs spotted by F. Van der Veken, right?

No, this is unrelated. But it is why I was looking at the code.

The NaN issue is not related to Scatter (although I'm sure the scatter process seeded the instability that led to it), but happens in the exact drift code. When unphysical coordinates are allowed to develop, the small angle version just sums them and continues, but the exact solver produces a negative square root and a complex particle energy. These particles are lost anyway, but some code should probably be added to filter them out. A DEBUG build will exit on them, a standard build will just run along with a quiet NaN.

In this particular case, a particle exited a magnet with the angles `[-1634.3278774109324, -122.50905792781495]`, and got caught in the next drift.

(Also, I think you guys should do the merging. I'm not really involved with SixTrack any more, and probably shouldn't have that level of access.)

amereghe:
@vkbo thanks for the analysis for the NaNs, will you still carry out the investigation?
For the merging, yes we take care of that. With @rdemaria we agreed to keep anyway the PR visible for at least 24h after approval before merging :)
...do you still have merge powers?

vkbo:
> @vkbo thanks for the analysis for the NaNs, will you still carry out the investigation?

I don't really have a lot of time to work much on this, but I can assist on issues related to Scatter, since the module was left in a beta state. I'll forward you the emails between me, Riccardo and Frederik on the NaN issue.

> For the merging, yes we take care of that. With @rdemaria we agreed to keep anyway the PR visible for at least 24h after approval before merging :)
> ...do you still have merge powers?

I do
image

amereghe:
ok no worries for the email threads, if @rdemaria is already in that.

I will fix your powers - actually, we should do some cleanup, though...

vkbo:
> ok no worries for the email threads, if @rdemaria is already in that.
> 
> I will fix your powers - actually, we should do some cleanup, though...

I think you just need to remove me from the Repo-Admins team. We're the only ones in it. Riccardo is Owner, so has access anyway.

amereghe:
I cannot remove you from the group repo-administrators

amereghe:
sorry I forgot also to ask you to edit the `CHANGELOG.md` file - I can do that for you at a later stage

vkbo:
Thanks!

I wouldn't add my two build fixes to the changelog, as they were broken after last release, so are not between-release changes. The Scatter bug is an old one though.

PR #1066, opened 2020-07-13
Update in e-lens module, radial profile from file

amereghe:
This PR aims at updating the radial profile from a file to be handled more consistently, in particular:
* the user requests to interpolate the profile with a polynomial of a given order; the coefficients are then calculated and used to integrate the profile and get the kicks;
* the user can request the use of the spline of third order (default option). This option ensures the continuity of the derivatives to first order, important for FOX calculations.

At the same time:
* fixing bug in expansion of arrays for storing radial profiles in case of more than one radial profile;
* removing automatic addition of point at R=0 when parsing radial profile. At the same time, found and fixed a bug in FOX with radial profiles.
* added some comments / clean up of code;

TO BE DONE:
* use coefficients for computing kick (FOX and tracking);
* update user/physics manuals and physics note;
* update tests;


JamesMolson:
@amereghe do you have any plans for this?

amereghe:
@JamesMolson I had plans... not sure when I will have time to follow this up.
Please check with @rdemaria @phermes Frederik and Stefano if you want to finalise this work or if you want to port it directly into XSuite - I've already had a preliminary discussion with @phermes , and there was not much interest in having this completed in SixTrack

PR #1068, opened 2020-07-17
Detect angular losses in exact drift

### LINKED ISSUE #1067: exact drift code generates silent NaNs
(from @vkbo )
> Exact drift fails with a debug build, and proceeds with a NaN value, for the particle momentum from the exact drift calculation because the square root becomes negative due to the sum of the squared angles exceed 1e6 here (track_thin.f90 on line 734):
```
          if(iexact) then ! EXACT DRIFT
            do j=1,napx
              pz       = sqrt(c1e6 - (yv1(j)**2 + yv2(j)**2))*c1m3 ! pz/p0
              xv1(j)   = xv1(j)   + stracki*(yv1(j)/pz)
              xv2(j)   = xv2(j)   + stracki*(yv2(j)/pz)
              sigmv(j) = sigmv(j) + stracki*((one - rvv(j)/pz)*c1e3)
            end do
```
> There is nothing that checks for these kind of numbers being in range, that would slow down the drift code. I'm not sure this is really an issue in itself, but it may be interesting to see exactly why this particle gets such large angles.
> The standard drift calculation doesn't fail on it because it just sums the angles.

amereghe:
This PR aims at inserting a check on the integrity of the `sqrt` in the exact drift code, and it fixes #1067 .
If a particle has a transverse momentum which is too large (and hence the exact drift code would miserably fail), the particle is considered as lost.

### DIFF
--- a/source/aperture.f90
+++ b/source/aperture.f90
@@ -756,7 +756,7 @@ subroutine aperture_checkApeMarker(turn, i, ix, llost)
 end subroutine aperture_checkApeMarker
 
 
-subroutine aperture_reportLoss(turn, i, ix)
+subroutine aperture_reportLoss(turn, i, ix, lbackin)
 !-----------------------------------------------------------------------
 !     P.Garcia Ortega, A.Mereghetti and D.Sinuela Pastor, for the FLUKA Team
 !     last modified: 17-01-2019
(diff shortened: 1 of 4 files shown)


rdemaria:
ok for me! Thanks!

JamesMolson:
@amereghe @rdemaria are there any objections to merging this?

rdemaria (approved):
ok for me, I did not see issues

amereghe:
Hi @JamesMolson sorry for the late reply.
Honestly I don't remember if I had anything else to change in mind - maybe a test of the functionality? In addition, I was wondering if porting this check also to the regular aperture checking - more because of the handled case rather than protecting against a potential FP issue.

PR #1069, opened 2020-07-17
New treatment of crystal miscut angle

MarcoDAndrea:
This pull request implements a new treatment of the crystal miscut angle, which includes a more precise calculation of the effective crystal length seen by the particles and of the resulting deflection angle.


JamesMolson:
@rdemaria I think you are the only one that can fix this?

rdemaria:
If you change 675000 to 680000 in https://github.com/SixTrack/SixTrack/blob/master/.circleci/config.yml#L16 should work

JamesMolson (approved):

PR #1076, opened 2021-02-03
Fix building with gcc >=10

JamesMolson:
So I tried to build with gcc/gfortran 10, and it currently does not.

This pr should fix that.

I had to disable building with distlib since it fails to link. The problem there seems to be a lack of header guards so the same variables are now pulled in multiple times and you get the same symbols defined multiple times. That would need to be fixed by @tpersson since it is in a separate repo.

crlibm had a similar issue which came from some internal diagnostic code that we don't need, so I just disabled that. This also should speed things up a tiny bit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,7 +74,7 @@ message(STATUS "Project Binary:   ${PROJECT_BINARY_DIR}")
 option(CRLIBM "Use correctly rounded libmath instead of system libmath" ON)
 option(SIXDA "Build differential algebra version (NOT dynamic aperture!)" ON)
 option(NAFF "Link to external NAFFlib for FMA" ON)
-option(DISTLIB "Link to external DISTlib for the beam distribution" ON)
+option(DISTLIB "Link to external DISTlib for the beam distribution" OFF)
 option(HASHLIB "Build the md5 hash library" ON)
 option(ZLIB "Build and link zlib. Needed for BOINC and ZIPF block" ON)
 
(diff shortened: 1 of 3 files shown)


rdemaria:
Thanks a lot! I will check with @tpersson ...

tpersson (approved):

rdemaria:
@JamesMolson could you confirm it is only a matter of adding `#ifndef distlib #define distlib ...`?

JamesMolson:
see as an example:
https://github.com/SixTrack/DISTlib/blob/master/source/distinterface.h

something like
```
#ifndef _distinterface_h_
#define _distinterface_h_

....

#endif
```

objects like dist, diststart, dim, etc etc are being  pulled in multiple times. I really don't know why the linker in earlier versions was ok with that, but 10 is not (or could be something in binutils).

```
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.34-990b2/x86_64-centos7/bin/ld: libdistlib.a(file_reader.c.o):(.bss+0x10): multiple definition of `dist'; libdistlib.a(distinterface.c.o):(.bss+0x10): first defined here
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.34-990b2/x86_64-centos7/bin/ld: libdistlib.a(file_reader.c.o):(.bss+0x4): multiple definition of `dim'; libdistlib.a(distinterface.c.o):(.bss+0x4): first defined here
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.34-990b2/x86_64-centos7/bin/ld: libdistlib.a(file_reader.c.o):(.bss+0x0): multiple definition of `distn'; libdistlib.a(distinterface.c.o):(.bss+0x0): first defined here
/cvmfs/sft.cern.ch/lcg/releases/binutils/2.34-990b2/x86_64-centos7/bin/ld: libdistlib.a(file_reader.c.o):(.bss+0x8): multiple definition of `diststart'; libdistlib.a(distinterface.c.o):(.bss+0x8): first defined here
```

etc

rdemaria:
I merge but distlib should be fixed...

PR #1078, opened 2021-03-03
Fix the time coordinate with geant4 collimation.

JamesMolson:
I finally enable the conversion of sigmv to time inside geant4.

dumping out the longitudinal phase space for a test bunch over 1000 turns makes the same pattern as for normal tracking.


rdemaria (approved):
Can you also update the changelog with the latest PRs?

PR #1083, opened 2021-05-18
Assorted updates

JamesMolson:


rdemaria (approved):

JamesMolson:
Ok, so here I add the following:

Being able to dump all the particle (bunch) info to a root file as part of the dump module.
Always outputting fort.208 energy loss for all collimation modes.
Some small bug fixes.
I also changed it so that if you disable CRLIBM it will now build with fma instructions.

PR #1086, opened 2021-06-08
Fix issue with the circular aperture check not working.

JamesMolson:
There was a missing line of code in the circular aperture check which allowed particles to pass through the apertures.


rdemaria (approved):
Good to know, I would not have guessed the line!

PR #1090, opened 2021-11-22
Add support for dealing with neutral particles in G4/FLUKA

JamesMolson:
Fairly simple...

We set mtc = 0 for neutrals.

This may have unintended outcomes if you don't know what you're doing, but you have to manually turn it on in both codes.

The fluka server code has already been updated to support this.


JamesMolson:
@rdemaria I need to merge this before I push all the checkpointing changes.

rdemaria (approved):

PR #1093, opened 2022-01-02
Scatter module cleanup and Pythia update

vkbo:
Massimo and Frederik asked me to go through the Scatter module again to clean up unfinished code. Most of it was already cleaned up, and the remaining unused code, mostly commented out, has now been removed.

I immediately had to update the canonicals for the scapper_pythia_realbeam test before I started changing anything. This is either due to the Pythia library, which I cannot guarantee was exactly the same as before because the old download location no longer works, or due to some other changes in the code. In any case, the changes were minimal.

In addition, I fixed the following issues:
* DistLib now builds again with GCC 10. There are multiple definitions exposed in the library, and GCC no longer allows this by default. I added the flag to revert to old behaviour for c code. The code should probably be updated. See https://gcc.gnu.org/gcc-10/porting_to.html
* I bumped the Pythia version in the Pythia build script from 8243 to 8306, which is the latest. The new library seems to be a lot faster. At least the tests ran about 40% faster. The test canonicals have been updated with minor changes. This happens every time a new Pythia version is used, so this was expected. Pythia is not binary/numerically consistent between versions. There are numerous minor changes in the scatter engine each time. This shouldn't affect the overall statistics of the beam.


rdemaria (approved):
Thanks a lot!

PR #1096, opened 2022-03-04
Fixed a bug in aperture check for transition markers, that appeared a…

bjlindst:
…fter the bug fix to the racetrack aperture check in the previous commit. The reason is that transition aperture check calls the check for racetrack. Because of this fix, the tests that were updated in the last commit have been reverted to the previous version, since they are now equivalent.


freddieknets (approved):
Continuation of bug fix in #1094. Thanks Bjorn!

PR #1098, opened 2022-04-04
Updated tests to be in line with elens physics correction

freddieknets:


rdemaria (approved):
excellent

amereghe:
Please keep in mind that in every elens test, there is a python script for crunching tracking results and making some quick overview plots showing the kick imparted by each elens in the various lattices. Since the change in kick from the bug fix is in the order of 1%, it is not worth re-committing the images, but it is (I think) good to cross-check that the plots do not substantially change.
Thanks guys!

